PR #996, opened 2023-05-18
feat: COPY ... TO ... (first iteration)

scsmithr:
```
db=> set gcs_service_account_key to '...';
SET
db=> copy (select * from glare_catalog.tables) to 'gs://sean-test-glaredb/test2.json';
COPY
```

Adds `COPY ... TO ...`, allow (unpartitioned) writes to object storage. Supported formats are json, csv, and parquet.

Note that `COPY table TO ...` parses correctly, but is currently unimplemented (for now). `COPY (SELECT ...) TO ...` works as expected.

Note that this isn't the final iteration. There was a bit of exploratory work done here to better understand how this would work, and if there's things we can reuse for querying object storage data sources. I plan on reworking that a bit to support more formats and features.

I also added in a `gcs_service_account_key` session var which _only_ works for `copy` right now. It made testing this a bit easier. I'm unsure if we want to leave that in. If we do, do we want to add similar vars for s3?


scsmithr:
Superseded by https://github.com/GlareDB/glaredb/pull/1159

PR #1086, opened 2023-06-03
feat: Add functions for reading data sources

scsmithr:
Adds table returning functions that allow for reading from external data
sources. These are meant to be used as an alternative to `CREATE EXTERANL
TABLE/DATABASE ...`.

What this looks like:

``` sql
select * from read_postgres('postgres://postgres:postgres@localhost/postgres', 'public', 'users)
```

These functions can be used anywhere in a query where a table identifier would
be used.

The list of supported functions can be found by querying the
`glare_catalog.functions` table:

``` text
> select function_name, function_type from glare_catalog.functions;
+----------------+---------------+
| function_name  | function_type |
+----------------+---------------+
| read_mongodb   | table         |
| read_mysql     | table         |
| read_bigquery  | table         |
| read_snowflake | table         |
| read_postgres  | table         |
+----------------+---------------+
```

Notable omissions are the functions for reading files from object storage, but
those should be added in soon.

# Code changes

Adds `sqlbuiltins` crate which provided the list and implementations for these
table functions. I did not move everything over yet, but I plan to move the
content of `metastore/src/builtins.rs` into this crate. I pulled this into its
own crate since both `metastore` and `sqlexec` need to reference this stuff,
whether that's for building the catalog, or executing the function itself.

The primary type of interest in this PR is the `TableFunc` trait. It's pretty
simple in that it's able to accept scalar arguments and return a
`TableProvider`. During sql planning, we'll do a lookup for a function by its
name, and create the table provider using the (constant) args provided. The
resulting logical plan is as if the table function was just another table.

# Docs

To come...

A very brief rundown for function args:

```
read_postgres(connection_str, schema, table)
read_bigquery(gcp_service_account_key, project_id, dataset_id, table_id)
read_mongodb(connection_str, database, collection)
read_mysql(connection_str, schema, table)
read_snowflake(account, username, password, database, warehouse, role, schema, table)
```

It might also be interesting the explore auto-generating docs from the json
output of the `glare_catalog.functions`:

``` text
% ./glaredb local --mode json --query 'select * from glare_catalog.functions'
[{"oid":16507,"schema_oid":16386,"function_name":"read_snowflake","function_type":"table","parameters":[],"parameter_types":[],"builtin":true},
{"oid":16506,"schema_oid":16386,"function_name":"read_postgres","function_type":"table","parameters":[],"parameter_types":[],"builtin":true},
{"oid":16508,"schema_oid":16386,"function_name":"read_mysql","function_type":"table","parameters":[],"parameter_types":[],"builtin":true},
{"oid":16505,"schema_oid":16386,"function_name":"read_mongodb","function_type":"table","parameters":[],"parameter_types":[],"builtin":true},
{"oid":16509,"schema_oid":16386,"function_name":"read_bigquery","function_type":"table","parameters":[],"parameter_types":[],"builtin":true}]
```

(Note `parameters` and `parameter_types` not currently writting to this table)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1618,6 +1618,7 @@ dependencies = [
  "env_logger",
  "paste",
  "rstest",
+ "sqlbuiltins",
 ]
 
 [[package]]
@@ -1635,7 +1636,6 @@ dependencies = [
  "decimal",
  "futures",
  "gcp-bigquery-client",
- "metastore",
  "metastoreproto",
  "mongodb",
  "mysql_async",
@@ -2998,6 +2998,7 @@ dependencies = [
  "proptest-derive",
  "prost",
  "prost-types",
+ "sqlbuiltins",
  "thiserror",
  "tokio",
  "tonic",
@@ -5119,6 +5120,31 @@ dependencies = [
 ]
 
 [[package]]
+name = "sqlbuiltins"
+version = "0.0.12"
+dependencies = [
+ "async-trait",
+ "datafusion",
+ "datasources",
+ "futures",
+ "logutil",
+ "metastoreproto",
+ "object_store",
+ "once_cell",
+ "pgrepr",
+ "regex",
+ "serde_json",
+ "sqlparser 0.34.0",
+ "telemetry",
+ "thiserror",
+ "tokio",
+ "tokio-postgres",
+ "tonic",
+ "tracing",
+ "uuid",
+]
+
+[[package]]
 name = "sqlexec"
 version = "0.0.12"
 dependencies = [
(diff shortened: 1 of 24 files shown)


vrongmeal (approved):
This looks very promising! I think we can go ahead with this.

Now that we support table functions, maybe we can change the virtual catalog to a function as well?

```sql
SELECT * FROM get_schemas(ext_db);
SELECT * FROM get_tables(ext_db, 'schema');
```

scsmithr:
> This looks very promising! I think we can go ahead with this.
> 
> Now that we support table functions, maybe we can change the virtual catalog to a function as well?
> 
> ```sql
> SELECT * FROM get_schemas(ext_db);
> SELECT * FROM get_tables(ext_db, 'schema');
> ```

Thinking the exact same thing.

PR #1089, opened 2023-06-05
feat: Support temporary tables

vrongmeal:
A couple of things to note here, I haven't used the `CatalogProvider` or `SchemaProvider`, instead just resolved the table myself. This is technically what is happening internally but there's more projection stuff that I am missing out on which makes the following query fail:

```
glaredb=> insert into abc (a, c) values (2, 4.678);
ERROR:  Error during planning: Inserting query must have the same schema with the table.
```

If we insert into all columns, this works:

```
glaredb=> insert into abc (a, b, c) values (2, 'abc', 4.678);
INSERT
```

Also need to refine `glare_catalog.tables` result (add oid there).

### DIFF
--- a/crates/sqlexec/src/planner/dispatch.rs
+++ b/crates/sqlexec/src/planner/dispatch.rs
@@ -21,9 +21,9 @@ use datasources::object_store::s3::{S3Accessor, S3TableAccess};
 use datasources::postgres::{PostgresAccessor, PostgresTableAccess};
 use datasources::snowflake::{SnowflakeAccessor, SnowflakeDbConnection, SnowflakeTableAccess};
 use metastore::builtins::{
-    DEFAULT_CATALOG, GLARE_COLUMNS, GLARE_DATABASES, GLARE_FUNCTIONS, GLARE_SCHEMAS,
-    GLARE_SESSION_QUERY_METRICS, GLARE_SSH_KEYS, GLARE_TABLES, GLARE_TUNNELS, GLARE_VIEWS,
-    VIRTUAL_CATALOG_SCHEMA,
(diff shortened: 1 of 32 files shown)


scsmithr:
> I haven't used the CatalogProvider or SchemaProvider, instead just resolved the table myself. 

I think this is for the best. We're not using these elsewhere, so I think it's fine that we're not using them here.

crates/sqlexec/src/planner/dispatch.rs
```diff
@@ -649,6 +659,19 @@ impl<'a> SystemTableDispatcher<'a> {
             datasource.append_value(table.options.as_str());
         }
 
+        // Append temporary tables.
+        for table in self.ctx.list_temp_tables() {
+            // TODO: Assign OID to temporary tables
```
scsmithr, line 664: I don't know if we'll want or need to assign oids to these tables. Going to check out what some other systems do.

scsmithr (commented):
So far looks good to me.

vrongmeal:
As far as I can see, DF doesn't support `insert into table (a, b) values (1, 2)` where the table has a different schema (yes, even the ordering of fields matters here).

I'll see if I can get a wrapper over `ExecutionPlan` to make it work. We have the schema, so it shouldn't be an issue. DF ignores the schema right now

crates/sqlexec/src/planner/dispatch.rs
scsmithr: Yep, seems like duckdb does too. I think we go with 0 for now. We'll probably want to extend the `SessionCatalog` eventually to support the temp stuff.

vrongmeal:
Removed most of "halts" (which were due to lack of `create table` and `insert into`). Couldn't get `insert into ABC (d, e, f)` working easily. Will look into it. The tests don't rely on it though :)

Issue #1090, opened 2023-06-05
Add sql syntax highlighting & formatting to terminal


universalmind303:
## Summary
When running `./glaredb local` from the terminal interface, It is rather difficult to navigate & hard to view the output of the queries.

## Description
for comparison, when using polars cli or duckdb, the table is truncated & sized according to the terminal window. SQL highlighting is also provided by many SQL cli interfaces. 

## Polars
![image](https://github.com/GlareDB/glaredb/assets/21327470/1882b950-751d-409a-88a4-e42d7fd8cf4b)
## Duckdb
![image](https://github.com/GlareDB/glaredb/assets/21327470/f1fa0774-e374-491c-8605-3613c456cf7b)

## GlareDB
![image](https://github.com/GlareDB/glaredb/assets/21327470/07e6391c-8789-4378-b775-1fafb39fad45)

universalmind303:
FWIW, It looks like the polars highlighter should port over pretty easily

https://github.com/pola-rs/polars/blob/main/polars-cli/src/highlighter.rs

PR #1091, opened 2023-06-05
feat(cli): add highlighting & multiline sql support

universalmind303:
adds support for the following to the cli: 
- syntax highlighting,  
- multiline support  _(as is common with other sql interfaces.)_ 
- history autofill

image

partially closes https://github.com/GlareDB/glaredb/issues/1090


universalmind303:
I'm not super familiar with how arrow displays the tables, but I'm also going to take a look to see if they have something similar to polars `POLARS_FMT_TABLE_*` arguments.

PR #1122, opened 2023-06-12
feat: Add `list_schemas` and `list_tables` table functions

vrongmeal:
Instead of using the `virtual_catalog` which was hacky, we can now use these functions as:

```
glaredb=> select * from list_schemas('abc');
 schema_name
-------------
 schema_0
 schema_1
(2 rows)

glaredb=> select * from list_tables('abc', 'schema_0');
 table_name
------------
 table_0
 table_1
(2 rows)
```


crates/datafusion_planner/Cargo.toml
```diff
@@ -30,6 +30,7 @@ datafusion = { workspace = true }
 async-trait = "0.1.68"
 async-recursion = "1.0.4"
 sqlbuiltins = {path = "../sqlbuiltins"}
+metastoreproto = {path = "../metastoreproto"}
```
vrongmeal, line 33: Unnecessary dependency, will remove it.

vrongmeal: Done

scsmithr (approved):
🚀

PR #1129, opened 2023-06-14
feat: Add native (internal) table support using delta-rs

scsmithr:
We're a data lake now

# What

Adds native table storage on top of delta files. This uses `TableOptionsInternal` for storing table information in metastore.

By default, the native tables are stored in memory. If the `--data-dir` flag is set (renamed from `local-file-path`), native tables will be persisted to disk. If `--service-account-path` is provided, native tables will be persisted to object storage. 

Relevant help:

```
  -f, --data-dir <DATA_DIR>
          Optional file path for persisting data.
          
          Catalog data and user data will be stored in this directory.

  -s, --service-account-path <SERVICE_ACCOUNT_PATH>
          Path to GCP service account to use when connecting to GCS.
          
          Sessions must be proxied through pgsrv, otherwise attempting to create a session will fail.
```

Note that if a service account is provided, it's required that the session is being proxied through pgsrv. There's an extra `gcs_storage_bucket` field on the response from Cloud that will be used during session create.

Creates and inserts "work".

# Bugs

Guaranteed.

Creating an empty table and immediately querying that table will return an empty batch (with no columns):

```
〉create table hello (a int);
create table
〉select * from hello;
++
++
```

Inserting into the table and querying will present the correct schema:

```
〉insert into hello values (1);
write success
〉select * from hello;
+---+
| a |
+---+
| 1 |
+---+
```

This is likely something we'll want to fix in delta-rs. I tried inserting an empty record batch with the correct schema, but that didn't work.

This is currently "all or nothing". We may want to optionally enable native tables if we uncover more issues in qa.

Also we need more tests.


crates/pgsrv/src/auth.rs
```diff
@@ -67,6 +67,8 @@ pub struct DatabaseDetails {
     pub database_id: String,
     /// ID of the user initiating the connection (UUID).
     pub user_id: String,
+    /// Bucket for session storage.
+    pub gcs_storage_bucket: String,
```
scsmithr, line 71: @f0ssel We need to return this from cloud during authenticate.

crates/glaredb/src/bin/main.rs
```diff
@@ -74,10 +72,18 @@ enum Commands {
...
         #[clap(short = 'f', long, value_parser)]
-        local_file_path: Option<PathBuf>,
+        data_dir: Option<PathBuf>,
+
+        /// Path to GCP service account to use when connecting to GCS.
+        ///
+        /// Sessions must be proxied through pgsrv, otherwise attempting to
+        /// create a session will fail.
+        #[clap(short, long, value_parser)]
+        service_account_path: Option<String>,
```
scsmithr, line 86: @f0ssel We need to supply this when running `glaredb server ...` in terraform and for dedicated deployments.

crates/glaredb/src/local.rs
```diff
@@ -136,9 +148,9 @@ impl LocalSession {
         println!("GlareDB (v{})", env!("CARGO_PKG_VERSION"));
         println!("Type \\help for help.");
 
-        let info = match (&self.opts.metastore_addr, &self.opts.local_file_path) {
-            (Some(addr), None) => format!("Persisting catalog on remote metastore: {addr}"),
-            (None, Some(path)) => format!("Persisting catalog at path: {}", path.display()),
+        let info = match (&self.opts.metastore_addr, &self.opts.data_dir) {
+            (Some(addr), None) => format!("Persisting catalog on remote metastore: {addr}"), // TODO: Should we continue to allow this?
```
vrongmeal, line 152: Let's keep allowing this. It might be handy to run some quick commands backed by a self-hosted metastore.

vrongmeal:
Looked at the PR. I don't understand how does it affect if metastore and native table storage are not backed up by the same mode? They both are independent entities. Metastore is the source of truth for catalog while native storage is something that's supposed to adhere with it.

vrongmeal (commented):
Left a comment

scsmithr:
> Looked at the PR. I don't understand how does it affect if metastore and native table storage are not backed up by the same mode? They both are independent entities. Metastore is the source of truth for catalog while native storage is something that's supposed to adhere with it.

My concern is mostly about losing data by changing configuration across starts. For example, if we store delta table storage in `bucket-a`, then later on configure the session to use `bucket-b`, we've "lost" data.

f0ssel:
@scsmithr I believe this should be merged and built before we can merge the respective cloud PRs, otherwise we will get crashloop backoff from providing unsupported flags.

scsmithr:
@vrongmeal @universalmind303 going to merge as-is and open up issues this evening. Feel free to postmortem review this, and I'll address.

Also we can tweak cli args/interfaces prior to tagging if necessary.

PR #1140, opened 2023-06-19
feat: Add support for credentials objects

### LINKED ISSUE #1126: Add credential database object
## Summary

Create new database object that can be used when accessing object storage.

### Specifications

Create

```sql
CREATE CREDENTIAL my_creds
PROVIDER aws
OPTIONS (access_key_id = '<access-key>', secret_access_key = '<secret>')
[COMMENT 'a comment'];
```

Drop

```sql
DROP CREDENTIAL [IF EXISTS] my_creds;
```

Usage (table creates)

```sql
CREATE EXTERNAL TABLE my_table
FROM s3
OPTIONS (credentials = my_creds, ...);
```

Usage (copy to)

```sql
COPY (SELECT * FROM glare_catalog.tables) 
TO 's3://sean-test-glaredb/test2.json'
OPTIONS (CREDENTIALS my_creds);
```

Usage (table func)
```sql
SELECT * FROM read_parquet('s3://...', my_creds);
```
(Note we don't have a `read_parquet` function yet.)

Like tunnel, credentials will be a top-level item and will live in its own unique namespace.

When resolving credentials during a table create, we’ll ensure that the data source type matches the credential type (e.g. `s3` tables should only allow `aws` credentials).

## Rationale

Definite

vrongmeal:
```
CREATE CREDENTIALS debug_creds
  PROVIDER debug
  OPTIONS (table_type = 'never_ending');

CREATE EXTERNAL TABLE debug_table
  FROM debug
  CREDENTIALS debug_creds;

DROP CREDENTIALS debug_creds;
```

Internal catalog:

```
〉select * from glare_catalog.credentials;
+-------+------------------+---------+----------+----------------+
| oid   | credentials_name | builtin | provider | comment        |
+-------+------------------+---------+----------+----------------+
| 20000 | aws_creds        | false   | aws      | test aws creds |
+-------+------------------+---------+----------+----------------+
```

Fixes https://github.com/GlareDB/glaredb/issues/1126, https://github.com/GlareDB/glaredb/issues/1141, https://github.com/GlareDB/glaredb/issues/1142

### TODO

- [x] Add more providers (AWS, GCS ...)
- [ ] Add credentials support for `COPY TO ...`
- [ ] Reference credentials instead of copying and add `ALTER CREDENTIALS ...` support

### DIFF
--- a/crates/metastore/src/database.rs
+++ b/crates/metastore/src/database.rs
@@ -9,8 +9,8 @@ use crate::validation::{
     validate_database_tunnel_support, validate_object_name, validate_table_tunnel_support,
 };
 use metastoreproto::types::catalog::{
-    CatalogEntry, CatalogState, DatabaseEntry, EntryMeta, EntryType, FunctionEntry, FunctionType,
-    SchemaEntry, TableEntry, TunnelEntry, ViewEntry,
+    CatalogEntry, CatalogState, CredentialsEntry, DatabaseEntry, EntryMeta, EntryType,
+    FunctionEntry, FunctionType, SchemaEntry, TableEntry, TunnelEntry, ViewEntry,
 };
 use metastoreproto::types::options::{
     DatabaseOptions, DatabaseOptionsInternal, TableOptions, TunnelOptions,
@@ -292,6 +292,8 @@ struct State {
     database_names: HashMap<String, u32>,
     /// Map tunnel names to their ids.
     tunnel_names: HashMap<String, u32>,
+    /// Map credentials names to their ids.
+    credentials_names: HashMap<String, u32>,
     /// Map schema names to their ids.
     schema_names: HashMap<String, u32>,
     /// Map schema IDs to objects in the schema.
@@ -309,6 +311,7 @@ impl State {
 
         let mut database_names = HashMap::new();
         let mut tunnel_names = HashMap::new();
+        let mut credentials_names = HashMap::new();
         let mut schema_names = HashMap::new();
         let mut schema_objects = HashMap::new();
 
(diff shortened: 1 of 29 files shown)


crates/sqlexec/src/parser.rs
```diff
@@ -765,6 +920,22 @@ mod tests {
         }
     }
 
+    #[test]
+    fn create_credentials_roundtrips() {
+        let test_cases = [
+            "CREATE CREDENTIALS qa PROVIDER debug OPTIONS (table_type = 'never_ending')",
+            "CREATE CREDENTIALS qa PROVIDER debug OPTIONS (table_type = 'never_ending') COMMENT 'for debug'",
```
scsmithr, line 927: After looking at this, I could potentially see PROVIDER causing some confusion since it someone could assume that PROVIDER is referencing some external service look hashicorp vault, and not the cloud provider we're trying to hook into.

I think this is good as-is, but wanted to bring it up in case there's any ideas for something better.

crates/metastore/src/database.rs
```diff
@@ -628,6 +656,36 @@ impl State {
...
+                        return Err(MetastoreError::DuplicateName(create_credentials.name));
+                    }
+
+                    // Create new entry
+                    let oid = self.next_oid();
+                    let ent = CredentialsEntry {
+                        meta: EntryMeta {
+                            entry_type: EntryType::Credentials,
+                            id: oid,
+                            // The tunnel, just like databases doesn't have any parent.
```
scsmithr, line 675:
```suggestion
                            // The credential, just like databases doesn't have any parent.
```

scsmithr (approved):

crates/sqlexec/src/parser.rs
vrongmeal: Let's just keep the keyword `FROM`? Just like tunnels, `aws`, `gcs` etc. will just be `credential_type`.

vrongmeal: @scsmithr what do you think about this?

scsmithr: Tbh I'd prefer `PROVIDER` over `FROM` here.

PR #1159, opened 2023-06-21
feat: `COPY TO` support

### LINKED ISSUE #962: `COPY TO` support
## Summary

From the plan:

- exporting data to parquet files in s3

## Tasks

- [ ] Implement
- [ ] https://docs.glaredb.com/docs/sql-reference/sql-commands/

vrongmeal:
Fixes #962

### DIFF
--- a/crates/datasources/src/common/sink/csv.rs
+++ b/crates/datasources/src/common/sink/csv.rs
@@ -1,0 +1,141 @@
+use std::fmt;
+use std::sync::Arc;
+
+use async_trait::async_trait;
+use datafusion::arrow::csv::{Writer as CsvWriter, WriterBuilder as CsvWriterBuilder};
+use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::common::Result as DfResult;
+use datafusion::error::DataFusionError;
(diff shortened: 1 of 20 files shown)


crates/datasources/src/common/sink/csv.rs
```diff
@@ -0,0 +1,141 @@
...
+        let (_id, obj_handle) = self.store.put_multipart(&self.loc).await?;
+        let mut writer = AsyncCsvWriter::new(obj_handle, BUFFER_SIZE, &self.opts);
+
+        while let Some(batch) = stream.next().await {
+            let batch = batch?;
+            writer.write_batch(&batch).await?;
+        }
+        writer.finish().await?;
+
+        Ok(0)
```
scsmithr, line 75: Should we always be returning 0, or should this be counting the number of rows being written.

scsmithr (approved):
Ship it!

crates/datasources/src/common/sink/csv.rs
vrongmeal: Aah yes, we should technically be returning number of rows.

PR #1160, opened 2023-06-21
feat: add "parquet_scan" function

universalmind303:
closes #1098 

note: this is currently limited to `parquet` for a single file. 

Todo: 
- add support for an array of files `parquet_scan([...])`

PR #1164, opened 2023-06-21
add support for reading from "file://" for `parquet_scan`

universalmind303:
follow up from #1160. 

in addition to the `http`/`https` schemes, we also support `file://`

```sql
select * from parquet_scan('file://path/to/file.parquet');
```

### DIFF
--- a/crates/datasources/src/object_store/mod.rs
+++ b/crates/datasources/src/object_store/mod.rs
@@ -2,6 +2,7 @@ use std::fmt::Debug;
 use std::str::FromStr;
 use std::sync::Arc;
 
+use datafusion::datasource::TableProvider;
 use errors::ObjectStoreSourceError;
 use object_store::path::Path as ObjectStorePath;
 use object_store::{ObjectMeta, ObjectStore};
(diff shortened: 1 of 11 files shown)


crates/datasources/src/object_store/mod.rs
```diff
@@ -37,10 +38,12 @@ impl FromStr for FileType {
     }
 }
 
+#[async_trait::async_trait]
 pub trait TableAccessor: Send + Sync {
     fn store(&self) -> &Arc<dyn ObjectStore>;
 
     fn object_meta(&self) -> &Arc<ObjectMeta>;
+    async fn into_table_provider(self, predicate_pushdown: bool) -> Result<Arc<dyn TableProvider>>;
```
scsmithr, line 46: imo we should remove `predicate_pushdown` (can do later). I don't know when there's a time where we _don't_ want this.

scsmithr (approved):
Ship it!

crates/datasources/src/object_store/mod.rs
universalmind303: if we remove, then couldn't we just change this to `impl TryFrom`?

scsmithr: Nah, we still need it be async since there's a bit of setup done in this function for some data sources (e.g. for postgres we need to make a connection and run a query to get the schema).

scsmithr: I mean would could do `IntoFuture` (https://doc.rust-lang.org/std/future/trait.IntoFuture.html), but I'm not really feelin it...

Issue #1150, opened 2023-06-20
Perf: Stream writes to native tables


scsmithr:
## Context

https://github.com/glaredb/glaredb/blob/c1362ac04d8882770ecc4b0eb20d577096287b67/crates/datasources/src/native/access.rs#L232-L248

Currently buffers everything in memory.

## Observed Behavior

High mem usage

## Expected Behavior

Periodically flush to avoid high mem usage

Issue #1157, opened 2023-06-20
Invalid data type timestamp when creating table


scsmithr:
## Context

Trying to run this:

```
create table execution_metrics (
    connection_id text,
    context_library_name text,
    context_library_version text,
    database_id text,
    elapsed_compute_ns int,
    error_message text,
    event text,
    event_text text,
    execution_status text,
    id text,
    loaded_at timestamp,
    original_timestamp timestamp,
    output_rows int,
    query_text text,
    received_at timestamp,
    sent_at timestamp,
    telemetry_tag text,
    timestamp timestamp,
    user_id text,
    uuid_ts timestamp
);
```

I get `Schema error: Invalid data type for Delta Lake: Timestamp(Nanosecond, None)`

greyscaled:
Delta parquet format supports microseconds, we use nanoseconds by default

PR #1165, opened 2023-06-22
feat: Stream writes to tables

scsmithr:
Closes https://github.com/GlareDB/glaredb/issues/1150

No longer buffering in memory.

When timestamps are properly supported in create tables (see https://github.com/GlareDB/glaredb/issues/1157), we can reuse a lot of our existing slt stuff to add tests for this.

### DIFF
--- a/crates/datasources/src/native/access.rs
+++ b/crates/datasources/src/native/access.rs
@@ -1,25 +1,22 @@
 use crate::native::errors::{NativeError, Result};
+use crate::native::insert::NativeTableInsertExec;
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::Schema as ArrowSchema;
 use datafusion::datasource::TableProvider;
 use datafusion::error::Result as DataFusionResult;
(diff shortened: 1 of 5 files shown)


universalmind303 (approved):

crates/datasources/src/native/access.rs
```diff
@@ -135,23 +132,19 @@ impl NativeTableStorage {
...
 pub struct NativeTable {
-    inner: Arc<NativeTableInner>,
+    delta: DeltaTable,
 }
 
 #[derive(Debug)]
-struct NativeTableInner {
-    delta: DeltaTable,
-}
+struct NativeTableInner {}
```
vrongmeal, line 139: Why not remove this?

vrongmeal (commented):
Looks good. Just a tiny comment.

crates/datasources/src/native/access.rs
scsmithr: Overlooked that, will fix.

PR #1168, opened 2023-06-22
feat: add "csv_scan" function

universalmind303:
can now read from csv''s both locally & remotely

```
〉select * from csv_scan('https://path/to/file.csv');
〉select * from csv_scan('file://path/to/file.csv');
```


scsmithr (approved):
Ship!

PR #1174, opened 2023-06-23
feat: Add clustercom module for compute cluster communication

scsmithr:


scsmithr:
closin, could be integrated with rpcsrv

PR #1182, opened 2023-06-23
feat: initial py bindings

universalmind303:
Exposes simple python bindings for glaredb. 

currently exposed methods

- `glaredb`: top level module
- `glaredb.connect(path: str) -> Session`: connect to an embedded instance using a path argument
- `Session.sql(query: str) -> ExecutionResult`: run a sql query 
- `ExecutionResult.to_arrow() -> pyarrow.Table`: convert result to a pyarrow table


universalmind303:
@scsmithr I noticed that I ended up having to duplicate a lot of code from the CLI. This is probably fine for a POC, but it may be a good idea to extract this shared logic into it's own crate. That'd make it a lot easier to have additional libraries too (such as rust or JS).

scsmithr:
> @scsmithr I noticed that I ended up having to duplicate a lot of code from the CLI. This is probably fine for a POC, but it may be a good idea to extract this shared logic into it's own crate. That'd make it a lot easier to have additional libraries too (such as rust or JS).

Duplicating the code for now seems reasonable to me. I'd rather duplicate now and figure out what we can abstract out later than try to foresee everything upfront.

universalmind303:
a few comments:

- There's a couple `todo`'s still in here for if you try to run a sql query with multiple statements
- this is using [just](https://github.com/casey/just) _(a better make)_ as the command runner. I'm not super familiar with xtask so i just used this as it was the easiest for me to get up & running with. We can leave it, or convert it to xtask, no strong preferences on my end. 
- once we have pypi setup, I'll go ahead & set up the CI for this.  see #1186

scsmithr:
> a few comments:
> 
>     * There's a couple `todo`'s still in here for if you try to run a sql query with multiple statements

Looks reasonable to me, we can follow up with more prs.

 
>     * this is using [just](https://github.com/casey/just) _(a better make)_ as the command runner. I'm not super familiar with xtask so i just used this as it was the easiest for me to get up & running with. We can leave it, or convert it to xtask, no strong preferences on my end.

I would prefer xtask, but we can leave this for now. I'm not familiar with python workflows yet, so we can keep just and use it for the time being.

>     * once we have pypi setup, I'll go ahead & set up the CI for this.  see [Chore: CI/CD for python bindings #1186](https://github.com/GlareDB/glaredb/issues/1186)

🚀

scsmithr (approved):
Ship! 🚀 

Gonna be messing around with this

PR #1183, opened 2023-06-24
feat: Windows builds

### LINKED ISSUE #1030: Chore: Conditionally compile ssh support
## Description

Our ssh stuff relies on openssh, and the crate is basically unix only. We'll want to conditionally compile in openssh only if the target is unix. If a `CREATE TUNNEL ...` command is executed on windows, then it should return an error along the lines of "SSH tunnels unsupported on windows" (and maybe link to docs or something).

### LINKED ISSUE #1053: Chore: Build release artifacts for windows
## Description

Build artifacts like we're doing with mac and linux.

Blocked on https://github.com/GlareDB/glaredb/issues/1030

scsmithr:
- Closes #1030
- Closes #1053

Add windows builds to the release yaml. 

To accomplish this, we need to conditionally compile in openssh, which impacts the ssh tunnels feature. On linux and mac, this continues to work as expected, while the windows code path will always return an error when attempting to create a tunnel.

This required a small amount of refactoring, and I tried to keep the conditional compilation to a single file. I also needed to tweak the cargo.toml to avoid pulling openssh if we're building on a platform that it doesn't support.

Conditional compilation is currently based on checking the target os, and not on adding a cargo feature. Longer term, we may want to put ssh tunnels behind its own flag, and enable/disable depending on the platform we're building for. I opted not to do that now because I'm unsure what we would want that to look like. Checking target os seemed like the quickest solution for now.

The executable successfully builds, but I have not actually tried running it on a windows machine...

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,16 +17,14 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # See https://github.com/GlareDB/glaredb/issues/1030
           - os: windows-latest
             target: x86_64-pc-windows-msvc
-          # - os: ubuntu-2004-8-cores
-          #   target: x86_64-unknown-linux-gnu
-          # - os: macos-12-xl
-          #   target: x86_64-apple-darwin
-          # - os: macos-12-xl
-          #   target: aarch64-apple-darwin
-
+          - os: ubuntu-2004-8-cores
+            target: x86_64-unknown-linux-gnu
+          - os: macos-12-xl
+            target: x86_64-apple-darwin
+          - os: macos-12-xl
+            target: aarch64-apple-darwin
 
     name: Build release (${{ matrix.target }})
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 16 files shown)


xtask/src/main.rs
```diff
@@ -113,7 +113,7 @@ fn run_fmt_check(sh: &Shell) -> Result<()> {
 
 fn run_dist(sh: &Shell, target: &Target) -> Result<()> {
     let triple = target.dist_target_triple()?;
-    cmd!(sh, "cargo build --release --bin glaredb --target {triple}").run()?;
+    cmd!(sh, "cargo build --bin glaredb --target {triple}").run()?;
```
scsmithr, line 116: revert before merge

.github/workflows/release.yaml
```diff
@@ -6,6 +6,8 @@ on:
     - cron: 0 12 * * *
   workflow_dispatch:
   push:
+    branches:
+      - "sean/windows-builds"
```
scsmithr, line 10: Remove before merge

universalmind303 (approved):

Issue #1189, opened 2023-06-26
Failing image build


scsmithr:

scsmithr:
Probably just missing python3

scsmithr:
Fixed

PR #1191, opened 2023-06-26
fix: Add python3 to dockerfile

scsmithr:


greyscaled:
Looks like it fixed #1189 , can we merge #1188 after?

scsmithr:
I think so?

PR #1195, opened 2023-06-27
feat: Allow reading from the python environment

scsmithr:
```python
import glaredb
import polars as pl

df = pl.DataFrame(
    {
        "A": [1, 2, 3, 4, 5],
        "fruits": ["banana", "banana", "apple", "apple", "banana"],
        "B": [5, 4, 3, 2, 1],
        "cars": ["beetle", "audi", "beetle", "beetle", "beetle"],
    }
)

con = glaredb.connect("/tmp/testing")

df = con.sql("select * from df where fruits = 'banana'").to_polars();

print(df)
```

Outputs

```
shape: (3, 4)
┌─────┬────────┬─────┬────────┐
│ A   ┆ fruits ┆ B   ┆ cars   │
│ --- ┆ ---    ┆ --- ┆ ---    │
│ i64 ┆ str    ┆ i64 ┆ str    │
╞═════╪════════╪═════╪════════╡
│ 1   ┆ banana ┆ 5   ┆ beetle │
│ 2   ┆ banana ┆ 4   ┆ audi   │
│ 5   ┆ banana ┆ 1   ┆ beetle │
└─────┴────────┴─────┴────────┘
```


crates/sqlexec/src/planner/context_builder.rs
```diff
@@ -70,6 +70,13 @@ impl<'a> PartialContextProvider<'a> {
                         }
                     }
                 }
+
+                if let Some(reader) = self.ctx.get_env_reader() {
+                    if let Some(table) = reader.resolve_table(table).unwrap() {
+                        return Ok(table);
+                    }
+                }
+
```
universalmind303, line 79: how will this work if we add support for other dataframe libraries? 

it looks like this only allows for a single "env_reader"? I'd kinda expect that just to be a static struct, not a dyn trait. Maybe you had something else in mind though?

universalmind303: oh nvm I think i answered my own question, the env reader is added to the sqlexec crate, so we could have a "python dataframe reader", or something that works for whatever other environment the exec is running in.

py-glaredb/src/environment.rs
```diff
@@ -0,0 +1,67 @@
...
+    fn resolve_table(
+        &self,
+        name: &str,
+    ) -> Result<Option<Arc<dyn TableProvider>>, Box<dyn std::error::Error + Send + Sync>> {
+        Python::with_gil(|py| {
+            let var = py.eval(name, None, None).unwrap();
+            let polars_type: &PyType = py
+                .import("polars.dataframe.frame")
+                .unwrap()
+                .getattr("DataFrame")
```
universalmind303, line 46: i _think_ this would work as well
```suggestion
                .import("polars")
                .unwrap()
                .getattr("DataFrame")
```

universalmind303 (approved):

PR #1198, opened 2023-06-27
chore: Prevent python linking for macos

scsmithr:
Enables the `extension-module` feature which seems to prevent linking in libpython. I'm not sure what that actually does, but it all still seems to work.

Before:

```
% otool -L target/debug/glaredb
target/debug/glaredb:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 60420.101.2)
	/nix/store/6y1di05r2i6ks6f737r0ckaih0gw3jss-python3-3.10.11/lib/libpython3.10.dylib (compatibility version 3.10.0, current version 3.10.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1971.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

After:

```
% otool -L target/debug/glaredb
target/debug/glaredb:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/System/Library/Frameworks/Security.framework/Versions/A/Security (compatibility version 1.0.0, current version 60420.101.2)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1971.0.0)
	/usr/lib/libiconv.2.dylib (compatibility version 7.0.0, current version 7.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1319.100.3)
```

I also removed the Cargo.lock in py-glaredb since it'll just use the workspace lock file.


universalmind303:
`extension-module`: 
> This will tell the linker to keep the Python symbols unresolved, so that your module can also be used with statically linked Python interpreters. Use this feature when building an extension module.

yeah i think that's definitely what we want. We aren't building a python app, but a library, so i think we definitely want this.

PR #1199, opened 2023-06-27
remove python dep from docker

universalmind303:
#1198 made it so we no longer need this in our dockerfile.


scsmithr:
Running workflow here just to make sure: https://github.com/GlareDB/glaredb/actions/runs/5392620549/jobs/9791226848

scsmithr (approved):

PR #1211, opened 2023-06-28
feat: Make data dir optional in python bindings

scsmithr:
If provided, both table storage and metastore storage will write to this directory. Otherwise, everything will be in memory.

Also sets up spill path too.

### DIFF
--- a/py-glaredb/examples/multiple.py
+++ b/py-glaredb/examples/multiple.py
@@ -1,0 +1,19 @@
+import glaredb
+import polars as pl
+
+con1 = glaredb.connect()
+con1.sql("create table hello (a int)").to_polars();
+
+t1 = con1.sql("select * from glare_catalog.tables where table_name = 'hello'").to_polars();
+print(t1)
+
+con2 = glaredb.connect()
+con2.sql("create table hello (a int)").to_polars();
+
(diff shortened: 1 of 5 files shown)


universalmind303:
not sure if it's really in scope of this PR, but what would happen if you wanted multiple in memory instances? 

I think something like this would be preferred as it matches many other in memory database formats

```py
mem_db1 = glaredb.connect(":memory1")
mem_db2 = glaredb.connect(":memory2")
disk_db = glaredb.connect("/tmp/glaredb1.db")
```

scsmithr:
`connect` is already spinning up independent engines, so multiple instances already works with multiple calls to `connect`.

I added a quick python script confirming this:

```python
import glaredb
import polars as pl

con1 = glaredb.connect()
con1.sql("create table hello (a int)").to_polars();

t1 = con1.sql("select * from glare_catalog.tables where table_name = 'hello'").to_polars();
print(t1)

con2 = glaredb.connect()
con2.sql("create table hello (a int)").to_polars();

t2 = con2.sql("select * from glare_catalog.tables where table_name = 'hello'").to_polars();
print(t2)

try:
    con2.sql("create table hello (a int)").to_polars();
except Exception:
    print("Duplicate table failed as expected")
```

PR #1224, opened 2023-06-29
remove py-glaredb from default members

universalmind303:
removes `py-glaredb` from the default members.

this makes it so that you do not need the pyo3 linking args set to run `cargo build`, but it also remains in the workspace.


py-glaredb/src/logical_plan.rs
universalmind303, line 1: remove

scsmithr (approved):

PR #1242, opened 2023-06-30
refactor: xtask --> justfile

### LINKED ISSUE #1238: Chore: switch from justfile to xtask for py-glaredb
## Description

switch to using an xtask command for python commands instead of the justfile

## What This Enables

unified tooling for the project.

universalmind303:
closes #1238 
this replaces the xtask build tool with `just`

I havent replaced any of the commands in the ci or docs, wanted to confirm this looks good before doing so. 

what I think is nice about this approach is that just will auto inherit the parent justfile & you don't have to manage the working dir at all when running the scripts, unless otherwise specified, it'll always use the dir of the justfile. 

```sh
# /glaredb
just fmt # runs `cargo fmt` from the root dir
just build # runs the build cmd from pwd

# /glaredb/py-glaredb
 just fmt # runs the parent `fmt` from the parent's dir `/glaredb`
 just build # runs the build cmd from pwd
```

it's also pretty easy to add subcommands
for example, this delegates all `python` commands to the subdir's justfile. 
```make
# Run py-glaredb subcommands. see `py-glaredb/justfile` for more details.
python cmd *args: protoc
  just py-glaredb/{{cmd}} {{args}}
```

```sh
just python build
# or 
cd py-glaredb && just build
```

like xtask, you can also list out the commands. 

image

closes https://github.com/GlareDB/glaredb/issues/1238

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - cory/justfile
   pull_request:
   workflow_dispatch:
 
@@ -24,22 +25,21 @@ jobs:
 
       - name: Cache
         uses: Swatinem/rust-cache@v2
-
+      - uses: extractions/setup-just@v1
       - name: Build
-        run: |
-          cargo xtask build
+        run: just build
 
       - name: Unit Tests
-        run: cargo xtask unit-tests
+        run: just unit-tests
 
       - name: Doc Tests
-        run: cargo xtask doc-tests
+        run: just doc-tests
 
       - name: Clippy
-        run: cargo xtask clippy
+        run: just clippy
 
       - name: Format
-        run: cargo xtask fmt-check
+        run: just fmt-check
 
       - name: Authenticate to Google Cloud
         uses: google-github-actions/auth@v1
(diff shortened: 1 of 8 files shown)


scsmithr:
This looks reasonable to me. I'd like to make sure @vrongmeal is good with this before merge.

universalmind303:
> This looks reasonable to me. I'd like to make sure @vrongmeal is good with this before merge.

sounds good, once/if i get the 👍 , i'll switch over the docs & ci & double check to make sure all the pipelines work.

vrongmeal:
This looks good to me!

.github/workflows/ci.yaml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - cory/justfile
```
universalmind303, line 7: remove before merge

universalmind303:
SLT failures seem unrelated

scsmithr:
> @scsmithr should we run the release builds off of this branch _just_ to be 💯 ?

Yeah I'm down with that

universalmind303:
> > @scsmithr should we run the release builds off of this branch _just_ to be 💯 ?
> 
> Yeah I'm down with that

is `main` in a state that we can run a release? I think we had a regression in the python `to_pandas` that we need to fix first?

scsmithr:
> is `main` in a state that we can run a release? I think we had a regression in the python `to_pandas` that we need to fix first?

Good point, let's fix that first, then we can test with a release.

universalmind303:
so just to confirm, we are good to merge this? or did you want to wait to merge this until we can test with a release?

scsmithr:
Let's merge!

PR #1251, opened 2023-07-03
fix: Fix explain queries

### LINKED ISSUE #1248: Explain is broken
## Context

```
explain select 1;
```

```
Error during planning: Invalid EXPLAIN command. Expression is empty
```

scsmithr:
Datafusion's `from_plan` function will error/panic if provided a logical plan for EXPLAIN/EXPLAIN ANALYZE alongside an empty expression list. This gets called when replacing paramters for query. Since we go through the prepare/bind/execute phases even for queries that don't have paramters, we would reach this every time.

The conditional check just pulls out the inner logical plan and replaces parameters for that.

```
% cargo run --bin glaredb -- local --query "explain select 1;"
    Finished dev [unoptimized + debuginfo] target(s) in 0.46s
     Running `target/debug/glaredb local --query 'explain select 1;'`
+---------------+--------------------------------------+
| plan_type     | plan                                 |
+---------------+--------------------------------------+
| logical_plan  | Projection: Int64(1)                 |
|               |   EmptyRelation                      |
| physical_plan | ProjectionExec: expr=[1 as Int64(1)] |
|               |   EmptyExec: produce_one_row=true    |
|               |                                      |
+---------------+--------------------------------------+
```

Closes #1248

### DIFF
--- a/crates/sqlexec/src/planner/logical_plan.rs
+++ b/crates/sqlexec/src/planner/logical_plan.rs
@@ -2,7 +2,7 @@ use crate::errors::{internal, Result};
 use datafusion::arrow::datatypes::{DataType, Field, Schema as ArrowSchema};
 use datafusion::common::OwnedTableReference;
 use datafusion::datasource::TableProvider;
-use datafusion::logical_expr::LogicalPlan as DfLogicalPlan;
+use datafusion::logical_expr::{Explain, LogicalPlan as DfLogicalPlan};
 use datafusion::scalar::ScalarValue;
 use datafusion::sql::sqlparser::ast;
 use metastoreproto::types::options::{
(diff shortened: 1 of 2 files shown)


crates/sqlexec/src/planner/logical_plan.rs
```diff
@@ -67,6 +67,25 @@ impl LogicalPlan {
     /// Note this currently only replaces placeholders for datafusion plans.
     pub fn replace_placeholders(&mut self, scalars: Vec<ScalarValue>) -> Result<()> {
         if let LogicalPlan::Query(plan) = self {
+            // Replace placeholders in the inner plan if the wrapped in an
+            // EXPLAIN.
+            //
+            // TODO: Make sure this is the correct behavior.
```
scsmithr, line 73: From some initial testing, it _does_ seem like postgres handles this case, but I'm getting some weird output when using pgprototest:

```
send
Parse {"query": "explain select $1"}
Bind {"values": ["4"]}
Execute
Sync
----
----

----
---- (no newline)

until
ReadyForQuery
----
ParseComplete 
BindComplete 
DataRow {"fields":["Result  (cost=0.00..0.01 rows=1 width=32)"]}
CommandComplete {"tag":"EXPLAIN"}
ReadyForQuery {"status":"I"}
```

scsmithr: Actually the extra lines just seem like a regression on our end.

vrongmeal (approved):

PR #1256, opened 2023-07-05
feat: support compute_engine authentication parameter

f0ssel:
Cloud can now target compute engines based on this option and will return the compute engine IP that matches the name when specified or otherwise will target the shared pool.

### DIFF
--- a/crates/pgsrv/src/proxy.rs
+++ b/crates/pgsrv/src/proxy.rs
@@ -339,6 +339,13 @@ impl<A: ProxyAuthenticator> ProxyHandler<A> {
                     None => user,
                 };
 
+                // Extract the compute engine name (optional) from startup params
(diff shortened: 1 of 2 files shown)


crates/pgsrv/src/proxy.rs
```diff
@@ -339,14 +339,21 @@ impl<A: ProxyAuthenticator> ProxyHandler<A> {
                     None => user,
                 };
 
+                // Extract the compute engine name (optional) from startup params
+                // Defaults to empty string which will target the shared pool
+                let compute_engine = match params.get("compute_engine") {
+                    Some(compute_engine) => compute_engine,
+                    None => "",
+                };
```
scsmithr, line 347: We'll want to grab this from `options` (which corresponds to the options part of the connection string).

scsmithr (approved):

PR #1258, opened 2023-07-06
chore: Add pytest

scsmithr:
As a consequence, I also needed to figure out how to get locally scoped variables.

This also fixes not being able to query pandas dataframes.

### DIFF
--- a/py-glaredb/src/environment.rs
+++ b/py-glaredb/src/environment.rs
@@ -18,19 +18,54 @@ impl EnvironmentReader for PyEnvironmentReader {
         name: &str,
     ) -> Result<Option<Arc<dyn TableProvider>>, Box<dyn std::error::Error + Send + Sync>> {
         Python::with_gil(|py| {
-            // Currently assuming any error getting the variable just means
-            // there's no variable with that name.
-            let var = match py.eval(name, None, None) {
-                Ok(var) => var,
-                Err(_) => return Ok(None),
+            let var = match search_var(py, name) {
+                Ok(Some(var)) => var,
+                _ => return Ok(None),
             };
 
             // since the resolve functions will err if the library is uninstalled,
             // dont `try` the results, we want to move on next resolver if this one errs.
-            resolve_polars(py, var)
-                .or_else(|_| resolve_pandas(py, var))
-                .or_else(|_| Ok(None))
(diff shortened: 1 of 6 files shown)


py-glaredb/src/environment.rs
```diff
@@ -18,22 +18,56 @@ impl EnvironmentReader for PyEnvironmentReader {
...
-                .or_else(|_| resolve_pandas(py, var))
-                .or_else(|_| Ok(None))
+            if let Ok(Some(table)) = resolve_polars(py, var) {
+                return Ok(Some(table));
+            }
+            if let Ok(Some(table)) = resolve_pandas(py, var) {
+                return Ok(Some(table));
+            }
+
+            Ok(None)
```
universalmind303, line 35: these code blocks should resolve the same. Should we generally try to prefer procedural if/let syntax over the functional `or_else`?

py-glaredb/src/environment.rs
```diff
@@ -18,22 +18,56 @@ impl EnvironmentReader for PyEnvironmentReader {
...
         })
     }
 }
 
+/// Search for a python variable in the current frame, or any parent frame.
+fn search_var<'py>(py: Python<'py>, name: &str) -> PyResult<Option<&'py PyAny>> {
```
universalmind303, line 41:
```suggestion
fn get_stack_locals<'py>(py: Python<'py>, name: &str) -> PyResult<Option<&'py PyAny>> {
```

universalmind303 (commented):
a few small comments, but overall looks good.

py-glaredb/src/environment.rs
scsmithr: The issue previously was `resolve_polars` would return `Ok(None)` when the variable exists, but isn't a polars dataframe. This would short circuit the `or_else` chain, and so we would never be able to get the table provider for reading the pandas dataframe.

The above change will return early only if the variable is found _and_ the variable is the correct dataframe type.

In general though, I don't have strong opinions on procedural vs functional. I just typically do what's easiest to reason about with the problem at hand.

universalmind303: ok got it. That makes more sense then.

universalmind303 (approved):

PR #1282, opened 2023-07-10
chore: Add remoteexec module for remote execution

scsmithr:
(wip)

Adds in a grpc service/client for remote execution of query plans.


scsmithr:
closin

PR #1286, opened 2023-07-11
fix: don't fetch filetype eagerly

### LINKED ISSUE #1284: Reading file extension is too eager
## Context

```
sean@Seans-Air glaredb % echo '{"hello": "world"}' > test
sean@Seans-Air glaredb % ./target/debug/glaredb local --query "select * from ndjson_scan('./test')"
Error: Error during planning: No file extension provided
```

https://github.com/GlareDB/glaredb/blob/17996a912dd6565738583771e7108cbd50d11093/crates/datasources/src/object_store/local.rs#L75

Getting the file extension should only be called if `file_type` is not provided.

adhish20:
Changed the logic to fetch filetype only if it is not provided

fixes #1284


vrongmeal (approved):
Yey! First PR 🥳

scsmithr (approved):

PR #1287, opened 2023-07-11
wip: binarycopyexec serialization

universalmind303:


crates/datasources/src/postgres/mod.rs
```diff
@@ -604,13 +633,22 @@ impl ExecutionPlan for BinaryCopyExec {
...
+        //     .await;
+        //     let stream = ChunkStream {
+        //         state: StreamState::Idle,
+        //         types: self.pg_types.clone(),
+        //         opener: opener?,
+        //         arrow_schema: self.arrow_schema.clone(),
+        //     };
+        //     stream
+        // })
     }
```
universalmind303, line 652: @scsmithr I've been struggling with figuring out how to implement the `execute` in a way that doesnt rely on `async`. 

the `tokio_postgres::Client` needs to be created inside the `execute`, but it relies on a tokio runtime. But since we need this serializeable, we can't have `BinaryCopyExec` hold a connection handle, or a runtime handle. 

We _could_ just block on this since `execute` is already sync. But wasn't sure if there was maybe a better way.

crates/metastoreproto/src/types/options.rs
```diff
@@ -916,7 +918,7 @@ impl From<TunnelOptionsDebug> for options::TunnelOptionsDebug {
     }
 }
 
-#[derive(Debug, Clone, Arbitrary, PartialEq, Eq)]
+#[derive(Debug, Clone, Arbitrary, PartialEq, Eq, Serialize, Deserialize)]
```
universalmind303, line 921: todo: remove these. Was just using them to make sure everything was properly serializable.

PR #1292, opened 2023-07-11
feat(py): add lazy execution to python

### LINKED ISSUE #1279: Lazy/eager evaluation api in python bindings
## execute
`sql` => lazy evaluation,
`execute` => eager evaluation

universalmind303:
closes #1279

### DIFF
--- a/py-glaredb/src/session.rs
+++ b/py-glaredb/src/session.rs
@@ -1,6 +1,7 @@
 use anyhow::Result;
 use datafusion::arrow::datatypes::Schema;
 use datafusion::arrow::record_batch::RecordBatch;
+use futures::lock::Mutex;
 use futures::StreamExt;
 use pgrepr::format::Format;
 use pyo3::{exceptions::PyRuntimeError, prelude::*, types::PyTuple};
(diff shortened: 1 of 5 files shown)


py-glaredb/src/session.rs
```diff
@@ -118,7 +130,7 @@ impl PyExecutionResult {
     /// Convert to Arrow Table
     /// Collect the batches and pass to Arrow Table
     #[allow(clippy::wrong_self_convention)] // this is consistent with other python API's
-    fn to_arrow(&mut self, py: Python) -> PyResult<PyObject> {
+    pub fn to_arrow(&mut self, py: Python) -> PyResult<PyObject> {
```
scsmithr, line 133: Does `pub` do anything special when using the `pymethods` macro?

scsmithr (approved):
Ship!

py-glaredb/src/session.rs
universalmind303: i don't think it has any impact on the python bindings. I used `pub` because i need to access from the other struct `PyLogicalPlan`. I suppose i could do `pub(crate)` as well.

scsmithr: I'm down keeping it `pub`.

PR #1305, opened 2023-07-13
WIP: use pr-target instead of pr

universalmind303:


universalmind303:
using this approach, we'll definitely want to double check to make sure our actions are using this settings.


image

or even the more aggressive
image

scsmithr:
> looks like it's stuck on `Expected — Waiting for status to be reported`.
> 
> I think i ran into this issue in the past, the `pull_request_target` has to be on main for this to trigger.

So would force merging this fix the issue for all future prs?

universalmind303:
> So would force merging this fix the issue for all future prs?

I think so. We can try it & if it's not working as expected we can revert the PR.

Issue #1297, opened 2023-07-12
Add `COPY .. TO ..` tests for GCS and S3


vrongmeal:
## Summary

Currently, we have limited support for `*_scan` functions. Once those are complete, add `COPY .. TO ..` tests for GCS and S3 (along-with testing credentials).

### Specifications

Just add more SLTs. Local tests cover a lot of cases, we only need to test creds and basic URL support for both.

## Goals

To have more tests :)

vrongmeal:
Resolved in #1314

Issue #1225, opened 2023-06-29
Allow multiple urls in `*_scan` functions


scsmithr:
## Summary

Eg

```
select * from parquet_scan(['./path/1.parquet', './path/2.parquet'])
```

scsmithr:
My initial thoughts here were we'd push it down to the exec and assume that all files provided are in the same bucket / use the same object store. If they differ, we could just error as I would think wanting to scan multiple parquet files from different buckets would be an exceptional use case.

The logical union is interesting, and definitely could be something we could implement to support the multiple buckets use case. However I think as a first pass, we just assume everything is in the same bucket for now.

scsmithr:
This was closed with #1314

PR #1314, opened 2023-07-14
feat: Support named arguments and idents in table functions.

### LINKED ISSUE #1308: Support named arguments in table functions
## Summary

Support named arguments in table functions.

### Specifications

```sql
SELECT * FROM csv_scan('s3://bucket/obj', 'aws_creds', region='us-east-1');
```

## Rationale

Passing strings and checking everything using the number of arguments causes many issues. Named arguments can act like "options".

## Tasks

- [ ] Support named arguments in `crates/datafusion_planner/src/relation/mod.rs`.
- [ ] Use this to support raw authentication parameters in `*_scan` functions:

    ```sql
    SELECT * FROM csv_scan('gs://bucket/obj', service_account_key = '...');
    SELECT * FROM csv_scan('s3://bucket/obj', region = '...', access_key_id = '...', secret_access_key = '...');
    ```

vrongmeal:
We can now add named arguments that can be passed into table functions which act like options basically.

Passing identifiers as arguments is also supported now.

```sql
SELECT * FROM csv_scan(
  's3://bucket/object',
  aws_creds,      -- Identifier, can also be quoted: "aws_creds"
  region => '...' -- Named argument, can be passed as `key => val`
);
```

Fixes #1308, #1297 

---

### ☑️ TODO

- [x] Raw credentials as named arguments.

### DIFF
--- a/crates/sqlbuiltins/src/functions.rs
+++ b/crates/sqlbuiltins/src/functions.rs
@@ -32,6 +32,7 @@ use metastore_client::types::options::{
 use num_traits::Zero;
 use once_cell::sync::Lazy;
 use std::collections::HashMap;
+use std::fmt::{self, Write};
 use std::ops::{Add, AddAssign};
 use std::pin::Pin;
 use std::sync::Arc;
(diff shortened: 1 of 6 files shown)


vrongmeal:
TODO: Raw credentials as named arguments.

scsmithr:
This looks very slick!

crates/sqlbuiltins/src/functions.rs
```diff
@@ -104,6 +104,53 @@ pub trait TableFuncContextProvider: Sync + Send {
     fn get_credentials_entry(&self, name: &str) -> Option<&CredentialsEntry>;
 }
 
+/// Value from a function parameter.
+#[derive(Debug, Clone)]
+pub enum FuncParamValue {
```
universalmind303, line 109: IMO, this is kinda hard to understand the flow. Couldn't we just pass around the entire `Expr` and do a `string_from_expr` instead of `string_from_scalar`? 


over in polars, I created a trait to extract values from `Expr`

```rust

pub(crate) trait FromSqlExpr {
    fn from_sql_expr(expr: &SqlExpr, ctx: &SQLContext) -> PolarsResult<Self>
    where
        Self: Sized;
}

impl FromSqlExpr for String {
    fn from_sql_expr(expr: &SqlExpr, _: &SQLContext) -> PolarsResult<Self>
    where
        Self: Sized,
    {
        match expr {
            SqlExpr::Value(v) => match v {
                SqlValue::SingleQuotedString(s) => Ok(s.clone()),
                _ => polars_bail!(ComputeError: "can't parse literal {:?}", v),
            },
            _ => polars_bail!(ComputeError: "can't parse literal {:?}", expr),
        }
    }
}
```

could likely do something similar here. so instead of 
```rust
string_from_scalar(args.next().unwrap().try_into()?)?;
```
we'd end up with
```rust
String::try_from_sql_expr(args.next.unwrap())?
```
which I feel is much more readable.

vrongmeal: That's a neat idea. I did it to explicitly differentiate between `Ident` and `ScalarValue`, so a little confused how should I implement the trait. Maybe have some default functions...

```rust
pub trait FromFuncParamValue {
    fn try_from_scalar(s: ScalarValue) -> Result<Self>;
    fn try_from_ident(s: String) -> Result<Self>;

    fn from_scalar(v: FuncParamValue) -> Result<Self> {
        match v {
            FuncParamValue::Scalar(s) => Self::try_from_scalar(s),
            _ => Err(...),
        }
    }

    fn from_ident(v: FuncParamValue) -> Result<Self> {
         match v {
            FuncParamValue::Ident(s) => Self::try_from_ident(s),
            _ => Err(...),
        }
    }
}
```

So we'll be able to:

```rust
String::from_scalar(args.next().unwrap())?
```

universalmind303: I think the idea is to defer the parsing of `Expr -> ScalarValue` and `Expr -> Ident`. We can just pass along the `Expr` 
instead of `args: Vec<FuncParamValue>,`
`args: Vec<Expr>,`

then just extract the needed value there. 

for example

```rust
async fn create_provider(
        &self,
        _: &dyn TableFuncContextProvider,
        args: Vec<FuncParamValue>,
        _opts: HashMap<String, FuncParamValue>,
        match args.len() {
            3 => {
                let mut args = args.into_iter();
                let conn_str = string_from_scalar(args.next().unwrap().try_into()?)?;
                let schema = string_from_scalar(args.next().unwrap().try_into()?)?;
                let table = string_from_scalar(args.next().unwrap().try_into()?)?;
               //...
            }
            _ => Err(BuiltinError::InvalidNumArgs),
        }
    }
```

would be something like this

```rust
async fn create_provider(
        &self,
        _: &dyn TableFuncContextProvider,
        args: Vec<Expr>,
        _opts: HashMap<String, Expr>,
        match args.len() {
            3 => {
                let mut args = args.into_iter();
                let conn_str = String::try_from_sql_expr(args.next().unwrap())?
                let schema = String::try_from_sql_expr(args.next().unwrap())?
                let table = String::try_from_sql_expr(args.next().unwrap())?
               // ...
            }
            _ => Err(BuiltinError::InvalidNumArgs),
        }
    }
```
and the `FromSqlExpr` would match on both scalar strings & ident
    
```rust
impl FromSqlExpr for String {
    fn from_sql_expr(expr: &SqlExpr, _: &SQLContext) -> PolarsResult<Self>
    where
        Self: Sized,
    {
        match expr {
           SqlExpr::Identifier(ident) => Ok(self.normalizer.normalize(ident).into())
            SqlExpr::Value(v) => match v {
                SqlValue::SingleQuotedString(s) => Ok(s.clone()),
                _ => polars_bail!(ComputeError: "can't parse literal {:?}", v),
            },
            _ => polars_bail!(ComputeError: "can't parse literal {:?}", expr),
        }
    }
}
```

universalmind303:
for more clarity... I have to do this for #1225. 

here's an example of what i'm thinking. 

https://github.com/GlareDB/glaredb/compare/universalmind303/multi-scan?expand=1

crates/sqlbuiltins/src/functions.rs
vrongmeal: The reason I'm not willing to do this is because it doesn't let us differentiate between the source of parameter.

For example:

```rust
let credentials: String = args.next().unwrap().from_sql_expr()?;
```

This will be parsed for both the cases, when the `Expr` is `Ident` or a `ScalarValue`. Basically, doesn't let us differentiate between a literal and an identifier, so both the following would work:

```
select * from csv_scan('gs://bucket/object', mycreds);

select * from csv_scan('gs://bucket/object', 'mycreds');
```

I understand that we don't defer parsing `Expr` in this case. What I think of this is, we eventually convert an `Expr` to a `FuncParamValue` (equivalent to an `Expr`).

If the above case I mentioned doesn't seem like a big deal, I think your method is better, but I am only confused about that. Will give it some thought today, would love to hear your opinions. Also tagging @scsmithr

vrongmeal: I thought about it; maybe it doesn't matter. If we ever need a strict distinction, we can create a wrapper over `String` like `Ident(String)` and implement the trait.

I will copy stuff from your PR to minimize conflict resolution...

vrongmeal: I think we still need something like `FuncParamValue` as in intermediate since we want named parameters as options that we can get by name. We can't force positions for named arguments. This is something we'll have to process before-hand.

vrongmeal: Pushed a change. I feel this is fine, having an intermediate `FuncParamValue`. It's much clearer to read with the trait `FromFuncParamValue`. I don't see any disadvantage of going this way since using this intermediate, we eventually just figure out during planning that the arguments are valid and something our functions will accept.

scsmithr: To throw in my input, I would _prefer_ if we were able to differentiate between idents and strings, but I think having a clear implementation is more important here.

I didn't have a chance to go through the previous iteration, but the code we have here looks fine to me. However I want to leave end decision up to you two.

scsmithr (approved):

crates/sqlbuiltins/src/functions.rs
vrongmeal: I am quite happy with this. I guess we can merge this for now. We'll make more changes later if required.

PR #1315, opened 2023-07-14
feat: adding show() method to python execution results

### LINKED ISSUE #1216: Add `show` method to python execution results
## Summary

Be able to print out results without needing to call `to_polars` or `to_pandas` first.

adhish20:
Fixes #1216


adhish20:
Output from the example :

pyarrow.Table
A: int64
fruits: large_string
B: int64
cars: large_string
----
A: [[1,2,5]]
fruits: [["banana","banana","banana"]]
...
pyarrow.Table
A: int64
fruits: string
B: int64
cars: string
-- schema metadata --
pandas: '{"index_columns": [{"kind": "range", "name": null, "start": 0, "' + 677
----
A: [[1,2,5]]
fruits: [["banana","banana","banana"]]
B: [[5,4,1]]
cars: [["beetle","audi","beetle"]]

py-glaredb/src/session.rs
```diff
@@ -171,4 +175,24 @@ impl PyExecutionResult {
             Ok(result)
         })
     }
+
+    #[allow(clippy::wrong_self_convention)] // this is consistent with other python API's
```
universalmind303, line 179: i don't think we need this for `show`

py-glaredb/src/logical_plan.rs
```diff
@@ -60,6 +60,9 @@ impl PyLogicalPlan {
     fn to_pandas(&self, py: Python) -> PyResult<PyObject> {
         self.collect(py)?.to_pandas(py)
     }
+    fn show(&self, py: Python, show_metadata: bool, n_cols: u64) -> PyResult<()> {
+        self.collect(py)?.show(py, show_metadata, n_cols)
+    }
```
universalmind303, line 65: shouldn't this be `n_rows` not `n_cols`?

adhish20: the to_string API from pyarrow supports only number of columns and not number of rows

https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.to_string

py-glaredb/examples/show.py
```diff
@@ -0,0 +1,27 @@
...
+        "A": [1, 2, 3, 4, 5],
+        "fruits": ["banana", "banana", "apple", "apple", "banana"],
+        "B": [5, 4, 3, 2, 1],
+        "cars": ["beetle", "audi", "beetle", "beetle", "beetle"],
+    }
+)
+
+con = glaredb.connect()
+
+con.sql("select * from df where fruits = 'banana'").show(True, 4);
```
universalmind303, line 27: I think the signature should use kwargs. 

`show(n_rows, show_metadata=False)`

such as 
```python
x.show(10)
x.show(n_rows=10)
x.show(10, show_metadata=True)
x.show(10, show_metadata=False)
x.show(n_rows=10, show_metadata=False)
```

py-glaredb/src/logical_plan.rs
```diff
@@ -60,6 +60,9 @@ impl PyLogicalPlan {
     fn to_pandas(&self, py: Python) -> PyResult<PyObject> {
         self.collect(py)?.to_pandas(py)
     }
+    fn show(&self, py: Python, show_metadata: bool, n_cols: u64) -> PyResult<()> {
+        self.collect(py)?.show(py, show_metadata, n_cols)
```
universalmind303, line 64: for the logical_plan, we should push down the `n_rows` to the plan. No sense in over fetching data 😃

universalmind303 (changes requested):

py-glaredb/src/session.rs
```diff
@@ -171,4 +175,24 @@ impl PyExecutionResult {
...
+
+            let key_vals: &[(&str, PyObject)] = &[
+                ("show_metadata", show_metadata.to_object(py)),
+                ("preview_cols", n_cols.to_object(py)),
+            ];
+            let dict = key_vals.into_py_dict(py);
+            let result = table.call_method(py, "to_string", (), Some(dict))?;
+            println!("{}", result);
+            Ok(())
+        })
```
universalmind303, line 195: i think instead of going in to pyarrow, we can just print the execution result as is.

adhish20: Makes sense. I will look into that

py-glaredb/examples/show.py
universalmind303: On second thought... I think this shouldn't have any args. I think having similar functionality to duckdb's [show](https://duckdb.org/docs/api/python/relational_api) would be a good starting point. 

if you want to limit the n_rows, you can just add `limit x` to your sql query.

py-glaredb/src/logical_plan.rs
universalmind303: On second thought... I think this shouldn't have any args. I think having similar functionality to duckdb's [show](https://duckdb.org/docs/api/python/relational_api) would be a good starting point.

if you want to limit the n_rows, you can just add limit x to your sql query.

py-glaredb/examples/show.py
adhish20: cool. will remove the args

adhish20:
Updated Output looks like

+---+--------+---+--------+
| A | fruits | B | cars   |
+---+--------+---+--------+
| 1 | banana | 5 | beetle |
| 2 | banana | 4 | audi   |
| 5 | banana | 1 | beetle |
+---+--------+---+--------+

+---+--------+---+--------+
| A | fruits | B | cars   |
+---+--------+---+--------+
| 1 | banana | 5 | beetle |
| 2 | banana | 4 | audi   |
| 5 | banana | 1 | beetle |
+---+--------+---+--------+

PR #1317, opened 2023-07-14
multiple urls in `_scan` functions

universalmind303:


universalmind303:
@scsmithr this should _functionally_ work now, Ill work on cleaning up the todo's & error handling next.

scsmithr:
Union seems reasonable to me.

Longer term, we can look at using partitioned files: https://github.com/apache/arrow-datafusion/blob/413eba124fed89b4189cba214c737b2b5e1de12a/datafusion/core/src/datasource/physical_plan/mod.rs#L147-L156, or the listing table: https://github.com/apache/arrow-datafusion/blob/413eba124fed89b4189cba214c737b2b5e1de12a/datafusion/core/src/datasource/listing/table.rs#L60-L70

Since we're using some of the native execs and have the object store registry set up, integrating the listing table might be more straightforward than it was previously.

universalmind303:
holy smokes! https://github.com/GlareDB/glaredb/pull/1314 pwned this PR. had to do a full rewrite, hoping this passes the SLT's.

crates/sqlbuiltins/src/functions/conversion.rs
```diff
@@ -0,0 +1,372 @@
...
+
+pub trait ExtractInto<T> {
+    fn extract(&self) -> Result<T>;
+    fn extract_named(&self, _name: &'static str) -> Result<T> {
+        Err(BuiltinError::Unimplemented("named parameters"))
+    }
+}
+pub trait ExtractNamedInto<T> {
+    fn extract_named(&self, name: &'static str) -> Result<T>;
+}
```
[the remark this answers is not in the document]
universalmind303: so the issue i had with `get_constant_function_arg` was that the Scalar List value requires a `FieldRef` which didn't really make sense in this context. 

I agree that these could likely be combined. But I feel in order to do so, we need to go from sql ast directly to our own struct.  If you have any other ideas on how to use the `FuncParamValue` for list types, i'd appreciate some help getting this one over the finish line.

crates/datafusion_planner/src/relation/mod.rs
```diff
@@ -133,9 +135,9 @@ impl<'a, S: AsyncContextProvider> SqlQueryPlanner<'a, S> {
         }
     }
 
-    /// Get a constant expression literal from a function argument.
-    ///
-    /// Returns an optional name for the argument.
+    // Get a constant expression literal from a function argument.
+    //
+    // Returns an optional name for the argument.
```
vrongmeal, line 140: Just curious, is there a reason you made this a normal comment instead of a doc-comment?

crates/sqlbuiltins/src/functions/conversion.rs
```diff
@@ -0,0 +1,212 @@
...
+                expected: DataType::List(Arc::new(Field::new("list", DataType::Utf8, false))),
+            }),
+        }
+    }
+
+    fn is_param_valid(value: &FuncParamValue) -> bool {
+        matches!(value, FuncParamValue::Scalar(ScalarValue::Utf8(Some(_))))
+    }
+}
+
```
vrongmeal, line 94: I'm not sure if it's allowed but can we implement this for `Vec<T> where T: FromFuncParamValue` instead?

vrongmeal (commented):
Review is all just minor comments. I have a couple of questions:

1. I want to understand why do we require changing the `TableFunc` interface (basically why do we need to create a logical plan here)?

2. What's the reason behind `create_lp_for_named` and `create_lp_for_positional`. Where does this make a difference. I assumed the functions would be something like:
  ```
  csv_scan( '<url>' , ... )
  csv_scan( [ '<url_1>', '<url_2>' ], ... )
  ```

universalmind303:
> I want to understand why do we require changing the TableFunc interface (basically why do we need to create a logical plan here)?

it's because the multiple urls don't return a `provider`, they return a union plan. the providers are set up in a way that don't really easily allow them to consume multiple, so it made sense to just perform a logical union when passing in an array of urls.

crates/sqlbuiltins/src/functions/conversion.rs
universalmind303: i think we'd need to add some additional logic to map it to a dtype for `is_param_valid` to work.

vrongmeal:
I feel handling arguments like this is very flaky. S3 takes an optional (named) argument `region` whereas GCS doesn't. We should handle named arguments as options and not something concrete. Handling them differently will cause issues.

crates/sqlbuiltins/src/functions/conversion.rs
vrongmeal: Yeah. Just noticing that the current logic would fail as well. We do need additional logic anyways.

vrongmeal:
> yes, but inside each of those, (`named` and `positional`) they each handle the datasource differently. 
> 
> I think eventually it'd make sense to split out the `ObjScanTableFunc` for each source (s3,gs,http,...). But this seems to work for now. 

The reason I feel arguments shouldn't be handled differently is because any case when a positional argument is introduced will need to be replicated again (since it may be optional). Even in the current code that we have in this PR, we have `unimplemented` where there are multiple urls with named arguments. Named arguments shouldn't make any difference since eventually we have everything that the datasource requires. So for current implementation we can't use s3 since it requires a named argument. The logic is all right there, seems like we're complicating it a lot. Eventually we just want to get the same parameters. Arguments are just the means to get those.

universalmind303:
> > yes, but inside each of those, (`named` and `positional`) they each handle the datasource differently.
> > I think eventually it'd make sense to split out the `ObjScanTableFunc` for each source (s3,gs,http,...). But this seems to work for now.
> 
> The reason I feel arguments shouldn't be handled differently is because any case when a positional argument is introduced will need to be replicated again (since it may be optional). Even in the current code that we have in this PR, we have `unimplemented` where there are multiple urls with named arguments. Named arguments shouldn't make any difference since eventually we have everything that the datasource requires. So for current implementation we can't use s3 since it requires a named argument. The logic is all right there, seems like we're complicating it a lot. Eventually we just want to get the same parameters. Arguments are just the means to get those.

> The logic is all right there, seems like we're complicating it a lot.

i totally agree, but from a different lens. the presence of  `FuncParamValue` actually complicates this a lot. Previously, it was just taking in the raw `FunctionArg` and handling it accordingly based on the signature, but now there is the extra step of converting to `FuncParamValue` and `ScalarValue` that complicated this pr quite a bit.  I definitely think the handling of various function signatures as a whole requires a bit more thought & design beyond the scope of this PR.

universalmind303:
previously, `FunctionArg` had everything needed to handle the function signature, now we need to look in both `args` and `opts` instead of just `args`, so it complicates the logic of getting the values. 

I think we are both in agreement that the function signatures need some revisiting. Maybe we should create a follow up ticket to get a more holistic approach in place for handling different function signatures.

vrongmeal:
Yes, it does seem like function arguments are complicated but named arguments are not part of positional arguments. Nothing's really changed, only that you can have more options when extracting something.

So we can:

```rust
let urls = if String::is_param_valid(&args[0]) {
    let url = String::param_into(args.next().unwrap());
    vec![url]
} else {
    Vec<String>::param_into(args.next().unwrap())
};

// parse URLs and get scheme
// get table providers for each URL -> this will be same as the code on main (i think no changes will be required here)
// create the union plan
```

Something like this would unify the code. Instead of a single URL, we just have multiple URLs.

vrongmeal:
Let me list down the reasoning behind the current design. Function signatures are similar to any other language:

```
function( pos_arg_1, pos_arg_2, ..., named_arg_1 => .., named_arg_2 => .. )
```

Named args can be in any order. `named_arg_1` can come after `named_arg_2`; hence, we collect all of these into a `HashMap` since they are to be referred to by their name. If we directly parse it from a `FunctionArg`, we don't get the flexibility of using named arguments in any order. Somewhere, we need to process all the arguments to collect, so we have the `FuncParamValue`. This can be considered as an alternative to `Expr`.

universalmind303:
i understand the reasoning & the logic, the reason they are in separate functions is because the logic to handle each signature is different. 

```sql
csv_scan('gs://my/file.csv', creds)
```
is not the same as 
```sql
csv_scan('gs://my/file.csv', service_account_id => 'creds')
```

similarly, the logic to handle 
```sql
csv_scan('gs://my/file.csv', service_account_id => 'creds')
```
is not the same as 
```sql
csv_scan('s3://my/file.csv', access_key => 'key', ... )
```
which is not the same as 
```sql
csv_scan(['s3://my/file.csv'], access_key => 'key', ... )
```

they all require different handling of the named & positional arguments. 

which is why I suggest that we further split out the object store to be able to more robustly handle all of these different function signatures. I felt that getting in the functionality was more important than doing a big refactor. I already had to refactor `functions.rs` as it was getting unruly. I didn't want this to become any larger of a PR than necessary. So this is very much a "let's get this to work" approach, knowing that we'll likely have to come back & refactor later anyways.

universalmind303:
i think a typestate pattern would be ideal for a refactor. 

```rust
trait ObjScanFileType: Send + Sync {
    fn file_type(&self) -> FileType;
    fn name(&self) -> &'static str;
}
pub struct ParquetScan;

impl ObjScanFileType for ParquetScan {
    fn file_type(&self) -> FileType {
        FileType::Parquet
    }
    fn name(&self) -> &'static str {
        "parquet_scan"
    }
}

trait ObjScanSource: Send + Sync {
    fn source(&self) -> DatasourceUrlScheme;
}

pub struct GcsScan;
impl ObjScanSource for GcsScan {
    fn source(&self) -> DatasourceUrlScheme {
        DatasourceUrlScheme::Gcs
    }
}

#[derive(Debug, Clone, Copy)]
struct ObjScanTableFunc<FILETYPE, SOURCE>(FILETYPE, SOURCE)
where
    FILETYPE: ObjScanFileType,
    SOURCE: ObjScanSource;
    
    
#[async_trait]
impl TableFunc for ObjScanTableFunc<ParquetScan, GcsScan> {
    fn name(&self) -> &str {
        self.0.name()
    }

    fn parameters(&self) -> &[TableFuncParameters] {
        todo!()
    }

    async fn create_provider(
        unreachable!()
    }

    async fn create_logical_plan(
    ) -> Result<LogicalPlan> {
       todo!()
    }
}
```

then just create some wrapper struct that matches on the filetype & scheme & delegates accordingly.  It'd make the code very clear & extensible, without having a bunch of nested logic to handle the different filetypes & schemes.

PR #1322, opened 2023-07-15
fix: Explicitly use `Ident` in table function parameters.

vrongmeal:
Can only specify identifiers as parameters when required. For example:

```sql
SELECT * FROM list_schemas( ext_db_name );

-- `Ident`s can be quoted.
SELECT * FROM list_tables( ext_db_name, "SchemaName" );

SELECT * FROM csv_scan( '<url>', creds_obj );
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-  pull_request_target:
+  pull_request:
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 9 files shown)


vrongmeal:
@greyscaled Can we update the query on cloud side to use double quotes instead of single quotes once this is merged?

```sql
SELECT * FROM list_schemas( "ext_db_name" );

SELECT * FROM list_tables( "ext_db_name", "schema_name" );
```

scsmithr (approved):
Love this

I believe all the relevant queries we need to change are in this file: https://github.com/GlareDB/cloud/blob/main/site/src/util/queries.ts

greyscaled:
> @greyscaled Can we update the query on cloud side to use double quotes instead of single quotes once this is merged?
> 
> ```sql
> SELECT * FROM list_schemas( "ext_db_name" );
> 
> SELECT * FROM list_tables( "ext_db_name", "schema_name" );
> ```

Looks good, and appreciate the contrib downstream to cloud, thanks @vrongmeal 🙌🏻 !

Issue #1324, opened 2023-07-16
Disable using local datasource on the server


vrongmeal:
## Context

The command `COPY my_table TO './local_path.csv'` would work and might overload the server. Similarly, a user can read a local file on the server which shouldn't be possible.

## Expected

Error when running local commands on the server.

## Actual

Command works as expected.

## Impact

Security concerns.

eitsupi:
Hello. I'm having trouble setting up GlareDB for CI in PRQL/prql#3658.
The ability to read local files is definitely useful.
Could you consider a workaround such as allowing only certain directories like ClickHouse?
<https://clickhouse.com/docs/en/operations/server-configuration-parameters/settings#user_files_path>

PR #1335, opened 2023-07-17
feat: Read-only session vars

scsmithr:
Adds read-only session variables, and moves a bunch of session configuration to `SessionVars`.

Moving session configuration to `SessionVars` provides the following benefits:

- Centralizes session configuration. This still _some_ amount of configuration not stored here, but we can work on moving stuff into this as needed.
- Allows users to inspect variables.

As a user, I can now view session limits (and more) via the `SHOW` command:

```sql
〉show memory_limit_bytes;
+--------------------+
| memory_limit_bytes |
+--------------------+
| None               |
+--------------------+
```

Attempting to set a read-only session var will error:

```sql
〉set memory_limit_bytes to 9999;
Error: Variable is readonly: memory_limit_bytes
```

This may be useful with https://github.com/GlareDB/glaredb/issues/1324 in that we can add an `enable_local_datasources` variable that gets set to false in pgsrv.

### DIFF
--- a/crates/pgsrv/src/handler.rs
+++ b/crates/pgsrv/src/handler.rs
@@ -19,7 +19,7 @@ use pgrepr::format::Format;
 use pgrepr::scalar::Scalar;
 use sqlexec::context::{OutputFields, Portal, PreparedStatement};
 use sqlexec::engine::{SessionLimits, SessionStorageConfig};
-use sqlexec::vars::VarSetter;
+use sqlexec::vars::{SessionVars, VarSetter};
 use sqlexec::{
     engine::Engine,
     parser::{self, StatementWithExtensions},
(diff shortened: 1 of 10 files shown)


universalmind303 (approved):

crates/pgsrv/src/handler.rs
```diff
@@ -233,20 +234,40 @@ impl ProtocolHandler {
...
+        // Limits
+        vars.max_datasource_count
+            .set_and_log(Some(max_datasource_count), VarSetter::System);
+        vars.memory_limit_bytes
+            .set_and_log(Some(memory_limit_bytes), VarSetter::System);
+        vars.max_tunnel_count
+            .set_and_log(Some(max_tunnel_count), VarSetter::System);
+        vars.max_credentials_count
+            .set_and_log(Some(max_credentials_count), VarSetter::System);
+
```
vrongmeal, line 255: Shouldn't we error (to the user) if we can't set any system var here? We don't want faulty sessions.

PR #1340, opened 2023-07-18
feat: allow creating tables from a source query

### LINKED ISSUE #1156: CREATE TABLE ... AS ... support
## Summary

Create table from query

Since we have native tables now, might as well add this in too.

## Rationale

Reading parquet files or json files from data sets

adhish20:
Fixes #1156


testdata/sqllogictests/table.slt
```diff
@@ -19,3 +19,6 @@ create table table_with_list (a text[])
 
 statement ok
 insert into table_with_list values (['hello', 'world'])
+
+statement ok
+create table t1 as select * from generate_series(1, 5, 2);
```
vrongmeal, line 24:
```suggestion

statement ok
create table t1 as select * from generate_series(1, 5, 2);

query I
select * from t1;
----
1
3
5
```

crates/sqlexec/src/context.rs
```diff
@@ -260,6 +263,34 @@ impl SessionContext {
...
+                .insert_into(self.get_df_state(), physical)
+                .await?;
+
+            let context = self.task_context();
+            let mut stream = execute_stream(physical, context)?;
+
+            // Drain the stream to actually "write" successfully.
+            while let Some(res) = stream.next().await {
+                let _ = res?;
+            }
```
vrongmeal, line 291: Can we add a helper function and call it both for `insert` and here?

vrongmeal (approved):
Looks good to me — just a couple of minor comments.

crates/sqlexec/src/context.rs
adhish20: sure. I will update that

vrongmeal (approved):
🚀

PR #1345, opened 2023-07-18
feat: Add `delta_scan` function

scsmithr:
```
〉select * from delta_scan('./testdata/delta/table1/') order by a;
+---+-------+
| a | b     |
+---+-------+
| 1 | hello |
| 2 | world |
+---+-------+
```

Supports querying delta tables from local paths, s3, and gcs. I kept the implementation separate from the other object store functions since constructing the accessor is a bit different for delta. Delta tables also only exist at one path, it doesn't make sense to support passing in more than one url.

Test data uploaded to s3 and gcs as appropriate. Currently the test data for delta consists of a single table made up of two parquet files and 3 json metadata logs.


universalmind303 (approved):

testdata/sqllogictests/functions/delta_scan.slt
```diff
@@ -0,0 +1,23 @@
...
+----
+1   hello
+2   world
+
+# # Relative path
+# query IT
+# select * from delta_scan('./testdata/delta/table1/') order by a;
+# ----
+# 1   hello
+# 2   world
```
vrongmeal, line 22: Does this not work yet?

crates/sqlbuiltins/src/functions/delta.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    async fn create_provider(
+        &self,
+        ctx: &dyn TableFuncContextProvider,
+        args: Vec<FuncParamValue>,
+        mut opts: HashMap<String, FuncParamValue>,
+    ) -> Result<Arc<dyn TableProvider>> {
+        let (loc, delta_opts) = match args.len() {
+            1 => {
+                let first = args.get(0).unwrap().clone();
```
vrongmeal, line 54:
```suggestion
                let args = args.into_iter();
                let first = args.next().unwrap();
```
Needless clone?

crates/sqlbuiltins/src/functions/delta.rs
```diff
@@ -0,0 +1,121 @@
...
+                            "Credentials required when accessing delta table in S3 or GCS",
+                        ))
+                    }
+                }
+            }
+            2 => {
+                let first = args.get(0).unwrap().clone();
+                let url = first.param_into()?;
+                let source_url = DatasourceUrl::new(&url)?;
+                let creds: IdentValue = args.get(1).unwrap().clone().param_into()?;
```
vrongmeal, line 71:
```suggestion
                let args = args.into_iter();
                let first = args.next().unwrap();
                let url = first.param_into()?;
                let source_url = DatasourceUrl::new(&url)?;
                let creds: IdentValue = args.next().unwrap().param_into()?;
```

vrongmeal (approved):
Just some minor comments. Looks good to 🚀

PR #1346, opened 2023-07-19
feat: Add pretty table formatting

scsmithr:
wip

```
〉select * from t1;
┌──────────┬──────────┐
│ a        │ b        │
│ Int32    │ Utf8     │
╞══════════╪══════════╡
│ 1        │ hello    │
│ 2        │ world    │
└──────────┴──────────┘
```


vrongmeal:
@scsmithr can we use the `Scalar` in `pgrepr` here? Will unite codec as well as local and will have to implement `Display` only once.

vrongmeal (approved):
This looks amazing 😍

I'll open a another of issue, so we can use `pgrepr::scalar::Scalar` instead of DF scalar. This is so we don't have to rework `Display` for all data types. I am hoping it should be a simple plug in 🤔

PR #1348, opened 2023-07-19
remove TableFunc::parameters

### LINKED ISSUE #1344: Chore: Remove `parameters` for `TableFunc`
## Description

https://github.com/glaredb/glaredb/blob/8ed5d57a000dc85b6ff77c444a6ea538f25e05f1/crates/sqlbuiltins/src/functions.rs#L80-L89

It's currently unused, and from some preliminary poking around, it seems like it'll be difficult to keep that in sync with positional/named arguments.

My original goal with it was for documentation purposes (e.g. querying the `functions` table), but it may be better to think about that _after_ we figure out the other stuff related to args.

universalmind303:
closes #1344


vrongmeal (approved):
Should we also remove `TableFuncParameters` and `TableFuncParameter` ?

scsmithr (approved):

PR #1353, opened 2023-07-20
feat: Support background jobs

### LINKED ISSUE #1167: Add background jobs
## Summary

A framework for running jobs in the background.

Some jobs we need:

- Clean up dropped tables. When dropping tables, it should be sufficient to just drop the table in the catalog, then submit a background job to actually delete from object storage.
- Compute table size. Similar to above. When a user inserts into a table, we should compute the table size following the insert. We may also want to try to "batch" these jobs to avoid a bunch of wasted work for tables that are frequently inserted into.
- Optimize table storage. Vacuum and repartition tables after some time.

Some design work may need to be done beforehand.

vrongmeal:
We have a framework to setup backgound jobs. Jobs with the same name are batched together. Each job can set a time when they start and until the existing job has finished, a new job is not added.

Fixes #1167

### DIFF
--- a/crates/glaredb/src/server.rs
+++ b/crates/glaredb/src/server.rs
@@ -100,17 +100,17 @@ impl Server {
 
         // Shutdown handler.
         let (tx, mut rx) = oneshot::channel();
-        let pg_hander = self.pg_handler.clone();
+        let pg_handler = self.pg_handler.clone();
         tokio::spawn(async move {
             match signal::ctrl_c().await {
                 Ok(()) => {
                     info!("shutdown triggered");
+                    let engine_shutdown = pg_handler.engine.shutdown();
+
                     loop {
-                        let sess_count = pg_hander.engine.session_count();
+                        let sess_count = pg_handler.engine.session_count();
                         if sess_count == 0 {
-                            // Shutdown!
-                            let _ = tx.send(());
-                            return;
+                            break;
                         }
 
                         info!(%sess_count, "shutdown prevented, active sessions");
@@ -119,6 +119,16 @@ impl Server {
                         // between.
                         tokio::time::sleep(tokio::time::Duration::from_millis(5000)).await;
                     }
+
+                    match engine_shutdown.await {
+                        Ok(()) => {}
+                        Err(err) => {
+                            error!(%err, "unable to shutdown the engine gracefully");
+                        }
+                    };
+
+                    // Shutdown!
+                    let _ = tx.send(());
                 }
                 Err(err) => {
                     error!(%err, "unable to listen for shutdown signal");
--- a/crates/sqlexec/src/background_jobs.rs
+++ b/crates/sqlexec/src/background_jobs.rs
@@ -1,0 +1,403 @@
+// FIXME: Remove this once we have real jobs.
+#![allow(dead_code)]
+
+use std::{
+    collections::HashMap,
+    fmt::{Debug, Write},
+    sync::Arc,
+    time::Duration,
+};
+
+use async_trait::async_trait;
+use parking_lot::Mutex;
+use tokio::{
+    sync::{mpsc, oneshot},
+    task::JoinHandle,
+    time::Instant,
+};
+use tracing::{debug, error, info};
+
+use crate::errors::{ExecError, Result};
+
+/// Collection of background jobs.
+pub struct JobRunner {
+    sender: mpsc::UnboundedSender<RequestMessage>,
+    listen: Arc<Mutex<Option<JoinHandle<()>>>>,
+}
+
+struct JobHandle {
+    join: JoinHandle<()>,
+    sleep: JoinHandle<()>,
+}
+
+#[derive(Debug)]
+enum RequestMessage {
+    NewJob(Box<dyn BgJob>),
+    JobComplete(String, Result<()>),
+    HealthCheck,
+    Close,
+}
+
+impl JobRunner {
+    /// Create a new collection of background jobs.
+    pub fn new() -> Self {
+        let (sender, receiver) = mpsc::unbounded_channel();
+        // Spawn the listener for receiving messages.
+        let sender_clone = sender.clone();
+        let listen = tokio::spawn(async move { Self::listen(sender_clone, receiver).await });
+        let listen = Arc::new(Mutex::new(Some(listen)));
+        Self { sender, listen }
+    }
+
+    /// Listens on the channel for new messages.
+    async fn listen(
+        sender: mpsc::UnboundedSender<RequestMessage>,
+        mut receiver: mpsc::UnboundedReceiver<RequestMessage>,
+    ) {
+        // Sends a "health check" mesage to the channel every two minutes.
+        let sender_clone = sender.clone();
+        tokio::spawn(async move {
+            let sender = sender_clone;
+
+            let tick_interval = if cfg!(test) {
+                Duration::from_secs(2)
+            } else {
+                Duration::from_secs(2 * 60)
+            };
+
+            let mut health_check = tokio::time::interval(tick_interval);
+            loop {
+                let _ = health_check.tick().await;
+                sender.send(RequestMessage::HealthCheck).unwrap();
+            }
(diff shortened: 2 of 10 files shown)


crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+}
+
+struct JobHandle {
+    join: JoinHandle<()>,
+    sleep: JoinHandle<()>,
+}
+
+enum ChannelMessage {
+    NewJob(BackgroundJob),
+    JobComplete(String, Result<()>),
```
vrongmeal, line 32: Should I add a new message, `JobStarted`, signalling, we can safely remove the job from the map. Can maintain a `running_jobs` map to mark jobs as started but incomplete. This allows us to add new jobs with the same name when an existing job has started. Not sure if it's for the better or worse.

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+
+pub enum BackgroundJobOptions {
+    Debug(BackgroundJobOptionsDebug),
+}
+
+pub struct BackgroundJobOptionsDebug {
+    pub batched: bool,
+    pub counter: Arc<AtomicUsize>,
+}
+
```
vrongmeal, line 211: Kept this pattern since we eventually will be serializing these options.

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+        // Sends a "health check" mesage to the channel every two minutes.
+        let send_clone = send.clone();
+        tokio::spawn(async move {
+            let send = send_clone;
+            let mut health_check = tokio::time::interval(Duration::from_secs(2 * 60));
+            loop {
+                let _ = health_check.tick().await;
+                send.send(ChannelMessage::HealthCheck).unwrap();
+            }
+        });
```
vrongmeal, line 62: A health check exists in case a job panics, and we aren't able to add a new job. This problem automatically disappears if we go with the abovementioned comment by maintaining a `running_jobs` map. But that comes with different properties.

vrongmeal (commented):
Currently the runner is `close`d only on the server. Need to add it to `local` run as well (and python?).

crates/glaredb/src/server.rs
```diff
@@ -100,17 +100,19 @@ impl Server {
...
                 Ok(()) => {
                     info!("shutdown triggered");
+                    let engine_shutdown = pg_handler.engine.shutdown();
+
                     loop {
-                        let sess_count = pg_hander.engine.session_count();
+                        let sess_count = pg_handler.engine.session_count();
                         if sess_count == 0 {
                             // Shutdown!
                             let _ = tx.send(());
```
vrongmeal, line 114: Send signal only when everything's ready (i.e. after the engine shutdown).

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+            BackgroundJobOptions::Debug(BackgroundJobOptionsDebug {
+                batched: false,
+                counter: counter.clone(),
+            }),
+        )
+        .unwrap();
+
+        // TODO: Add mechanism to wait for a certain job to finish. Sleeps lead
+        // to flaky tests.
+        tokio::time::sleep(Duration::from_millis(10)).await;
```
scsmithr, line 265: We could use a channel instead of an atomic, then await the recv side.

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
+use std::{
+    collections::HashMap,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc, Mutex,
```
scsmithr, line 5: I typically prefer to use the parking_lot mutex. Most of the times it's faster, and no need to `unwrap` after `lock`.

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+    task::JoinHandle,
+    time::Instant,
+};
+
+use crate::errors::{ExecError, Result};
+
+/// Collection of background jobs.
+pub struct JobRunner {
+    send: mpsc::UnboundedSender<ChannelMessage>,
+    listen: Arc<Mutex<Option<JoinHandle<()>>>>,
```
scsmithr, line 22: Not immediately obvious to me, why is this `Option`?

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+            let send = send_clone;
+            let mut health_check = tokio::time::interval(Duration::from_secs(2 * 60));
+            loop {
+                let _ = health_check.tick().await;
+                send.send(ChannelMessage::HealthCheck).unwrap();
+            }
+        });
+
+        let mut jobs: HashMap<String, JobHandle> = HashMap::new();
+        while let Some(msg) = recv.recv().await {
```
scsmithr, line 65: Could we add a debug log whenever we get a msg? I think that would be useful in case we need to debug if jobs are running, etc.

crates/sqlexec/src/background_jobs.rs
```diff
@@ -0,0 +1,326 @@
...
+            match msg {
+                ChannelMessage::NewJob(job) => {
+                    let job_name = job.name;
+
+                    if jobs.contains_key(&job_name) {
+                        continue;
+                    }
```
scsmithr, line 72: Is it expected that we'll have unique job names per session? E.g. what would happen if session 1 submits a job for dropping a table, and session 2 submits a job for dropping a table?

crates/sqlexec/src/background_jobs.rs
vrongmeal: Yeah, just need to add another set of channels here (for reverse communication).

crates/sqlexec/src/background_jobs.rs
vrongmeal: Will update using that.

crates/sqlexec/src/background_jobs.rs
vrongmeal: So we can `take` the value out when `close()` is called. Basically to get ownership over `JoinHandle`.

crates/sqlexec/src/background_jobs.rs
vrongmeal: Yup, on my list. Will add logs.

crates/sqlexec/src/background_jobs.rs
vrongmeal: I imagined jobs would be named consistently.

So dropping a table `tab` from any session would create a job `"drop_table_tab"`. If both sessions drop the same table, they'll be batched.

And unique job names are not per session; they are per engine. Maybe each job has a method `prefix()` or simply `name()` (that's created from options) so they are consistent, and it's not up to the user to provide a name.

scsmithr: Seems reasonable. Since everything pretty much has a UUID with it, we can just stick that in the job name, so no concerns here.

crates/sqlexec/src/background_jobs.rs
vrongmeal: @scsmithr what are your opinions on this?

vrongmeal: Gave it some thought, feels like the current implementation is fine. If a job is running and another job comes in that could've been batched just seconds ago, it's ok to let go of the job.

For example: if `"storage_calculation"` job is running, and another `"storage_calculation"` comes in, we already have the latest data being collected, so we can ignore the new job safely.

scsmithr: I'm down to keep this as is, and then as we start building out jobs, we can reevaluate if we need a new message.

scsmithr (approved):

PR #1354, opened 2023-07-20
chore: add better python ide support

universalmind303:
so i think we need to create wrappers for full IDE support, but this atleast gets it so it doesn't error out via pylint immediately. it also adds a new `fmt` command that'll format the python files in `py-glaredb/` 

note: it's not red anymore. 
image


scsmithr (approved):

adhish20:
LGTM

adhish20 (approved):

PR #1355, opened 2023-07-20
feat: support creating temp tables from source query

adhish20:
create temp table t1 as select * from generate_series(1, 5, 2);


vrongmeal (approved):
Good to 🚀

PR #1376, opened 2023-07-21
fix(datasources): disable reading from local filesystem on remote ins…

### LINKED ISSUE #1324: Disable using local datasource on the server

universalmind303:
…tances

closes #1324

I'm not really sure the best way to properly test this one out, but I've tested it by manually changing the boolean flag & it works as expected.

# Summary:

I had to move the `vars` into a new shared crate `datafusion_ext` as it was resulting in circular dependencies otherwise. The idea is that any native extensions for datafusion could go in here (such as the upcoming protobuf extensions).

other than that, there were a few other small changes to get `vars` to work with datafusion namely implementing `Clone` and adding a description. 

It clones the sessionvars when the session is created, so these are readonly at the moment. We can make them writable if the need arises. 

the actual validation logic happens in `scan` where we have access to `ctx`. I was thinking of passing `ctx` to some of the tableprovider functions, but it seemed to have a ripple effect that i thought would muddle this PR. So it will error when the initial scan occurs.

### DIFF
--- a/crates/datafusion_ext/Cargo.toml
+++ b/crates/datafusion_ext/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "datafusion_ext"
+description = "Shared datafusion extensions"
+version = { workspace = true }
+edition = { workspace = true }
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+datafusion = { workspace = true }
+uuid = { version = "1.4.0", features = ["v4", "fast-rng", "macro-diagnostics"] }
+regex = "1.8"
+once_cell = "1.18.0"
+tracing = "0.1"
+thiserror.workspace = true
--- a/crates/datafusion_ext/src/vars.rs
+++ b/crates/datafusion_ext/src/vars.rs
@@ -1,11 +1,11 @@
 //! Server and session variables.
-use crate::errors::{ExecError, Result};
 use datafusion::arrow::array::StringArray;
 use datafusion::arrow::datatypes::{DataType, Field, Schema};
 use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::config::{ConfigEntry, ConfigExtension, ExtensionOptions};
+use datafusion::error::{DataFusionError, Result};
 use once_cell::sync::Lazy;
 use regex::Regex;
-use sqlbuiltins::builtins::DEFAULT_SCHEMA;
 use std::borrow::{Borrow, ToOwned};
 use std::fmt::Display;
 use std::str::FromStr;
@@ -13,12 +13,31 @@ use std::sync::Arc;
 use tracing::error;
 use uuid::Uuid;
 
+#[derive(Debug, thiserror::Error)]
+pub enum ExecError {
+    #[error("Invalid value for session variable: Variable name: {name}, Value: {val}")]
+    InvalidSessionVarValue { name: String, val: String },
+
+    #[error("Variable is readonly: {0}")]
+    VariableReadonly(String),
+
+    #[error("Unknown variable: {0}")]
+    UnknownVariable(String),
+}
+
+impl From<ExecError> for DataFusionError {
+    fn from(e: ExecError) -> Self {
+        DataFusionError::Execution(e.to_string())
+    }
+}
+
 // TODO: Decide proper postgres version to spoof/support
 const SERVER_VERSION: ServerVar<str> = ServerVar {
     name: "server_version",
     value: "15.1",
     group: "postgres",
     user_configurable: false,
+    description: "Version of the server",
 };
 
 const APPLICATION_NAME: ServerVar<str> = ServerVar {
@@ -26,6 +45,7 @@ const APPLICATION_NAME: ServerVar<str> = ServerVar {
     value: "",
     group: "postgres",
     user_configurable: true,
+    description: "Name of the application",
 };
 
 const CLIENT_ENCODING: ServerVar<str> = ServerVar {
@@ -33,6 +53,7 @@ const CLIENT_ENCODING: ServerVar<str> = ServerVar {
     value: "UTF8",
     group: "postgres",
     user_configurable: true,
+    description: "Encoding of the client",
 };
 
 const EXTRA_FLOAT_DIGITS: ServerVar<i32> = ServerVar {
@@ -40,6 +61,7 @@ const EXTRA_FLOAT_DIGITS: ServerVar<i32> = ServerVar {
     value: &1,
     group: "postgres",
     user_configurable: true,
+    description: "Extra precision in float values",
 };
 
 const STATEMENT_TIMEOUT: ServerVar<i32> = ServerVar {
@@ -47,6 +69,7 @@ const STATEMENT_TIMEOUT: ServerVar<i32> = ServerVar {
     value: &0,
     group: "postgres",
     user_configurable: true,
+    description: "Statement timeout in milliseconds",
 };
 
 const TIMEZONE: ServerVar<str> = ServerVar {
@@ -54,6 +77,7 @@ const TIMEZONE: ServerVar<str> = ServerVar {
     value: "UTC",
     group: "postgres",
     user_configurable: true,
+    description: "Timezone of the client, default UTC",
 };
 
 const DATESTYLE: ServerVar<str> = ServerVar {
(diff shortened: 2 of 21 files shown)


crates/datafusion_ext/src/vars.rs
```diff
@@ -188,6 +235,78 @@ pub struct SessionVars {
...
+            user_id,
+            user_name,
+            database_name,
+            max_datasource_count,
+            memory_limit_bytes,
+            max_tunnel_count,
+            max_credentials_count,
+            is_cloud_instance
+        )
+    }
```
scsmithr, line 309: Do we need this anymore?

crates/datafusion_ext/src/vars.rs
```diff
@@ -1,81 +1,108 @@
...
-use sqlbuiltins::builtins::DEFAULT_SCHEMA;
 use std::borrow::{Borrow, ToOwned};
 use std::fmt::Display;
 use std::str::FromStr;
 use std::sync::Arc;
 use tracing::error;
 use uuid::Uuid;
 
+#[derive(Debug, thiserror::Error)]
+pub enum ExecError {
```
scsmithr, line 17:
```suggestion
pub enum VarError {
```

crates/datafusion_ext/src/vars.rs
```diff
@@ -361,10 +483,16 @@ pub enum VarSetter {
...
 #[allow(dead_code)] // `group` currently unused
-pub struct ServerVar<T: Value + ?Sized + 'static> {
+pub struct ServerVar<T>
+where
+    T: Value + ?Sized + 'static,
+{
     /// Name of the variable.
     name: &'static str,
 
+    /// Descriptino of the variable
```
scsmithr, line 493:
```suggestion
    /// Description of the variable
```

crates/datafusion_ext/src/vars.rs
```diff
@@ -524,6 +711,7 @@ impl Value for Uuid {
 
 impl<V> Value for Option<V>
 where
+    V: Value + ?Sized + 'static,
     V: Value + ToOwned + Clone + FromStr + Display,
```
scsmithr, line 715: Combine

crates/datasources/Cargo.toml
```diff
@@ -21,11 +21,14 @@ futures = "0.3.28"
...
 logutil = {path = "../logutil"}
 metastore_client = { path = "../metastore_client" }
+datafusion_ext = { path = "../datafusion_ext" }
 mongodb = "2.6.0"
 mysql_async = { version = "0.32.2", default-features = false, features = ["default-rustls"] }
 mysql_common = { version = "0.30.6", features = ["chrono"] }
 object_store = { workspace = true, features = ["gcp", "aws", "http"] }
 object_store_util = { path = "../object_store_util" }
+
+
```
scsmithr, line 31:
```suggestion
```

crates/datafusion_ext/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+datafusion = { workspace = true }
+uuid = { version = "1.4.0", features = ["v4", "fast-rng", "macro-diagnostics"] }
+regex = "1.8"
+once_cell = "1.18.0"
+tracing = "0.1"
+thiserror.workspace = true
```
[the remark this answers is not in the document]
universalmind303: i'll take a look & see if we can move this into another crate.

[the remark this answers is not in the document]
universalmind303: so i was examining the dependencies and we _could_ definitely move it to `sqlbuiltins`, but i don't really think of the session variables as sql builtins & have really nothing to do with sql, but with the session context. So while we _could_, i don't think we **should** move it into there. 


one suggestion

rename `datafusion_planner` to datafusion_ext` & move the sessionvar stuff into there. 
This would require moving a few other components from sqlbuiltins into datafusion_ext as well to break the dependency. 


alternatively, we could just move it in to `sqlbuiltins` as is. 

I'm in favor of moving it into a datafusion extension crate, but i'd like to get your opinion before moving ahead. 

@scsmithr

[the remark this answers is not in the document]
scsmithr: > rename `datafusion_planner` to datafusion_ext` & move the sessionvar stuff into there. This would require moving a few other components from sqlbuiltins into datafusion_ext as well to break the dependency.
> 
> alternatively, we could just move it in to `sqlbuiltins` as is.
> 
> I'm in favor of moving it into a datafusion extension crate, but i'd like to get your opinion before moving ahead.

I'm down with moving the datafusion planner stuff.

universalmind303:
so i refactored the crates as mentioned above

## previous
```mermaid
flowchart LR
    datafusion_ext(datafusion_ext)
    pgsrv --> glaredb
    metastore_client ----> datasources
    metastore_client --> glaredb
    metastore_client --> sqlbuiltins
    metastore_client --> sqlexec
    metastore_client --> metastore
    sqlexec --> glaredb
    sqlexec --> pgsrv
    datafusion_planner --> sqlexec
    sqlbuiltins --> datafusion_planner
    sqlbuiltins ---> sqlexec
    datafusion_ext ---------> glaredb
    datafusion_ext --> pgsrv
    datafusion_ext --> sqlbuiltins
    datafusion_ext --> sqlexec
    sqlbuiltins --> metastore
    metastore --> glaredb
    metastore --> sqlexec
    datasources --> sqlbuiltins
    datasources --> sqlexec
```
    
## current
```mermaid
flowchart LR
    datafusion_ext(datafusion_ext)
    pgsrv --> glaredb
    metastore_client ----> datasources
    metastore_client --> glaredb
    metastore_client --> sqlbuiltins
    metastore_client --> sqlexec
    metastore_client --> metastore
    metastore_client --> datafusion_ext
    sqlexec --> glaredb
    sqlexec --> pgsrv
    sqlbuiltins ---> sqlexec
    datafusion_ext ---------> glaredb
    datafusion_ext --> pgsrv
    datafusion_ext --> sqlbuiltins
    datafusion_ext --> sqlexec
    sqlbuiltins --> metastore
    metastore --> glaredb
    metastore --> sqlexec
    datasources --> sqlbuiltins
    datasources --> sqlexec
```

scsmithr (approved):
Let's go with this for now. I think there's some reorganizing of crates we need to do, but it's more of a gut feeling I have, nothing concrete.

For testing this, we can follow up with a PR that implements a hook that sets this variable accordingly.

universalmind303:
> Let's go with this for now. I think there's some reorganizing of crates we need to do, but it's more of a gut feeling I have, nothing concrete.

i had the same feeling. I didn't want to go too overboard with reorganizing, but I agree that there is further room for improvement.

universalmind303:
@scsmithr so this is interesting.. it looks like our slts run as if it is in cloud mode. Do the slt's use an embedded client, or is it being passed through `pgsrv`?

crates/pgsrv/src/handler.rs
```diff
@@ -252,6 +252,7 @@ impl ProtocolHandler {
             .set_and_log(Some(max_tunnel_count), VarSetter::System);
         vars.max_credentials_count
             .set_and_log(Some(max_credentials_count), VarSetter::System);
+        vars.is_cloud_instance.set_and_log(true, VarSetter::System);
```
scsmithr, line 255: So this will set the variable to true when running as a server (e.g. via `glaredb server`). Our slts connect to this since it'll spin up an in-process server, then use the postgres client to connect.

However, we'll be able to tell if the connection has been proxied through the existence of some parameters that get set by pgsrv. For example, we could do something like this:

```rust
// Determine if we're running as a cloud instance by trying to 
// read a non-standard parameter that gets set by pgsrv.
let is_cloud_instance = params.get(&GLAREDB_DATABASE_ID_KEY.key).is_ok();
```

This would require that we make the `key` field public.

scsmithr:
> @scsmithr It looks like our slts run as if it is in cloud mode. It looks like they are using `pgsrv` which i thought was made for cloud only.
> 
> the current implementation hardcodes the `is_cloud_server` if it's being started from `pgsrv`. I think we may need to set that variable elsewhere.

Left comment about how this can be resolved.

For reference:

- `pgsrv` the package provides the Postgres wire protocol implementation and handler. When starting `glaredb server` (or running an SLT test), a `ProtocolHandler` gets instantiated.
- `pgsrv` the service provides the proxying between users and destination databases. The proxy instantiates `ProxyHandler` that sets a bunch of parameters for the connection.

crates/pgsrv/src/handler.rs
scsmithr: I _think_ that should work

PR #1398, opened 2023-07-25
feat: Add background job to track storage usage

### LINKED ISSUE #1155: Track storage usage
## Summary

And expose through system table.

Currently Cloud is planned to check storage usage in the buckets, but longer term that should be handled in glaredb to avoid needlessly having to fetch bucket info.

vrongmeal:
Runs a background job on `INSERT`s (with a batch delay of 5 minutes) for the deployment.

```
〉select * from glare_catalog.deployment_metadata;
+--------------+-------+
| key          | value |
+--------------+-------+
| storage_size | 3852  |
+--------------+-------+
```

Fixes #1155

### DIFF
--- a/crates/sqlexec/src/background_jobs.rs
+++ b/crates/sqlexec/src/background_jobs.rs
@@ -1,12 +1,6 @@
-// FIXME: Remove this once we have real jobs.
-#![allow(dead_code)]
+pub mod storage;
 
-use std::{
-    collections::HashMap,
-    fmt::{Debug, Write},
-    sync::Arc,
-    time::Duration,
-};
+use std::{collections::HashMap, fmt::Debug, sync::Arc, time::Duration};
 
 use async_trait::async_trait;
 use parking_lot::Mutex;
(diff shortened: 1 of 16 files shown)


crates/sqlexec/src/background_jobs.rs
```diff
@@ -15,33 +9,60 @@ use tokio::{
...
         let (sender, receiver) = mpsc::unbounded_channel();
+
+        // Sends a "health check" mesage to the channel.
+        let sender_clone = sender.clone();
+        tokio::spawn(async move {
+            let sender = sender_clone;
+
+            std::panic::set_hook(Box::new(|_| {
+                warn!("health checker exited");
+            }));
```
scsmithr, line 57: This will replace the existing (global) hook. I think we should remove this and instead just log in the loop below if we get an error.

scsmithr (approved):
lgtm, just a small comment about the panic hook

crates/sqlexec/src/background_jobs.rs
vrongmeal: Updated!

PR #1402, opened 2023-07-26
feat: Support glob patterns for local file scans

### LINKED ISSUE #1396: can't read a folder of parquet files
it seems multiple files are not supported

```
import glaredb 
con = glaredb.connect()
lineitem = con.sql(f""" select * FROM parquet_scan('./1/lineitem/*.parquet') """)
```
Thanks

vrongmeal:
Some cleanups included:
* Removing `super::*` from functions since it was too messy
* Un-aliasing `BuiltinError` to `ExtensionError`
* Updating errors to be more diverse

Fixes #1396


crates/sqlbuiltins/src/functions/object_store.rs
```diff
@@ -35,21 +51,46 @@ impl TableFunc for ObjScanTableFunc {
...
+                    .map_err(|e| ExtensionError::Access(Box::new(e)))?;
+
+                    let mut urls = Vec::new();
+                    for path in paths {
+                        let path = path.map_err(|e| ExtensionError::Access(Box::new(e)))?;
+                        let url = DatasourceUrl::File(path);
+                        urls.push(url);
+                    }
+                    urls
+                }
```
vrongmeal, line 83: For ease of review, the code concerning `glob` lies here.

scsmithr (approved):
This should work for a first pass. I do think we'll need to look at pushing this down into the object store execs themselves instead of unioning the plans at this level.

Lets keep the linked issue open too since I'd want this working for S3/GCS as well before closing.

vrongmeal:
> I do think we'll need to look at pushing this down into the object store execs themselves instead of unioning the plans at this level.

I agree. I'll open a different issue for this (or link to an existing one for partitioned files).

> Lets keep the linked issue open too since I'd want this working for S3/GCS as well before closing.

I did think about this. It would require us to figure out the `prefix` from the store and then filter files using the pattern. Listing files won't be cheap in many cases (say path is `*_abc.csv` where `prefix` is empty). Do we want to go down that route? I suppose we can support it and list the limitations (using glob on object store might not be fast)?

I'll merge this in and re-open the issue for S3/GCS

greyscaled:
Nice one @vrongmeal will be awesome for `v0.3.0`, and maybe we even get full object storage support too

PR #1410, opened 2023-07-27
feat: Support `glob` patterns in for S3/GCS

### LINKED ISSUE #1396: can't read a folder of parquet files

vrongmeal:
This is a big change:

* Introduces `ObjStoreAccess` and `ObjStoreAccessor` making the code much more generic for object store and easy to configure.
* Introduces `FileTypeAccess` for different object files.

Refactored scan functions using the changes.

### Example for globs:

Following query will get all objects with `some_prefix/` as prefix matching the given pattern. Valid matches would be:
* `some_prefix/abc/obj_1_pattern.csv`
* `some_prefix/def/xyz/obj_something_pattern.csv`

```
SELECT * FROM csv_scan('gs://bucket/some_prefix/**/obj_*_pattern.csv');
```

**TODO:**

- [x] Use `ObjStoreAccessor` for external tables.
- [ ] Support hive partitioning

Fixes #1396


crates/datasources/src/object_store/mod.rs
```diff
@@ -31,35 +48,275 @@ impl FromStr for FileType {
...
+            limit,
+            table_partition_cols: Vec::new(),
+            output_ordering: Vec::new(),
+            infinite_source: false,
+        };
+
+        // We register the store at scan time so that it can be used by the
+        // exec plan.
+        ctx.runtime_env()
+            .register_object_store(self.base_url.as_ref(), self.store.clone());
```
scsmithr, line 266: Since we're now doing this here, I _believe_ we should be able to remove https://github.com/glaredb/glaredb/blob/904db3a14448568a41ea34c5857f7197c31bb3d2/crates/sqlexec/src/planner/dispatch.rs#L434-L437 (and the one for s3 too).

crates/datasources/src/object_store/mod.rs
```diff
@@ -31,35 +48,275 @@ impl FromStr for FileType {
...
+            }),
+            TableOptions::S3(opts) => Arc::new(S3StoreAccess {
+                region: opts.region.clone(),
+                bucket: opts.bucket.clone(),
+                access_key_id: opts.access_key_id.clone(),
+                secret_access_key: opts.secret_access_key.clone(),
+            }),
+            // Continue on all others. Explicityly mentioning all the left
+            // over options so we don't forget adding object stores that are
+            // supported in the furure (like azure).
```
scsmithr, line 306: 👍

crates/datasources/src/object_store/mod.rs
```diff
@@ -31,35 +48,275 @@ impl FromStr for FileType {
...
+}
+
+#[async_trait]
+pub trait ObjStoreAccess: Debug + Display + Send + Sync {
+    /// Returns the base URL against which the object store is registered.
+    ///
+    /// Examples:
+    /// * `s3//bucket/path/to/file.csv`: `s3://bucket`
+    /// * `/some/local/file`: `file://`
+    fn base_url(&self) -> Result<ObjectStoreUrl>;
```
scsmithr, line 91: Could we add a quick example for http too since that's a bit special? I think that'll make it clear that `ObjectStoreUrl` here is acting as a key to the registry and not being used to build up the actual url to fetch the object.

scsmithr (approved):
🚢

crates/datasources/src/object_store/mod.rs
scsmithr: nvm, just saw that it's been removed. Missed that on my first pass.

PR #1418, opened 2023-07-28
ci: update scripts to only run integration tests when owner = glaredb

### LINKED ISSUE #1323: Chore: Only run action steps that use secrets when `github.repository == glaredb/glaredb`
## Description

Currently we're kinda stuck with our integration tests. Github actions won't use secrets when a PR is opened from a public fork even after it's been manually approved.

We still want to run:

- Unit tests
- Clippy/doc tests
- SLTs that don't depend on secrets

We'll want to skip integration tests like connecting to snowflake, etc

universalmind303:
closes #1323

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -44,23 +44,32 @@ jobs:
 
       - name: Authenticate to Google Cloud
         uses: google-github-actions/auth@v1
+        if: github.repository_owner == 'glaredb'
         with:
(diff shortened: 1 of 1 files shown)


universalmind303:
I'd like to follow up with this one a bit to refactor the ci for the tests. I think we could likely run quite a few more tests publicly, but it'd involve a bit of restructuring of the ci and `testdata`

scsmithr (approved):

PR #1423, opened 2023-07-28
chore: add symlink to parent license in py-glaredb

### LINKED ISSUE #1349: add LICENSE to py-glaredb
py-glaredb should include a license that is visible on [pypi](https://pypi.org/project/glaredb/). 

@scsmithr should we use a MIT/Apache license for this, or the same AGPL license as the core library?

universalmind303:
just adds a symlink to the root LICENSE. I figured this'll be much easier to keep in sync than copy/paste. 

closes #1349


scsmithr:
Is this good to go?

scsmithr (approved):

PR #1429, opened 2023-07-30
feat: Split optimizer/physical planner from session state

scsmithr:
wip

### DIFF
--- a/crates/pgrepr/src/scalar.rs
+++ b/crates/pgrepr/src/scalar.rs
@@ -280,12 +280,14 @@ fn get_naive_date_time_nano(nanos: i64) -> NaiveDateTime {
     Utc.timestamp_nanos(nanos).naive_utc()
 }
 
+// TODO: Figure out if this should be parsing time zone names like
+// 'Australia/Melbourne' or offsets like '+03:00'.
 fn get_timezone(tz: &str) -> Tz {
     // TODO: Make a map at compile time to use (to speed this up).
     *TZ_VARIANTS
         .iter()
         .find(|&v| v.name() == tz)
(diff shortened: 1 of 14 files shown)


crates/pgrepr/src/scalar.rs
```diff
@@ -280,12 +280,14 @@ fn get_naive_date_time_nano(nanos: i64) -> NaiveDateTime {
...
 
+// TODO: Figure out if this should be parsing time zone names like
+// 'Australia/Melbourne' or offsets like '+03:00'.
 fn get_timezone(tz: &str) -> Tz {
     // TODO: Make a map at compile time to use (to speed this up).
     *TZ_VARIANTS
         .iter()
         .find(|&v| v.name() == tz)
-        .unwrap_or_else(|| panic!("invalid timezone in scalar value: {tz}"))
+        .unwrap_or(&chrono_tz::UTC)
```
scsmithr, line 290: @vrongmeal I was hitting the panic here. I was testing with fixed offset time zones in sql (e.g. "+03:00"). Things seem to work with this, but I'm not sure what we're going for with this.

scsmithr: @greyscaled this was the cause of the disconnect on qa.

vrongmeal: The `TZ_VARIANTS` basically has a list of valid timezone abbreviations (like `"UTC"`, `"IST"`, etc.) It doesn't parse `+03:00`, so this panic shouldn't exist. For now, we should `unwrap_or(&chrono_tz::UTC)` and create an issue for this.

vrongmeal (approved):

PR #1437, opened 2023-07-31
fix: Run delete table jobs after tables are removed from catalog

scsmithr:
Otherwise the user could end up in a situation where table data was deleted, but the table still exists in the catalog if the request to mutate the catalog failed.


testdata/sqllogictests/drop.slt
```diff
@@ -292,3 +292,26 @@ drop schema glare_catalog;
...
+1
+
+statement error
+drop table drop_data_check, table_does_not_exist;
+
+# Since the above query fails due to a table not existing, we should _not_ drop
+# any data.
+#
+# Note that this doesn't guarantee that we drop data in the correct order since
+# we don't have a guarantee on when the background job runs.
```
vrongmeal, line 313: We can't really guarantee that the data isn't dropped here either. Maybe the background job hasn't still run 🤔

Does this test error on main?

scsmithr: Yeah this was the easiest test case to write, but definitely not comprehensive.

I was able to get this to fail on main. It was easier to repeat when running locally:

```
GlareDB (v0.2.1)
Type \help for help.
Using in-memory catalog
〉\open db
Created new session. New database path: db
〉create table drop_data_check as values (1);
create table
〉drop table drop_data_check, table_does_not_exist;
Error: Failed Metastore request: Missing database object; schema: public, name: table_does_not_exist
〉select * from drop_data_check;
Error: Error during planning: Unable to fetch table provider for 'drop_data_check': Failed to create table provider for 'drop_data_check': Not a Delta table: No snapshot or version 0 found, perhaps /Users/sean/Code/github.com/glaredb/glaredb/db/databases/00000000-0000-0000-0000-000000000000/tables/20000/ is an empty dir?
```

adhish20 (approved):

PR #1443, opened 2023-08-01
fix: display correct number of columns based on terminal width

### LINKED ISSUE #1440: Not all columns shown even though terminal is wide enough when running locally
## Context
Screenshot 2023-07-31 at 11 12 16 PM

Getting "..."

universalmind303:
closes #1440

So the logic is as follows. 

```mermaid
flowchart TD

A[Start] --> B{max_columns is Some?}
B -- Yes --> C[Use max_columns]
B -- No --> D{width is Some?}

D -- Yes --> E[Calculate avg_column_len]
D -- No --> F[Use DEFAULT_MAX_COLUMNS]

E --> G["Use inferred max columns (width - 10) / avg_column_len"]
```


scsmithr (approved):
I like the mermaid chart

Issue #1350, opened 2023-07-19
Chore: Add additional examples for connecting to glaredb from other languages


scsmithr:
## Description

We already have R: https://github.com/GlareDB/glaredb/blob/main/examples/R/out-of-process.R

Let's see if we can get Go and JavaScript/Typescript as well.

greyscaled:
I've got a Golang example in flight here: https://github.com/GlareDB/glaredb/pull/1452

Would still be cool to see a `Node.js` example as well.

We could also show JDBC/Java?

f0ssel:
I'll throw up a node one

PR #1452, opened 2023-08-01
chore: Golang example

greyscaled:
Progress towards: #1350


scsmithr (approved):
This looks good. Could we change the directory to `go` instead of `Go`. The `R` directory is capitalized to make it more recognizable (since R files end with `.R`).

greyscaled:
> This looks good. Could we change the directory to `go` instead of `Go`. The `R` directory is capitalized to make it more recognizable (since R files end with `.R`).

for sure

PR #1453, opened 2023-08-01
chore: gracefully handle bucket field in dev

f0ssel:
In cloud development we provide this field but run metastore in memory mode, so we get this error.


scsmithr (approved):
Session configuration a bit gnarly right now.

PR #1458, opened 2023-08-02
chore: Add simple typescript example

f0ssel:
Closes https://github.com/GlareDB/glaredb/issues/1350


examples/typescript/example.ts
```diff
@@ -0,0 +1,27 @@
+import { Client } from "pg";
+
+async function main() {
+  const localHostStr =
+    "postgresql://user:pass@my_org.proxy.glaredb.com:6543/my_db"; // Replace with your connection string
```
greyscaled, line 5: Are we able to make this runnable like the Go one?

f0ssel: Done here: https://github.com/GlareDB/glaredb/pull/1461

PR #1465, opened 2023-08-02
tip: support deletes

### LINKED ISSUE #1352: Support deletes
## Description

```
DELETE FROM table_name WHERE a > 5;
```

## Rationale

- Hels round out native tables
  - User ask
- Storage/billing related (see: #1155)

adhish20:
Fixes #1352

### DIFF
--- a/crates/sqlexec/src/planner/session_planner.rs
+++ b/crates/sqlexec/src/planner/session_planner.rs
@@ -4,12 +4,14 @@ use std::sync::Arc;
 use datafusion::arrow::datatypes::{
     DataType, Field, TimeUnit, DECIMAL128_MAX_PRECISION, DECIMAL_DEFAULT_SCALE,
 };
-use datafusion::common::{OwnedSchemaReference, OwnedTableReference};
-use datafusion::sql::planner::{object_name_to_table_reference, IdentNormalizer};
+use datafusion::common::{OwnedSchemaReference, OwnedTableReference, ToDFSchema};
+use datafusion::datasource::DefaultTableSource;
+use datafusion::sql::planner::{object_name_to_table_reference, IdentNormalizer, PlannerContext};
 use datafusion::sql::sqlparser::ast::AlterTableOperation;
 use datafusion::sql::sqlparser::ast::{self, Ident, ObjectName, ObjectType};
 use datafusion::sql::TableReference;
 use datafusion_ext::planner::SqlQueryPlanner;
+use datafusion_ext::AsyncContextProvider;
 use datasources::bigquery::{BigQueryAccessor, BigQueryTableAccess};
 use datasources::common::ssh::{key::SshKey, SshConnection, SshConnectionParameters};
 use datasources::common::url::{DatasourceUrl, DatasourceUrlType};
@@ -592,13 +594,13 @@ impl<'a> SessionPlanner<'a> {
 
     async fn plan_statement(&self, statement: ast::Statement) -> Result<LogicalPlan> {
         let mut context_provider = PartialContextProvider::new(self.ctx)?;
-        let mut planner = SqlQueryPlanner::new(&mut context_provider);
         match statement {
             ast::Statement::StartTransaction { .. } => Ok(TransactionPlan::Begin.into()),
             ast::Statement::Commit { .. } => Ok(TransactionPlan::Commit.into()),
             ast::Statement::Rollback { .. } => Ok(TransactionPlan::Abort.into()),
 
             ast::Statement::Query(q) => {
+                let mut planner = SqlQueryPlanner::new(&mut context_provider);
                 let plan = planner.query_to_plan(*q).await?;
                 Ok(LogicalPlan::Query(plan))
             }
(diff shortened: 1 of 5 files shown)


adhish20:
Documentation PR : https://github.com/GlareDB/glaredb.github.io/pull/134

crates/sqlexec/src/context.rs
```diff
@@ -336,6 +336,22 @@ impl SessionContext {
...
+
+        if let Some(table_entry) = self
+            .metastore_catalog
+            .resolve_native_table(&database, &schema, &name)
+        {
+            self.tables
+                .delete_rows_where(table_entry, plan.expr)
+                .await?
+        } else {
+            return Err(ExecError::Internal("Table does not exist".to_string()));
```
vrongmeal, line 350:
```suggestion
            return Err(ExecError::MissingObject {
                typ: "table",
                name: plan.table_name,
            });
```

testdata/sqllogictests/delete.slt
```diff
@@ -0,0 +1,22 @@
...
+3
+5
+7
+
+statement ok
+delete from t1;
+
+query I
+select * from t1
+----
```
vrongmeal, line 22: Can we add more tests here?

* Error when no table given/many tables given
* Some more complex where clauses

vrongmeal (commented):
A couple of comments, but this looks good overall.

testdata/sqllogictests/delete.slt
adhish20: Added the tests

crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -947,6 +953,61 @@ impl<'a> SessionPlanner<'a> {
...
+                returning: None,
+            } if tables.is_empty() => {
+                let (table_name, schema) = match from.len() {
+                    0 => {
+                        return Err(PlanError::InvalidDeleteStatement {
+                            msg: "DELETE FROM should have atleast one table name",
+                        })
+                    }
+                    1 => {
+                        let table_factor = from[0].relation.clone();
```
vrongmeal, line 974: We can avoid this clone here easily. Since there's just one element in the array, and it's not used again, you can:

```suggestion
                        let table_factor = from.pop().relation;
```

crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -947,6 +953,61 @@ impl<'a> SessionPlanner<'a> {
...
+                        };
+                        validate_object_name(&table_name)?;
+                        let table_name = object_name_to_table_ref(table_name)?;
+
+                        let table_source = context_provider
+                            .get_table_provider(table_name.clone())
+                            .await?;
+                        let table_source: &DefaultTableSource =
+                            table_source.as_any().downcast_ref().unwrap();
+                        let schema = table_source.table_provider.schema().to_dfschema()?;
```
vrongmeal, line 991: Since we just need the schema, we can avoid the downcast here:

```suggestion
                        let table_source = context_provider
                            .get_table_provider(table_name.clone())
                            .await?;
                        let schema = table_source.schema().to_dfschema()?;
```

crates/sqlexec/src/planner/session_planner.rs
adhish20: from is immutable here

vrongmeal (approved):

crates/sqlexec/src/planner/session_planner.rs
vrongmeal: You can add `mut` in front of `from`. It's owned here so you should be able to do anything with it:

```rust
            ast::Statement::Delete {
                 // ...
                 mut from,
                 // ...
             } if tables.is_empty() => {
```

universalmind303:
@adhish20 Just curious, what happens if you try to delete from an external table? does it panic, or does it error with an appropriate error message?

adhish20:
It will throw an error that the table is not found

adhish20:
> @adhish20 Just curious, what happens if you try to delete from an external table? does it panic, or does it error with an appropriate error message?

It will throw an error that the table is not found

crates/sqlexec/src/context.rs
```diff
@@ -336,6 +336,22 @@ impl SessionContext {
...
+
+        if let Some(table_entry) = self
+            .metastore_catalog
+            .resolve_native_table(&database, &schema, &name)
+        {
+            self.tables
+                .delete_rows_where(table_entry, plan.expr)
+                .await?
+        } else {
+            return Err(ExecError::MissingObject { typ: "table", name });
```
scsmithr, line 350: Minor, but would prefer an error that says something along the lines of "Deleting from external tables not currently supported"

crates/sqlexec/src/session.rs
```diff
@@ -57,6 +57,8 @@ pub enum ExecutionResult {
     Commit,
     /// Transaction rolled abck.
     Rollback,
+    /// Data successfully deleted.
+    DeleteSuccess,
```
scsmithr, line 61: (future): We'll want to also return how many rows were deleted so that we can show that to the user.

scsmithr (approved):
Very minor stuff, good with this being merged as-is

PR #1469, opened 2023-08-03
wip support compressed files

universalmind303:
So i still have a lot of cleanup on this to do. But in summary, the compressed formats weren't working with our scan providers, even after adding in the support. I was still getting some errors with the encoding. So I started looking in to what datafusion already provides. In doing this, I realized almost all of our object store table providers (with the exception of http) can be replaced with datafusions providers via the ListingTable. 

There's a bit of finessing to get them wired up correctly, but I think going this route can dramatically simplify parts of the codebase & give us much less code to maintain.


crates/glaredb/src/local.rs
```diff
@@ -79,8 +91,6 @@ impl LocalClientOpts {
                 "\\mode MODE",
                 "Set the output mode [table, json, ndjson, csv]",
             ),
-            ("\\max-rows NUM", "Max number of rows to display"),
-            ("\\max-columns NUM", "Max number of columns to display"),
```
scsmithr, line 83: We still want these.

crates/glaredb/src/local.rs
```diff
@@ -296,11 +308,11 @@ impl LocalSession {
                 println!("Created new session. New database path: {path}");
                 *self = new_sess;
             }
-            ("\\quit", None) => return Ok(ClientCommandResult::Exit),
+            ("\\quit", None) => std::process::exit(0),
```
scsmithr, line 311: This prevents history from being saved, hence the `ClientCommandResult`

crates/glaredb/src/local.rs
```diff
@@ -284,8 +298,6 @@ impl LocalSession {
                 self.opts.mode = OutputMode::from_str(val, true)
                     .map_err(|s| anyhow!("Unable to set output mode: {s}"))?;
             }
-            ("\\max-rows", Some(val)) => self.opts.max_rows = Some(val.parse()?),
-            ("\\max-columns", Some(val)) => self.opts.max_rows = Some(val.parse()?),
```
scsmithr, line 288: Still want these.

crates/glaredb/src/local.rs
universalmind303: will update this file. Was a bad merge.

universalmind303:
most of the sources are working now. a few left to resolve
| Scheme | Glob | Compression | Complete |
|--------|------|-------------|----------|
| HTTP   |      |             | ✅        |
| HTTP   | ✔️   |             | N/A        |
| HTTP   |      | ✔️           | ✅        |
| HTTP   | ✔️   | ✔️           | N/A        |
| File   |      |             | ✅        |
| File   | ✔️   |             | ✅        |
| File   |      | ✔️           | ✅        |
| File   | ✔️   | ✔️           | ✅        |
| GS     |      |             | ✅        |
| GS     | ✔️   |             | ✅        |
| GS     |      | ✔️           | ✅         |
| GS     | ✔️   | ✔️           | ❌        |
| S3     |      |             | ❌       |
| S3     | ✔️   |             | ❌        |
| S3     |      | ✔️           | ❌        |
| S3     | ✔️   | ✔️           | ❌        |

scsmithr:
Pinging @vrongmeal.

There's changes here super familiar with, will need help on making sure this works for us.

universalmind303:
@vrongmeal I can't figure out why the parquet readers aren't working anymore? it seems like they are trying to do some extra work that the csv & json readers arent. Did you run into this with the original implementation?

vrongmeal:
I did remove the old parquet reader factory since we were using the default registry. Didn't face any issues with it. What's the issue you're facing right now?

vrongmeal:
I am guessing, introducing a custom registry breaks the parquet stuff. We did configure parquet reader because of the same.

universalmind303:
closing in favor of #1490

PR #1476, opened 2023-08-03
add datafusion & arrow to readme acknowledgements

universalmind303:
not sure how i feel about the images. LMK what you think. I can remove them if they are too much.


README.md
```diff
@@ -122,3 +122,12 @@ this repository.
 [Snowflake]: https://docs.glaredb.com/docs/data-sources/supported/snowflake.html
 [Postgres]: https://docs.glaredb.com/docs/data-sources/supported/postgres.html
 [Supported data sources]: https://docs.glaredb.com/docs/data-sources/supported/
+
+## Acknowledgements
+
+![https://arrow.apache.org/datafusion/](https://arrow.apache.org/datafusion/_images/DataFusion-Logo-Background-White.png)
+
+
+![https://arrow.apache.org/](https://arrow.apache.org/img/arrow-inverse.png)
```
greyscaled, line 131: This image is very hard to see at least in "light" mode, which is what I use for GH:

Acknowledgements

I'm kind of wondering if we can mention it in the "About" of the README and use smaller images. One way of doing that would be searching for images that match our needds, or just using a straight up `<image>` html tag as long as the resolution didn't suffer from scaling.

@scsmithr how do you feel about moving to About ?

scsmithr: I'm fine with this acknowledgment section, but let's just skip the images for now.

greyscaled (approved):

PR #1499, opened 2023-08-06
feat: rpcsrv

scsmithr:
# Description

Adds in an `rpcsrv` crate for facilitating remote execution of query plans. Currently this only supports logical plans generated by datafusion, and we send the _entire_ plan always.

The goal here is to get everything ready for us to start getting into the details of splitting plans up, and making our plans serializable (whether that's by implementing our plans as datafusion plans or something else).

Essentially what's happening is we have two sessions running, one locally, and one on the remote service. Query execution will go through the following steps:

1. Query gets serialized to logical plan
2. Query wrapped in a `RemoteLogicalExec`
3. `RemoteLogicalExec` gets executed, which sends the plan to the remote service.
4. The remote service receives this plan and turns it into a physical plan.
5. The physical plan gets executed, with results being streamed back to the local session.

All communication between local and remote sessions is over grpc, and Arrow batches are serialized to IPC.

Follow up steps include:

- Proxy for authenticating and routing
- Support our custom plans
- Nice interface for connecting through glaredb local and python

This PR is just a start, and I expect a lot of issues to be created out of this.

# Running

In a terminal, run:

```
cargo run --bin glaredb -- -v server --rpc-bind 0.0.0.0:6443
```

In another terminal, run:

```
RPC_HOST_URL="http://localhost:6443" cargo run --bin glaredb -- local
```

Then run a sql command:

```
> select * from (values (1, 2), (3, 4));
┌─────────┬─────────┐
│ column1 │ column2 │
│ ──      │ ──      │
│ Int64   │ Int64   │
╞═════════╪═════════╡
│ 1       │ 2       │
│ 3       │ 4       │
└─────────┴─────────┘
```

In the first terminal, you should have seen an "executing" log.

Note that the number of commands supported right now is pretty low since we have a lot of custom stuff that isn't being serialized.


universalmind303:
server does not shut down on sigterm (ctrl+C) or sigquit (ctrl+D). 

image

scsmithr:
> server does not shut down on sigterm (ctrl+C) or sigquit (ctrl+D).

"Expected" right now, remote sessions are never closed, will open issue.

universalmind303:
one thing i'd like to see _(in a follow up PR)_ is a more holistic way to set the variables. 

> In a terminal, run:
`cargo run --bin glaredb -- -v server --rpc-bind 0.0.0.0:6443`
In another terminal, run:
`RPC_HOST_URL="http://localhost:6443" cargo run --bin glaredb -- local`


It seems like there is a bit of a mix of env vars & cli args. We can clean this up across the board. IMO, everything that can be set via a cli arg should also be settable via env vars & vice-versa. We could also expose a config file with all of the same arguments.

scsmithr:
> one thing i'd like to see _(in a follow up PR)_ is a more holistic way to set the variables.
> 
> > In a terminal, run:
> > `cargo run --bin glaredb -- -v server --rpc-bind 0.0.0.0:6443`
> > In another terminal, run:
> > `RPC_HOST_URL="http://localhost:6443" cargo run --bin glaredb -- local`
> 
> It seems like there is a bit of a mix of env vars & cli args. We can clean this up across the board. IMO, everything that can be set via a cli arg should also be settable via env vars & vice-versa. We could also expose a config file with all of the same arguments.

For sure, forgot to mention that `RPC_HOST_URL` is just there to test this out. I plan on that being removed when we have the new `\open` behavior.

crates/glaredb/src/bin/main.rs
```diff
@@ -48,10 +48,14 @@ enum Commands {
 
     /// Starts the sql server portion of GlareDB.
     Server {
-        /// TCP address to bind to.
+        /// TCP address to bind to for the Postgres interface.
         #[clap(short, long, value_parser, default_value_t = String::from("0.0.0.0:6543"))]
         bind: String,
```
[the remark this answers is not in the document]
scsmithr: Possibly, but I'm not completely sure how we want this to be run outside of Cloud.

crates/rpcsrv/src/session.rs
```diff
@@ -0,0 +1,58 @@
...
+                let plan = LogicalPlanNode::try_decode(&buf)?
+                    .try_into_logical_plan(&fake_ctx, &DefaultLogicalExtensionCodec {})?;
+
+                let session = self.session.lock().await;
+                let physical = session.create_physical_plan(plan).await?;
+                let stream = session.execute_physical(physical)?;
+
+                Ok(stream)
+            }
+            Some(Plan::PhysicalPlan(_)) => Err(RpcsrvError::PhysicalPlansNotSupported),
```
vrongmeal, line 54: In what scenario do we expect to execute physical plans this way?

vrongmeal (approved):
This looks amazing 💯 

Just have a tiny question. But all is good from my side :D

crates/rpcsrv/src/session.rs
scsmithr: I don't know if/when we want to support physical plan proxying, but figured I throw it in anyways. Logical plans are a lot easier to start with, but are limited. And I figured if we want to extend this out to allow execution across multiple compute nodes, passing around physical plans would allow the planner to put in things like broadcast joins and stuff like that.

PR #1508, opened 2023-08-07
feat: RPC proxy

### LINKED ISSUE #1501: Proxy service for rpcsrv
## Description

Very similar to `pgsrv` in that it sits in front of everything, and authenticates connections to Cloud. This should use the same auth endpoint.

The proxy will implement the service defined in the protobuf with no modifications. Authentications details will be provided in request metadata (at least for now).

Going through two layers of grpc has some inefficiencies, but I believe the performance around queries with remote services is around actual execution, and not transport.

scsmithr:
Needs #1499 

Closes #1501 

Adds in the rpcsrv proxy.

```
Starts an instance of the rpcsrv proxy

Usage: glaredb rpc-proxy [OPTIONS] --cloud-api-addr <CLOUD_API_ADDR> --cloud-auth-code <CLOUD_AUTH_CODE>

Options:
  -b, --bind <BIND>                        TCP address to bind to [default: 0.0.0.0:6444]
      --cloud-api-addr <CLOUD_API_ADDR>    Address of the GlareDB cloud server
      --cloud-auth-code <CLOUD_AUTH_CODE>  Authorization code for communicating with Cloud
  -h, --help                               Print help
```

Internally this uses the same auth mechanism as in pgsrv, so credentials that work for the postgres stuff will also work for this.


universalmind303:
proper diff here: https://github.com/GlareDB/glaredb/compare/sean/rpc...sean/rpc-proxy

vrongmeal (approved):
Looks good 🚀

scsmithr:
@f0ssel gonna merge this to try to get it in before the terraform stuff. Hopefully image build finishes by then...

PR #1516, opened 2023-08-08
fix: Fix pg handler not running

scsmithr:
Needed to run rpcsrv in the background

### DIFF
--- a/crates/glaredb/src/server.rs
+++ b/crates/glaredb/src/server.rs
@@ -168,7 +168,7 @@ impl ComputeServer {
                 {
                     // TODO: Maybe panic instead? Revisit once we have
                     // everything working.
-                    error!(%e, "rpc serviced died");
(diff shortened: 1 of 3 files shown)


crates/glaredb/src/server.rs
```diff
@@ -159,11 +159,18 @@ impl ComputeServer {
...
+            tokio::spawn(async move {
+                if let Err(e) = Server::builder()
+                    .trace_fn(|_| debug_span!("rpc_service_request"))
+                    .add_service(ExecutionServiceServer::new(handler))
+                    .serve(addr)
+                    .await
+                {
+                    // TODO: Maybe panic instead? Revisit once we have
+                    // everything working.
+                    error!(%e, "rpc serviced died");
```
greyscaled, line 171:
```suggestion
                    error!(%e, "rpc service died");
```

PR #1518, opened 2023-08-08
feat: Authenticated rpc client

scsmithr:
Adds connection authentication to the rpc client.

Significant changes:
- Authentication parameters are set in the request metadata. `AuthenticatedExecutionServiceClient` was added to do this automatically.
- Properly passes along the session storage config to the remote session.
- Remotes `db_id` from the (client) initialize session request since the client won't know what the ID is.

Things to do in follow up PRs:
- Create remote clients through python lib.


crates/glaredb/src/local.rs
```diff
@@ -139,11 +144,15 @@ impl LocalSession {
         )
         .await?;
 
-        // TODO: Make this configurable through client commands.
-        let sess = if let Ok(url) = std::env::var("RPC_HOST_URL") {
```
greyscaled, line 143: This effectively drops "support" for that env var, right?

scsmithr: Yeah, that env var was always a temporary thing to test it out. I planned on removing it.

crates/sqlexec/src/remote/client.rs
```diff
@@ -0,0 +1,257 @@
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ProxyAuthParams {
+    /// User (generated Cloud credentials)
+    pub user: String,
+    /// Password (generated Cloud credentials)
+    pub password: String,
+    /// DB name
+    pub db_name: String,
+    /// Org name.
+    pub org: String,
```
greyscaled, line 29: Is this allowed to be name or id, or always just name?

scsmithr: `org` can be name or id. We have logic to check that and send the appropriate params: https://github.com/GlareDB/glaredb/blob/main/crates/proxyutil/src/cloudauth.rs#L106-L124

`db_name` can only be name.

I'd like for us to phase out being able to provide org ID (https://github.com/GlareDB/glaredb/blob/main/crates/proxyutil/src/cloudauth.rs#L58-L60). As a user, I'm only ever going to use my org name. I think ID is sticking around for the frontend stuff.

crates/glaredb/src/local.rs
```diff
@@ -295,14 +304,25 @@ impl LocalSession {
...
-                    data_dir: Some(PathBuf::from(path)),
-                    metastore_addr: None,
-                    ..self.opts.clone()
-                };
-                let new_sess = LocalSession::connect(new_opts).await?;
-                println!("Created new session. New database path: {path}");
-                *self = new_sess;
+                if let Ok(url) = Url::parse(path) {
+                    let new_opts = LocalClientOpts {
+                        data_dir: None,
```
greyscaled, line 309: I'm assuming this is because we're connecting remotely `cloud_url` and won't have local data to persist? Might make sense to mention it in CLI options

```sh
$ cargo run --bin glaredb -- --help
  -f, --data-dir <DATA_DIR>
          Optional file path for persisting data.

          Catalog data and user data will be stored in this directory.

  -c, --cloud-url <CLOUD_URL>
          URL for connecting to a GlareDB Cloud deployment
```

greyscaled (approved):
Seems good in how the proxy + metadata works via wrappers that attach in `AuthenticatedExecutionServiceClient`.

I think updating `--help` text for `--data-dir` is a nice-to-have, but ✔️

crates/sqlexec/src/remote/client.rs
greyscaled: Currently it is used for FE, yes. Should make those changes in Cloud (+verify), which would allow us to pull that impl out from here as "dead code" afterwards.

crates/glaredb/src/local.rs
greyscaled: Yea no prob, I never got around to using it just thought i'd double-check

crates/sqlexec/src/remote/client.rs
greyscaled: https://github.com/GlareDB/cloud/issues/1960

crates/glaredb/src/local.rs
scsmithr: updated

PR #1528, opened 2023-08-09
proto serialize logical plans

universalmind303:
this uses the datafusion `Extension` as a means to transmit the custom nodes. 

In the process, we convert our logicalPlan to datafusion extension nodes.

for example
`LogicalPlan(DDL(CreateTable))` -> `DfLogicalPlan(Extension(CreateTable))`.


crates/sqlexec/src/session.rs
```diff
@@ -231,10 +232,20 @@ impl Session {
...
+        if let DfLogicalPlan::Extension(extension) = &plan {
+            match extension.node.name() {
+                "CreateTable" => {
+                    let create_table = extension.node.as_any().downcast_ref::<CreateTable>();
+                    return self.create_table(create_table.unwrap().clone()).await;
+                }
+                _ => {}
+            }
+        }
+
```
universalmind303, line 248: we want to handle the extensions using our session context, not datafusions, so we need to handle those cases here.

crates/sqlexec/Cargo.toml
```diff
@@ -32,6 +32,10 @@ tokio-postgres = "0.7.8"
 once_cell = "1.18.0"
 url.workspace = true
 parking_lot = "0.12.1"
+serde = { workspace = true }
+
+[dependencies.sqlparser]
+features = ["serde"]
```
scsmithr: What's this for?

scsmithr (commented):
lgtm, definitely feels like something we can work off of.

scsmithr (approved):

crates/sqlexec/Cargo.toml
universalmind303: was looking into something & forgot to remove this.

vrongmeal (approved):

Issue #1502, opened 2023-08-07
Serializable external table plans


scsmithr:
## Description

Currently rpcsrv cannot execute queries that make use of any of our custom plans (including table plans).

This issue is to track external table plans, and should be expanded upon. We'll likely be able to use our existing protobuf definitions for table entries.

PR #1534, opened 2023-08-11
feat: Support remote execution for queries with tables

vrongmeal:
Basically everything happens on remote (except creating the logical plan). Table providers and Physical plans are created on remote and executed on remote server. All the calls are RPC. Everything is stored in session context against a unique ID which the client uses to call on the server.

**Internal changes:**

* Renamed the server's `--ignore-auth` to `--ignore-pg-auth`.
* Use `--disable-rpc-auth` to start server for calling rpc directly without proxy.
* Use `--ignore-rpc-auth` with local to call the rpc server directly when `--disable-rpc-auth` is set.

Fixes: #1502

### DIFF
--- a/crates/glaredb/src/bin/main.rs
+++ b/crates/glaredb/src/bin/main.rs
@@ -110,7 +110,17 @@ enum Commands {
         /// This is only relevant for internal development. The postgres
         /// protocol proxy will drop all authentication related messages.
         #[clap(long, hide = true, value_parser)]
-        ignore_auth: bool,
+        ignore_pg_auth: bool,
+
+        /// Allow rpc messages directly from the client without proxy.
+        ///
+        /// (Internal)
+        ///
+        /// This is only relevant for internal development. The RPC proxy will
+        /// allow `Client` messages directly. Need to use `--ignore-rpc-auth`
+        /// with the `local` command to access this.
+        #[clap(long, hide = true, value_parser)]
+        disable_rpc_auth: bool,
 
         /// API key for segment.
         ///
@@ -222,7 +232,8 @@ fn main() -> Result<()> {
             service_account_path,
             mut segment_key,
             spill_path,
-            ignore_auth,
+            ignore_pg_auth,
+            disable_rpc_auth,
             rpc_bind,
         } => {
             // Map an empty string to None. Makes writing the terraform easier.
@@ -231,7 +242,7 @@ fn main() -> Result<()> {
             let auth: Box<dyn LocalAuthenticator> = match password {
                 Some(password) => Box::new(SingleUserAuthenticator { user, password }),
                 None => Box::new(PasswordlessAuthenticator {
-                    drop_auth_messages: ignore_auth,
+                    drop_auth_messages: ignore_pg_auth,
                 }),
             };
 
@@ -249,7 +260,7 @@ fn main() -> Result<()> {
                 data_dir,
                 service_account_key,
                 spill_path,
-                ignore_auth,
+                disable_rpc_auth,
             )?;
         }
         Commands::PgProxy {
@@ -336,7 +347,7 @@ fn begin_server(
     data_dir: Option<PathBuf>,
     service_account_key: Option<String>,
     spill_path: Option<PathBuf>,
-    ignore_auth: bool,
+    disable_rpc_auth: bool,
 ) -> Result<()> {
     let runtime = build_runtime("server")?;
     runtime.block_on(async move {
@@ -353,7 +364,7 @@ fn begin_server(
             service_account_key,
             spill_path,
             /* integration_testing = */ false,
-            ignore_auth,
+            disable_rpc_auth,
         )
         .await?;
         server.serve(conf).await
--- a/crates/protogen/src/lib.rs
+++ b/crates/protogen/src/lib.rs
@@ -5,8 +5,54 @@
 //! project. There should be a minimal amount of logic in this crate.
 
 pub mod metastore;
+pub mod rpcsrv;
 pub mod export {
     pub use prost;
+}
+
+pub mod errors {
+    /// Errors related to converting to/from protobuf types.
+    #[derive(thiserror::Error, Debug)]
+    pub enum ProtoConvError {
+        #[error("Field required: {0}")]
+        RequiredField(String),
+
+        #[error("Unknown enum variant for '{0}': {1}")]
+        UnknownEnumVariant(&'static str, i32),
+
+        #[error("Received zero-value enum variant for '{0}'")]
+        ZeroValueEnumVariant(&'static str),
+
+        #[error("Unsupported serialization: {0}")]
+        UnsupportedSerialization(&'static str),
+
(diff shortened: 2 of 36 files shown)


crates/glaredb/src/bin/main.rs
```diff
@@ -110,7 +110,17 @@ enum Commands {
...
         /// protocol proxy will drop all authentication related messages.
         #[clap(long, hide = true, value_parser)]
-        ignore_auth: bool,
+        ignore_pg_auth: bool,
+
+        /// Allow rpc messages directly from the client without proxy.
+        ///
+        /// (Internal)
+        ///
+        /// This is onlu relevant for internal development. The RPC proxy will
```
universalmind303, line 119:
```suggestion
        /// This is only relevant for internal development. The RPC proxy will
```

universalmind303:
@vrongmeal would it be too much work to add a mermaid diagram to the PR? This is a lot to take in & a visual would definitely be helpful!

crates/rpcsrv/src/session.rs
```diff
@@ -35,25 +36,49 @@ impl RemoteSession {
...
-            None => Err(RpcsrvError::Internal("missing plan on request".to_string())),
-        }
+        logical_plan: impl AsRef<[u8]>,
+    ) -> Result<PhysicalPlanResponse> {
+        let mut session = self.session.lock().await;
+
+        // TODO: Use a context that actually matters.
+        let fake_ctx = SessionContext::new();
+        let plan = LogicalPlanNode::try_decode(logical_plan.as_ref())?
+            .try_into_logical_plan(&fake_ctx, &session.extension_codec())?;
```
scsmithr, line 48: I would have thought we'd need to change this context with one that had the tables registered. Is handling for the custom table providers done entirely with extension codec?

crates/rpcsrv/src/session.rs
```diff
@@ -35,25 +36,49 @@ impl RemoteSession {
...
+        let schema = provider.schema();
+        let provider_id = session.add_table_provider(provider);
+
+        Ok(TableProviderResponse {
+            id: provider_id,
+            schema: Schema::clone(&schema),
+        })
+    }
+
+    pub async fn physical_plan_execute(&self, exec_id: Uuid) -> Result<SendableRecordBatchStream> {
```
universalmind303, line 75: I'm worried that this stateful pattern of `add/get` physical plan may cause some contention as we scale up, or run multiple clients. 

What issues were there with the remoteExec node? It seems to fit better with datafusion. couldn't we just setup the extensionCodec with a registry of table providers?

crates/sqlexec/src/context.rs
```diff
@@ -87,6 +87,10 @@ pub struct SessionContext {
     env_reader: Option<Box<dyn EnvironmentReader>>,
     /// Job runner for background jobs.
     background_jobs: JobRunner,
+    /// Session's table providers.
+    table_providers: HashMap<uuid::Uuid, Arc<dyn TableProvider>>,
+    /// Session's physical plans.
+    physical_plans: HashMap<uuid::Uuid, Arc<dyn ExecutionPlan>>,
```
scsmithr, line 93: Context getting chunky.

Are these specific to the "remote" side?

crates/glaredb/src/bin/main.rs
```diff
@@ -110,7 +110,17 @@ enum Commands {
...
+        /// allow `Client` messages directly. Need to use `--ignore-rpc-auth`
+        /// with the `local` command to access this.
+        #[clap(long, hide = true, value_parser)]
+        allow_client_rpc_init: bool,
```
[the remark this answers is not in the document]
scsmithr: I think `--ignore-rpc-auth` is a typo. `allow_client_rpc_init` is what enables ignoring auth.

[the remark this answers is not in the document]
vrongmeal: For the server I called it `--allow-client-rpc-init`. For local it's `--ignore-rpc-auth`. I can rename the flag for both to `--ignore-rpc-auth`.

crates/rpcsrv/src/session.rs
vrongmeal: The issue is serializing table providers and physical plans. It's not as simple. We'll have to change a lot to make them serializable. For example, how do you serialize a provider with postgres connection? We either need the connection on the server (stateful again), or we create a new connection everytime.

crates/rpcsrv/src/handler.rs
```diff
@@ -94,74 +87,131 @@ impl RpcHandler {
...
+        _request: Request<service::TableProviderScanRequest>,
+    ) -> Result<Response<service::PhysicalPlanResponse>, Status> {
+        todo!()
+    }
+
+    async fn table_provider_insert_into(
+        &self,
+        _request: Request<service::TableProviderInsertIntoRequest>,
+    ) -> Result<Response<service::PhysicalPlanResponse>, Status> {
+        todo!()
```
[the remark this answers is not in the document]
vrongmeal: Yeah, I wanted to get this in and will add it in another PR.

crates/protogen/src/lib.rs
```diff
@@ -5,10 +5,56 @@
...
+        UnknownEnumVariant(&'static str, i32),
+
+        #[error("Received zero-value enum variant for '{0}'")]
+        ZeroValueEnumVariant(&'static str),
+
+        #[error("Unsupported serialization: {0}")]
+        UnsupportedSerialization(&'static str),
+
+        #[error("Invalid table reference: {0:?}.{1:?}.{2:?}")]
+        InvalidTableReference(String, String, String),
```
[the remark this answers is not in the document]
vrongmeal: I did this because debug formatting adds quotation marks, which will be helpful for identifying what's wrong with the reference.

```rust
// Invalid table ref:
(Some("database"), None, "table")

// will print as:
// "database".""."table"
```

IMO more readable than `database..table`. Moreover, complies with how we accept identifiers.

crates/sqlexec/src/context.rs
vrongmeal: Yeah, specific to remote side.

scsmithr: Let's move these into a separate struct to avoid bloating up this Context struct with remote specific state.

scsmithr (approved):
Approve (with small changes)

crates/rpcsrv/src/session.rs
vrongmeal: Yeah, we really should change it ASAP. Will create an issue for this.

Handling for all custom providers has to be done through extension codec... I don't see any other way right now 🤔

crates/rpcsrv/src/session.rs
vrongmeal: Yeah, we do have physical plan extension similar to logical plans. I did think about this but at this moment, getting everything there requires a lot of effort. But we should definitely go there in the future.

universalmind303 (approved):

PR #1538, opened 2023-08-11
fix: number of deleted rows returned when there is no where clause

### LINKED ISSUE #1537: Delete not returning correct number of rows deleted
## Description

```
> create table a (a int);
┌┐
└┘
> insert into a values (1);
Write success
> delete from a;
Deleted 0 rows
> select * from a;
┌┐
└┘
```

adhish20:
Fixes #1537

### DIFF
--- a/crates/datasources/src/native/access.rs
+++ b/crates/datasources/src/native/access.rs
@@ -172,21 +172,32 @@ impl NativeTableStorage {
 
     pub async fn delete_rows_where(
         &self,
-        table: &TableEntry,
+        table_entry: &TableEntry,
         where_expr: Option<Expr>,
     ) -> Result<usize> {
-        let table = self.load_table(table).await?;
-        let builder = if let Some(where_expr) = where_expr {
-            DeleteBuilder::new(table.delta.object_store(), table.delta.state)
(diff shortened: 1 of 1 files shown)


crates/datasources/src/native/access.rs
```diff
@@ -172,21 +172,32 @@ impl NativeTableStorage {
...
+            let deleted_rows = DeleteBuilder::new(table.delta.object_store(), table.delta.state)
                 .with_predicate(where_expr)
+                .await?
+                .1
+                .num_deleted_rows;
+            if let Some(rows) = deleted_rows {
+                Ok(rows)
+            } else {
+                Ok(0_usize)
+            }
```
scsmithr, line 189:
```suggestion
            Ok(deleted_rows.unwrap_or_default())
```

scsmithr (approved):

greyscaled:
@adhish20 && @scsmithr is this something we should add a test for ?

adhish20:
> @adhish20 && @scsmithr is this something we should add a test for ?

It's difficult to add a test for this currently as we do not have a way to assert the return from a sql command

PR #1539, opened 2023-08-11
remote exec for DDL's

universalmind303:
so far I have the following DDL nodes implemented
- createTable
- createSchema
- dropTables

I think the template I have set up should be pretty easy to follow for adding all of our custom nodes as extension nodes. 
Also think we can probably merge this in as is & follow up with the additional DDL nodes in a separate PR.


crates/protogen/src/sqlexec/logical_plan.rs
```diff
@@ -0,0 +1,41 @@
...
+    pub if_not_exists: bool,
+    #[prost(message, optional, tag = "3")]
+    pub schema: Option<DfSchema>,
+    #[prost(message, optional, tag = "4")]
+    pub source: Option<LogicalPlanNode>,
+}
+
+#[allow(clippy::derive_partial_eq_without_eq)]
+#[derive(Clone, PartialEq, Message)]
+pub struct LogicalPlanExtension {
```
universalmind303, line 18: since protobuf messages are not self-describing, we have to roll all extensions up into a single message containing a `oneof`. Otherwise we lose context of what the original type was & have to try all variants until we find a match.

crates/sqlexec/src/session.rs
```diff
@@ -245,16 +245,19 @@ impl Session {
...
-            match extension.node.name() {
-                "CreateTable" => {
-                    let create_table = extension.node.as_any().downcast_ref::<CreateTable>();
-                    return self.create_table(create_table.unwrap().clone()).await;
-                }
-                _ => {}
-            }
-        }
+        
+        // TODO! if its connected to a remote instance, we want them to execute the plan.
```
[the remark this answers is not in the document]
scsmithr: I added a `remote_session_id` var that defaults to None: https://github.com/glaredb/glaredb/blob/142b723808570fb32c06c2d39a8dd731a65eadd9/crates/datafusion_ext/src/vars.rs#L149-L155

Can be assumed that if None, everything is happening local. If Some, then it's connected to a remote session.

Could also check if `exec_client` is Some on the session context, but I think that's subject to change.

universalmind303:
@vrongmeal @scsmithr Not all of the ddl is implemented yet, but I'm pretty happy with the layout of everything here. I think the rest of them should be pretty simple boilerplate based off of the ones i implemented [createTable, createSchema, dropTables].

crates/sqlexec/src/planner/logical_plan/alter_database_rename.rs
```diff
@@ -0,0 +1,5 @@
+#[derive(Clone, Debug)]
```
universalmind303, line 1: i know these are rather empty right now, but they will fill up with all of the boilerplate for (de)serialization & wanted to keep things a bit more organized.

crates/sqlexec/src/planner/logical_plan/create_schema.rs
```diff
@@ -0,0 +1,82 @@
...
+        &self,
+        _exprs: &[datafusion::prelude::Expr],
+        _inputs: &[DfLogicalPlan],
+    ) -> Self {
+        self.clone()
+    }
+}
+
+impl ExtensionType for CreateSchema {
+    const EXTENSION_NAME: &'static str = "CreateSchema";
```
universalmind303, line 53: TIL: you can specify a const in a trait.

crates/sqlexec/src/planner/logical_plan.rs
```diff
@@ -20,7 +66,7 @@ pub enum LogicalPlan {
     Write(WritePlan),
     /// Plans related to querying the underlying data store. This will run
     /// through datafusion.
-    Query(DfLogicalPlan),
+    Datafusion(DfLogicalPlan),
```
universalmind303, line 69: renamed this because as more of the DDL gets moved in to Extension nodes id expect us to potentially get rid of the DDL variant all together. so calling it `Query` didn't really make sense anymore, as it could potentially be any kind of operation now.

crates/protogen/src/sqlexec/logical_plan.rs
scsmithr, line 1: Is this file generated? If not, why not?

crates/sqlexec/src/planner/logical_plan/create_schema.rs
```diff
@@ -0,0 +1,82 @@
...
+pub struct CreateSchema {
+    pub schema_name: OwnedSchemaReference,
+    pub if_not_exists: bool,
+}
+
+impl TryFrom<protogen::sqlexec::logical_plan::CreateSchema> for CreateSchema {
+    type Error = ProtoConvError;
+
+    fn try_from(proto: protogen::sqlexec::logical_plan::CreateSchema) -> Result<Self, Self::Error> {
+        let schema_name = proto.schema_name.unwrap().try_into().unwrap();
```
scsmithr, line 13: Should we be `?`ing these instead of unwrapping?

crates/sqlexec/src/remote/planner.rs
```diff
@@ -1,38 +1,82 @@
...
     }
+    fn create_physical_expr(
+        &self,
+        _expr: &Expr,
+        _input_dfschema: &DFSchema,
+        _input_schema: &Schema,
+        _session_state: &SessionState,
+    ) -> DataFusionResult<Arc<dyn PhysicalExpr>> {
+        todo!("create_physical_expr")
+    }
```
scsmithr, line 81: We'll probably want to use the default implementation. I don't imagine we'll be working with custom expressions where this is required.

scsmithr:
This is looking good. Had a question about the logical plan file since that looks generated. Was wondering why it's being checked in.

crates/protogen/src/sqlexec/logical_plan.rs
universalmind303: no, this is actually manually created. It was the only way I could find that allowed me to both skip the intermediate structs & reference datafusion imported protobuf definitions.

crates/sqlexec/src/planner/logical_plan/create_schema.rs
universalmind303: yes, looks like i missed a couple after testing things.

crates/protogen/src/sqlexec/logical_plan.rs
vrongmeal: Let's just use `bytes` in proto and encode and decode those into the correct struct ourselves? I feel that's much cleaner and we also have the proto file. This way we lose the proto definitions.

See:

https://github.com/GlareDB/glaredb/blob/758b0d799657483e6b0c4eea166dc0a41b64777c/crates/protogen/src/rpcsrv/types/service.rs#L220

This is an example of intermediate struct but you can implement `TryFrom` on the actual struct for proto definitions. Our intermediate structs are essentially what we use (at least for options). I was thinking of moving those sometime before.

PR #1543, opened 2023-08-14
fix: bug with printing tables that has multiple record batches

### LINKED ISSUE #1536: Incorrect values printed when results contain multiple record batches
## Description

```
> create table a (a int);
┌┐
└┘
> insert into a values (1);
Write success
> insert into a values (2);
Write success
> select * from a;
┌───────┐
│ a     │
│ ──    │
│ Int32 │
╞═══════╡
│ 1     │
│ 1     │
│ 2     │
└───────┘
```

adhish20:
Fixes #1536

### DIFF
--- a/crates/arrow_util/src/pretty.rs
+++ b/crates/arrow_util/src/pretty.rs
@@ -45,7 +45,7 @@ fn create_table(
         return Ok(default_table());
     }
     let num_columns = batches[0].num_columns();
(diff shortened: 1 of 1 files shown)


scsmithr (approved):
Tested it and it works for me.

I think we'll want to start looking at adding tests to this... surprisingly a good bit of logic here.

universalmind303 (approved):

PR #1552, opened 2023-08-15
feat(rpcsrv): rpc alter stmts

universalmind303:
this implements remote exec for the following; 

- [x] alterTable
- [x] alterDatabase
- [x] alterTunnelRotateKeys

### DIFF
--- a/crates/sqlexec/src/planner/extension.rs
+++ b/crates/sqlexec/src/planner/extension.rs
@@ -2,7 +2,7 @@
 use datafusion_proto::logical_plan::LogicalExtensionCodec;
 use std::sync::Arc;
 
-use crate::errors::Result;
+use crate::{errors::Result, LogicalPlan};
 use datafusion::logical_expr::{Extension as LogicalPlanExtension, UserDefinedLogicalNodeCore};
 
 pub trait ExtensionType: Sized + UserDefinedLogicalNodeCore {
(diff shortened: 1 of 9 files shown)


crates/sqlexec/src/planner/extension.rs
```diff
@@ -12,8 +12,13 @@ pub trait ExtensionType: Sized + UserDefinedLogicalNodeCore {
...
         }
     }
+    fn into_logical_plan(self) -> LogicalPlan {
+        LogicalPlan::Datafusion(datafusion::logical_expr::LogicalPlan::Extension(
+            self.into_extension(),
+        ))
+    }
     fn try_decode_extension(extension: &LogicalPlanExtension) -> Result<Self>;
-    fn try_encode(&self, buf: &mut Vec<u8>, codec: &dyn LogicalExtensionCodec) -> Result<()>;
+    fn try_encode(&self, buf: &mut Vec<u8>, _codec: &dyn LogicalExtensionCodec) -> Result<()>;
```
scsmithr, line 21: Why underscore (since this a trait method)?

scsmithr (approved):
Ship!

crates/sqlexec/src/planner/extension.rs
universalmind303: using `_codec` because many of the implementers don't need the codec. the codec is only needed when you have non leaf nodes. 

So it's there if you need it, but many implementers don't need it. so when you implement the trait, you dont have to manually add the `_` on every implementation that doesn't need it. seemed like a sensible default to me.

PR #1553, opened 2023-08-15
feat: Support remote table scans and inserts

vrongmeal:
```
> select * from abc;
┌───────┐
│ a     │
│ ──    │
│ Int32 │
╞═══════╡
│ 1     │
│ 2     │
│ 3     │
└───────┘
> insert into abc values (4), (5);
Write success
> select * from abc;
┌───────┐
│ a     │
│ ──    │
│ Int32 │
╞═══════╡
│ 1     │
│ 2     │
│ 3     │
│ 4     │
│ 5     │
└───────┘
```

**Logs:**

```
initializing remote session session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668
dispatching table access session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 table_ref=abc
creating physical plan session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668
executing physical plan session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 exec_id=e72dc636-faf9-4953-b3bb-0bd786a626e5
dispatching table access session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 table_ref=abc
creating physical plan session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668
insert into table provider session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 provider_id=21f8f6c2-51af-44bf-b3e3-57367363ca0c
executing physical plan session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 exec_id=9c2a7f9f-7449-408e-b667-357e10975801
dispatching table access session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 table_ref=abc
creating physical plan session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668
executing physical plan session_id=accc8ed7-d0e1-463a-834d-6f5ae2199668 exec_id=ea5525e1-38a4-45db-b90a-da9b499e5216
```

### DIFF
--- a/crates/rpcsrv/src/session.rs
+++ b/crates/rpcsrv/src/session.rs
@@ -2,11 +2,10 @@ use crate::errors::Result;
 use datafusion::arrow::datatypes::Schema;
 use datafusion::common::OwnedTableReference;
 use datafusion::physical_plan::SendableRecordBatchStream;
-use datafusion::prelude::SessionContext;
+use datafusion::prelude::{Expr, SessionContext};
 use datafusion_proto::logical_plan::AsLogicalPlan;
 use datafusion_proto::protobuf::LogicalPlanNode;
 use protogen::metastore::types::catalog::CatalogState;
-use protogen::rpcsrv::types::service::{PhysicalPlanResponse, TableProviderResponse};
 use sqlexec::engine::TrackedSession;
 use std::sync::Arc;
 use tokio::sync::Mutex;
@@ -39,7 +38,7 @@ impl RemoteSession {
     pub async fn create_physical_plan(
         &self,
         logical_plan: impl AsRef<[u8]>,
-    ) -> Result<PhysicalPlanResponse> {
+    ) -> Result<(Uuid, Schema)> {
         let mut session = self.session.lock().await;
 
         // TODO: Use a context that actually matters.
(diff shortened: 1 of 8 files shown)


universalmind303 (approved):
I think this is fine for a stop-gap solution. But like i've mentioned, I think moving to serializable physical plans will be a much better long term solution.

crates/rpcsrv/src/session.rs
```diff
@@ -51,26 +50,50 @@ impl RemoteSession {
...
+        let provider = session.get_table_provider(&provider_id)?;
+        // TODO: Use a context that actually matters.
+        let fake_ctx = SessionContext::new();
+        let physical = provider
+            .scan(&fake_ctx.state(), projection, filters, limit)
+            .await?;
+        let schema = physical.schema();
+        let exec_id = session.add_physical_plan(physical)?;
+        Ok((exec_id, Schema::clone(&schema)))
+    }
```
scsmithr, line 81: Is scanning here actually needed? Since we're passing in logical plans, those will get converted into physical plans on the remote side, which is what's actually doing the scanning.

As an aside, I think we'll want to convert everything into an execution plan so that we're only dealing with logical and physical plans. Currently we have things like `DropTable` logical plans that we execute directly. If we instead do `DropTable` -> `DropTableExec`, we don't have to special case on the logical plans.

scsmithr (approved):
Comment shouldn't block, just something we need to think about. I'm focused on getting stuff working, adding tests, then going back to tweak things as needed.

crates/rpcsrv/src/session.rs
vrongmeal: Yeah, it's not required. Not sure it won't be when we have hybrid execution, though. I am not sure how we're going to modify the query planner. Keeping it here, we can remove it later if not required.

vrongmeal:
> I think this is fine for a stop-gap solution. But like i've mentioned, I think moving to serializable physical plans will be a much better long term solution.

Yup! Probably going to work on this next.

PR #1556, opened 2023-08-15
feat: rpc for create stmts

universalmind303:
adds rpc for the rest of the create statements. 

also a small refactor to the extensions so that we have exhaustive pattern matching. (was finding myself missing a few spots when adding a new extension node). Should be a bit easier as it'll compile error now.

### DIFF
--- a/crates/glaredb/src/highlighter.rs
+++ b/crates/glaredb/src/highlighter.rs
@@ -107,9 +107,14 @@ fn colorize_sql(query: &str, st: &mut StyledText) -> std::io::Result<()> {
                 | Keyword::WITH
                 | Keyword::INSERT
                 | Keyword::INTO
+                | Keyword::DATABASE
                 | Keyword::VALUES => {
                     st.push((Style::new().fg(Color::LightGreen), format!("{w}")));
                 }
+                Keyword::NoKeyword => match w.value.to_uppercase().as_str() {
+                    "TUNNEL" => st.push((Style::new().fg(Color::LightGreen), format!("{w}"))),
+                    _ => st.push((Style::new(), format!("{w}"))),
+                },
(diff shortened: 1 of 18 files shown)


crates/glaredb/src/highlighter.rs
```diff
@@ -107,9 +107,14 @@ fn colorize_sql(query: &str, st: &mut StyledText) -> std::io::Result<()> {
...
                 | Keyword::INTO
+                | Keyword::DATABASE
                 | Keyword::VALUES => {
                     st.push((Style::new().fg(Color::LightGreen), format!("{w}")));
                 }
+                Keyword::NoKeyword => match w.value.to_uppercase().as_str() {
+                    "TUNNEL" => st.push((Style::new().fg(Color::LightGreen), format!("{w}"))),
+                    _ => st.push((Style::new(), format!("{w}"))),
+                },
                 // TODO: add more keywords
```
universalmind303, line 118: cli highlighting was missing these keywords

crates/sqlexec/src/extension_codec.rs
```diff
@@ -18,6 +18,8 @@ pub struct GlareDBExtensionCodec<'a> {
     table_providers: Option<&'a HashMap<Uuid, Arc<dyn TableProvider>>>,
 }
 
+// This tracks all of our extensions so that we can ensure an exhaustive match on anywhere that uses the extension
```
scsmithr, line 21: The codec? Or is there something missing here?

universalmind303: whoops, i originally had `ExtensionType` in this file, but moved it to `extension`. i guess i forgot to move the comment.

scsmithr (approved):

PR #1558, opened 2023-08-15
wip: physical plan nodes

universalmind303:
just pushing this up to start some conversation on this. 

the general idea is to add an <Action>Exec node for each action we can perform. 

most of the exec is currently taking place in `crates/sqlexec/src/context.rs`. I'd expect we'd move most of that exec logic into the individual nodes. 

for example instead of 

```rust
impl SessionContext {
    pub async fn create_temp_table(&mut self, plan: CreateTempTable) -> Result<()> {
       // implement the creation of tables here
    }
```

this will all happen inside `CreateTempTableExec::execute`


crates/sqlexec/src/planner/physical_plan/create_tunnel_exec.rs
```diff
@@ -0,0 +1,62 @@
...
+    ) -> datafusion::error::Result<std::sync::Arc<dyn ExecutionPlan>> {
+        Ok(self)
+    }
+
+    fn execute(
+        &self,
+        partition: usize,
+        context: std::sync::Arc<datafusion::execution::TaskContext>,
+    ) -> datafusion::error::Result<datafusion::physical_plan::SendableRecordBatchStream> {
+        todo!()
```
universalmind303, line 56: It seems like in order to use custom physical nodes, we'd have to move a lot of the sessionContext in to native datafusion extensions

```rust
pub struct SessionContext {
    /// The execution client for remote sessions.
    // TODO: This is currently unused, but we'll likely need it for running some
    // of our custom plans on a remote service.
    exec_client: Option<RemoteSessionClient>,
    /// Database catalog.
    catalog: SessionCatalog,
    /// In-memory (temporary) tables.
    current_session_tables: HashMap<String, Arc<MemTable>>,
    /// Native tables.
    tables: NativeTableStorage,
    /// Session variables.
    vars: SessionVars,
    /// Prepared statements.
    prepared: HashMap<String, PreparedStatement>,
    /// Bound portals.
    portals: HashMap<String, Portal>,
    /// Track query metrics for this session.
    metrics: SessionMetrics,
    /// Datafusion session context used for planning and execution.
    df_ctx: DfSessionContext,
    /// Read tables from the environment.
    env_reader: Option<Box<dyn EnvironmentReader>>,
    /// Job runner for background jobs.
    background_jobs: JobRunner,
    /// Specific stuff for remote session.
    remote_ctx: Option<RemoteSessionContext>,
}
```
we'd need to execute the same logic that happens here,
```rust
    pub async fn create_tunnel(&mut self, plan: CreateTunnel) -> Result<()> {
        if let &Some(limit) = self.vars.max_tunnel_count.value() {
            if self.get_tunnel_count() >= limit {
                return Err(ExecError::MaxObjectCount {
                    typ: "tunnels",
                    max: limit,
                    current: self.get_tunnel_count(),
                });
            }
        }

        self.catalog
            .mutate([Mutation::CreateTunnel(service::CreateTunnel {
                name: plan.name,
                if_not_exists: plan.if_not_exists,
                options: plan.options,
            })])
            .await?;
        Ok(())
    }
```
and move it into the `execute` function. 

but that only has access to the datafusion `TaskContext`. So we'd likely have to move the relevent properties into datafusion session extensions.

```rust
fn execute(...) {
  let catalog = context.session_config().get_extension::<SessionCatalog>().unwrap();
  let vars = context.session_config().get_extension::<SessionVars>().unwrap();
}
// ... do rest of logic

File: crates/sqlexec/src/planner/physical_plan/create_tunnel_exec.rs [hunk 2]
```diff
@@ -0,0 +1,62 @@
...
+use protogen::metastore::types::options::TunnelOptions;
+
+use crate::planner::EMPTY_SCHEMA;
+
+#[derive(Debug)]
+pub struct CreateTunnelExec {
+    pub name: String,
+    pub if_not_exists: bool,
+    pub options: TunnelOptions,
+}
```
universalmind303, line 11: it's important that similar to the `CreateTunnel` struct, that we can convert this to/from protos. 

I'd expect the table scans to be a little more difficult, as we may need to create a connection during logical planning, then another connection during physical planning (or create a connection pool). something like that 🤷

crates/sqlexec/src/planner/physical_plan/create_tunnel_exec.rs
scsmithr: I actually like the idea of implementing the extensions. We currently have a lot of stuff shoved into `SessionContext` that isn't needed during the actual execution. The extensions would naturally break up stuff needed for planning vs stuff needed for execution (with some amount) of overlap.

And if we go this route, it'd make testing things like `CreateTable` easier since we'd only need a subset for the test.

crates/sqlexec/src/remote/planner.rs
```diff
@@ -86,3 +92,89 @@ impl PhysicalPlanner for RemotePhysicalPlanner {
...
+}
+
+impl<'a> CustomPhysicalPlanner<'a> {
+    pub fn new(ctx: &'a SessionContext) -> Self {
+        Self { ctx }
+    }
+}
+
+#[async_trait]
+impl<'a> PhysicalPlanner for CustomPhysicalPlanner<'a> {
```
scsmithr, line 108: Should be able to implement `ExtensionPlanner` to avoid needing to implement all the nodes: https://github.com/apache/arrow-datafusion/blob/e253b8e505d22dc4b427109917470db39aa3815b/datafusion/core/src/physical_planner.rs#L395-L416

universalmind303: I was thinking about that. I think we should try to start with the `ExtensionPlanner`. We'll likely end up having to do a completely custom physical planner at some point. But if we can get away with the ExtensionPlanner for now, that'd be great!

crates/sqlexec/src/planner/physical_plan/create_tunnel_exec.rs
universalmind303: @scsmithr for these extensions, won't we have to throw them into an `Arc<Mutex>` though? 

as of right now, all of our DDL execs just mutate the catalog then return an `EmptyExec`. So wouldn't we need a mutable reference to `catalog` inside the exec?

scsmithr: Yeah we would need to wrap in a mutex. Datafusion will automatically wrap in in Arc.

Mutex shouldn't be an issue as they'll be relatively uncontended .

PR #1559, opened 2023-08-15
feat: Physical execs for client batch exchange

scsmithr:
Adds in some of the pieces needed for record batch exchange. None of it's integrated currently, as we need to work on getting physical plans serializable.

Copied from slack:

> And as far as what this looks like with client exchange, I pushed up a LocationAwareTableProvider in the broadcast branch, with the idea we use that wrap all existing table providers. Table providers that require record batches be sent from the client will return a ClientExchangeRecvExec instead of the table provider itself.So that flow would look something like this:
> 
> 1. Do all table planning, wrap everything in a LocationAwareTableProvider. Data frames and local files will have the Client scan location.
> 2. For each table provider that has a Client scan location, create a ClientExchangeSendExec using the broadcast id in the table provider. This is what's responsible for actually reading the table.
> 3. Send physical plan to server (which will include ClientExchangeRecvExec plans), begin executing.
> 4. On client side, execute all ClientExchangeSendExec plans in their own tokio tasks which will stream up batches that the server needs.
> 
> Later we can add a physical optimizer rule to pull up ClientExchangeRecvExec plans. E.g. if we're joining two "client" tables, we'll send the result of the join, and not the tables individually.


crates/sqlexec/src/remote/broadcast/exchange_exec.rs
```diff
@@ -0,0 +1,499 @@
...
+    fn schema(&self) -> Arc<Schema> {
+        self.schema.clone()
+    }
+}
+
+/// The actual execution plan for reading batches from the client.
+#[derive(Debug)]
+pub struct ClientExchangeRecvExec {
+    schema: Arc<Schema>,
+    stream: Mutex<Option<ClientExchangeRecvStream>>,
```
universalmind303, line 174: won't this make it impossible to serialize the physical plan?

scsmithr: Sort of. I'm currently approaching this where the logical plans I'm defining for this are the "serializable" physical plans right now (https://github.com/GlareDB/glaredb/pull/1559/files#diff-9fb769d3f9c2de52b94a5d19ad4b3a25c53fcf5aed31ef06c5eb44e163b0e3c6R85-R88). When we have the proper extensions ready, we could make the appropriate changes to this where get the stream from the task context which would remove the need for this.

scsmithr: E.g. we would make an extension that would get us this: https://github.com/GlareDB/glaredb/pull/1559/files#diff-e3453d481fa630bb58c1e5fe35d4730223f55457bdb744b654f42b60a9639541R30-R33

scsmithr: Actually I'll go ahead and make that change right now to see how well it works.

crates/sqlexec/src/planner/logical_plan/client_exchange.rs
```diff
@@ -0,0 +1,162 @@
...
+            inner: Some(extension),
+        };
+
+        lp_extension
+            .encode(buf)
+            .map_err(|e| internal!("{}", e.to_string()))?;
+
+        Ok(())
+    }
+}
```
universalmind303, line 83: so i know we don't have the physical planning fully in place yet, but i don't think we'd want a `ClientExchangeSend` node at the logical level. Just checking to see if this is just part of the draft, or do you see a case for having some of this information at the logical level?

scsmithr: Yep I agree. Fully plan on these getting removed when the physical plan stuff is in place.

crates/sqlexec/src/remote/broadcast/mod.rs
universalmind303, line 1: one small suggestion, IMO the broadcast exec shouldn't be part of the `remote` module, but under a new `planner/physical_plan` module similar to what i did with all of the logical nodes. 

I think it'd be much easier to grok all of physical execs if they were in one directory instead of splitting them up based on remote/local. 

No strong feelings if you prefer keeping it here, Just my 2 cents.

scsmithr: will update

crates/rpcsrv/src/handler.rs
```diff
@@ -168,6 +171,22 @@ impl RpcHandler {
...
+        req: Streaming<service::BroadcastExchangeRequest>,
+    ) -> Result<BroadcastExchangeResponse> {
+        let stream = ClientExchangeRecvStream::try_new(req).await?;
+        let session_id = stream.session_id();
+        let session = self.get_session(session_id)?;
+        info!(session_id=%session_id, broadcast_id=%stream.broadcast_id(), "beginning client exchange stream");
+
+        session.register_broadcast_stream(stream).await?;
+
+        // TODO: We might need to await here for stream completion.
```
scsmithr, line 185: This should be easy enough with a oneshot if it's actually needed.

crates/rpcsrv/src/proxy.rs
```diff
@@ -229,3 +245,31 @@ impl<A: ProxyAuthenticator + 'static> service::execution_service_server::Executi
...
+    type Item = M;
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        match self.inner.poll_next_unpin(cx) {
+            Poll::Ready(Some(Ok(m))) => Poll::Ready(Some(m)),
+            Poll::Ready(Some(Err(e))) => {
+                // Don't know what we want to do yet, so just log and close down
+                // the stream.
+                warn!(%e, "received error when proxying request stream");
+                Poll::Ready(None)
+            }
```
scsmithr, line 270: I'm not completely sure what we want to do here. The error would be a connection error, and so returning `Poll::Ready(None)` to end the stream seems reasonable right now.

crates/sqlexec/src/context.rs
```diff
@@ -130,7 +131,10 @@ impl SessionContext {
 
         config_opts.catalog = catalog_opts;
         config_opts.optimizer = optimizer_opts;
-        let config: SessionConfig = config_opts.into();
+        let mut config: SessionConfig = config_opts.into();
+
+        // Add in custom extensions. These will be accessible during execution.
+        config = config.with_extension(Arc::new(StagedClientStreams::default()));
```
scsmithr, line 137: Indiscriminately added right now, but shouldn't affect anything.

crates/sqlexec/src/planner/physical_plan/client_send.rs
```diff
@@ -0,0 +1,274 @@
...
+use std::task::{Context, Poll};
+use uuid::Uuid;
+
+/// Execution plan for sending batches to a remote node.
+#[derive(Debug)]
+pub struct ClientExchangeInputSendExec {
+    broadcast_id: Uuid,
+    client: RemoteSessionClient, // TODO: Extension
+    input: Arc<dyn ExecutionPlan>,
+}
```
scsmithr, line 30: Adding in an extension won't be necessary right now since these will always be executed on the client side (and so will always have the client available).

crates/sqlexec/src/planner/physical_plan/mod.rs
scsmithr, line 1: Likely we'll need to move this module (and some other stuff in the `remote` module) into `datafusion_ext`.

crates/sqlexec/src/remote/staged_stream.rs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        // Handle case where stream arrived first.
+        streams.insert(id, PendingStream::StreamArrivedFirst(stream));
+    }
+
+    /// Resolve a pending stream by id.
+    ///
+    /// This will wait until we have the stream available.
+    pub fn resolve_pending_stream(&self, id: Uuid) -> ResolveStreamFut<S> {
```
universalmind303, line 61: wouldn't we want to do some housekeeping after the stream is resolved & clear the entry from `StagedStreams`?

scsmithr: Yep we're removing from the map here.

We will need to add in something to remove stale streams (e.g. server starts executing, inserts into the map, then client errors causing the entry to never be removed).

crates/sqlexec/src/remote/location_aware.rs
```diff
@@ -0,0 +1,109 @@
...
+}
+
+impl LocationAwareTableProvider {
+    pub fn new(inner: Arc<dyn TableProvider>, location: ScanLocation) -> Self {
+        Self { inner, location }
+    }
+}
+
+#[async_trait]
+impl TableProvider for LocationAwareTableProvider {
```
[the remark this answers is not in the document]
scsmithr: Not yet. I wanted to avoid making a large change that started to touch other bits of the code. Ideally we'll be returning `LocationAwareTableProvider` instead of `Arc<dyn TableProvider>` from our table code, so that at a higher level we could do something like:

```rust
// Create associated send execs
let send_execs: Vec<_> = location_aware_tables
  .iter()
  .filter(|t| t.needs_client_scan())
  .map(|t| t.create_client_scan_send_exec())
  .collect();

// Do planning, serialize and send physical plans to remote here...

let mut handles = Vec::new();
for send_exec in send_execs {
  let stream = send_exec().execute(...)?;
  handles.push(tokio::spawn(async move {
    while let Some(result) = stream.next().await {
      let _ = result?; // The stream just returns number of rows sent.
    }
    Ok(())
  }));
}

// Join handles...
```

crates/sqlexec/src/remote/location_aware.rs
```diff
@@ -0,0 +1,109 @@
...
+        /// ID to use when constructing the recv exec.
+        ///
+        /// When the server begins executing, it will attempt to pull from the
+        /// recv exec. The client will be responsible for sending record batches
+        /// to the server using this ID.
+        broadcast_id: Uuid,
+    },
+    /// Server should scan the table.
+    Server,
+}
```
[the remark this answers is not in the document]
scsmithr: I was thinking doing it in our dispatch code since the rules are pretty simple. If table is a local file scan or data frame, set scan location to client. Otherwise set to server.

We'll need additional logic in this `scan` function to avoid creating the client exec if we're not actually connected to a remote node.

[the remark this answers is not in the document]
universalmind303: that makes sense. as of right now, i cant think of any explicit reasons that logic would **need** to be in the planner.  Since that's probably the simplest, i'd say go with it. 

However...

 I do think longterm it would be the most fitting in the physical/extension planner. I don't think a logical plan should know anything about client/server relationships.

adhish20 (approved):

crates/sqlexec/src/remote/location_aware.rs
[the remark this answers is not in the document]
scsmithr: I generally agree with wanting this in the planner, and I looked a bit at this and see two road blocks:

1. If we continue returning `Arc<dyn TableProvider>`, we'd need to try to downcast to every possible table that should be scanned by the client. And while we only care about data frames and local file scans for that, we'd need to make sure that both have a unique and well-known type. The data frame would be easy, but local scanning would be a bit trickier, since it'd use the `ObjStoreTableProvider` type, which is _also_ used for s3 and gcs.
2. The default planner doesn't give us a good extension point for custom table sources, so to swap out the providers, we'd need to walk the plan prior to physical planning

And now that I'm writing this out, I think both of these are pretty small roadblocks that are easily solvable... I'll try to avoid littering the code with `LocationAwareTableProvider` and see if that can all be centralized to planning.

vrongmeal (approved):
This looks good in general to me. Excited to see this integrate!

PR #1562, opened 2023-08-16
feat: support record, struct, big numeric type for bigquery

### LINKED ISSUE #1521: BigQuery: Unsupported type: Record
## Description

We should go ahead and implement the rest of the types.

adhish20:
Fixes #1521

### DIFF
--- a/crates/datasources/src/bigquery/mod.rs
+++ b/crates/datasources/src/bigquery/mod.rs
@@ -506,7 +506,7 @@ fn table_field_schema_to_arrow_datatype(field: &BigQuerySchema) -> Result<Field>
         FieldType::Timestamp => DataType::Timestamp(TimeUnit::Nanosecond, Some("UTC".into())),
         FieldType::Time => DataType::Time64(TimeUnit::Nanosecond),
         FieldType::Numeric => DataType::Decimal128(38, 9),
-        FieldType::Bignumeric => DataType::Decimal256(76, 9),
+        FieldType::Bignumeric => DataType::Decimal256(76, 38),
         FieldType::Geography => DataType::Utf8,
         FieldType::Record | FieldType::Struct => {
             let mut record: Vec<Field> = Vec::new();
--- a/crates/datasources/src/bigquery/mod.rs
+++ b/crates/datasources/src/bigquery/mod.rs
@@ -10,8 +10,8 @@ use bigquery_storage::yup_oauth2::{
     ServiceAccountAuthenticator,
 };
 use bigquery_storage::{BufferedArrowIpcReader, Client};
-use datafusion::arrow::ipc::reader::StreamReader as ArrowStreamReader;
 use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::arrow::{datatypes::Fields, ipc::reader::StreamReader as ArrowStreamReader};
 use datafusion::datasource::TableProvider;
 use datafusion::error::{DataFusionError, Result as DatafusionResult};
 use datafusion::execution::context::SessionState;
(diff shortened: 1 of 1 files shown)


crates/datasources/src/bigquery/mod.rs
```diff
@@ -476,37 +477,55 @@ fn bigquery_table_to_arrow_schema(table: &Table) -> Result<ArrowSchema> {
...
+                for field in fields.iter() {
+                    let resolved_field = table_field_schema_to_arrow_datatype(field);
+                    if let Ok(resolved_field) = resolved_field {
+                        record.push(resolved_field);
+                    }
+                }
+            }
+            DataType::Struct(Fields::from(record))
+        }
+        other => return Err(BigQueryError::UnsupportedBigQueryType(other.clone())),
```
scsmithr, line 524: I think the last type we'd need to support is array. Would that be something we could throw in?

scsmithr (approved):
We'll want to open an issue for testing these new types.

We're currently using a common set of data for our data source tests, but we'll want to start adding in tests for data types that are unique to each data source.

crates/datasources/src/bigquery/mod.rs
adhish20: The only unsupported type left is Json

crates/datasources/src/bigquery/mod.rs
```diff
@@ -476,37 +477,55 @@ fn bigquery_table_to_arrow_schema(table: &Table) -> Result<ArrowSchema> {
...
+        // aim to work only with nanoseconds to have uniformity accross the
+        // codebase. It's also easier to have interop with datafusion since
+        // with many things like type inference datafusion uses nanosecond.
+        // This cast is done when the stream is received by using the
+        // `crate::common::util::normalize_batch` function.
+        FieldType::Datetime => DataType::Timestamp(TimeUnit::Nanosecond, None),
+        FieldType::Timestamp => DataType::Timestamp(TimeUnit::Nanosecond, Some("UTC".into())),
+        FieldType::Time => DataType::Time64(TimeUnit::Nanosecond),
+        FieldType::Numeric => DataType::Decimal128(38, 9),
+        FieldType::Bignumeric => DataType::Decimal256(76, 9),
```
vrongmeal, line 509: This was intentionally omitted, but it's okay. Using big numeric will still cause errors since we don't know how to process these decimals.

vrongmeal: Also, Is it for sure `(76, 9)` ?

adhish20: /// The maximum precision for [DataType::Decimal128] values
pub const DECIMAL128_MAX_PRECISION: u8 = 38;

/// The maximum scale for [DataType::Decimal128] values
pub const DECIMAL128_MAX_SCALE: i8 = 38;

/// The maximum precision for [DataType::Decimal256] values
pub const DECIMAL256_MAX_PRECISION: u8 = 76;

/// The maximum scale for [DataType::Decimal256] values
pub const DECIMAL256_MAX_SCALE: i8 = 76;

vrongmeal: Oh yeah, I meant is the default scale 9 here as well?

adhish20: default scale is 10 for both Decimal128 and Decimal256

adhish20: https://cloud.google.com/bigquery/docs/reference/storage#arrow_schema_details

It's 38 for big numeric

vrongmeal:
Looks good. Just need tests for new data types. We can add erroring tests for the ones that don't yet work.

vrongmeal (approved):

PR #1572, opened 2023-08-17
feat: Allow reconstructing postgres table provider and execs

scsmithr:
Small refactor on the postgres stuff which allow reconstructing a postgres table provider and/or the binary copy exec with serializable values.

Started this because the broadcast/exchange stuff will be a lot easier when working with physical plans, and so I want to try to get the physical plan serialization done quickly. Other data sources should be able to follow this pattern of if we have the client, use that, otherwise create one.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1760,6 +1760,7 @@ dependencies = [
  "datafusion",
  "decimal",
  "env_logger",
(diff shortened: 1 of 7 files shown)


vrongmeal (approved):
Looks good 🚀

PR #1576, opened 2023-08-17
feat: Client side table planning and execs

scsmithr:


crates/sqlexec/src/planner/physical_plan/send_recv.rs
```diff
@@ -0,0 +1,220 @@
...
+        // Will be empty if this isn't the first call to execute, which is fine
+        // because we only want these handled once.
+        let send_execs: Vec<ClientExchangeSendExec> =
+            std::mem::take(self.send_execs.lock().as_mut());
+
+        let mut join_set: JoinSet<Result<(), DataFusionError>> = JoinSet::new();
+        for send_exec in send_execs {
+            let context = context.clone();
+            join_set.spawn(async move {
+                let mut stream = send_exec.execute(0, context)?;
```
vrongmeal, line 106: We "know" here that a `send_exec` has one output partition. Would this be a potentially dangerous assumption? 🤔

vrongmeal: In `ClientExchangeSendExec`, is it really safe to assume it's only one partition? There it seems even more doubtful.
```
98: let input = self.input.execute(0, context)?;
```

crates/sqlexec/src/remote/exec.rs
```diff
@@ -100,7 +100,7 @@ impl ExecutionPlan for RemoteExecutionPlan {
     }
 }
 
-impl DisplayAs for RemoteExecutionPlan {
+impl DisplayAs for RemoteExecutionExec {
     fn fmt_as(&self, _t: DisplayFormatType, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "RemoteExecutionPlan")
```
vrongmeal, line 105:
```suggestion
        write!(f, "RemoteExecutionExec")
```

crates/sqlexec/src/remote/planner.rs
```diff
@@ -45,13 +45,13 @@ impl QueryPlanner for RemoteLogicalPlanner {
 #[derive(Debug, Clone)]
 pub struct RemotePhysicalPlanner {
     /// Client to remote services.
-    remote_client: RemoteSessionClient,
+    _remote_client: RemoteSessionClient,
```
vrongmeal, line 48: Why not remove this?

crates/sqlexec/src/planner/physical_plan/send_recv.rs
scsmithr: It's currently assumed that we're only working with a single partition across `ClientExchangeSendExec` and `ClientExchangeRecvExec`: https://github.com/GlareDB/glaredb/blob/main/crates/sqlexec/src/planner/physical_plan/client_send.rs#L85.

We could pretty easily extend this though to handle more partitions. Instead of of creating one tokio task per send exec, we would create one _per partition_ per send exec. I believe the code changes would be minimal for this. `ClientExchangeRecvExec` would need to be tweaked to take in a vector of broadcast ids and pull from all streams during execution. Right now it just assumes a single partition (using a single broadcast id).

crates/sqlexec/src/remote/planner.rs
scsmithr: The planner stuff is going to change quite a bit. I'm working on splitting up session context here https://github.com/GlareDB/glaredb/pull/1589 because I was getting confused about what's execution where. By splitting the session context, it'll be easier to tell when we need a custom planner (and it'll probably clean up some other confusing logic too).

vrongmeal (commented):
Correct me if I'm wrong, but we potentially want to get rid of `RemotePhysicalExec` once we start sending the serialized physical plans.

scsmithr:
> Correct me if I'm wrong, but we potentially want to get rid of `RemotePhysicalExec` once we start sending the serialized physical plans.

Yeah I think we'd be able to remove. By just having serialized physical plans, we actually remove a lot of logic from the remote side. All it needs to do is deserialize the plan and execution. There's no planning involved. 

With the session split I'm working on, the entirety of the logic needed on the remote side is:
- Getting table providers for external tables/databases (note that I plan on removing the `scan` and `insert_into` impls since those will never actually be called, we just need the provider for the schema).
- Executing a physical plan directly.

You can see that here: https://github.com/GlareDB/glaredb/pull/1589/files#diff-eea77af3fa72ef01be4f33d5f3191f228a67826580a528d979949c8801673ca0

PR #1578, opened 2023-08-17
chore: move sesionvars into df context

universalmind303:


crates/datafusion_ext/src/vars/constants.rs
universalmind303, line 1: a lot of this is just splitting up the `vars` into a few smaller files. it was getting very hard to navigate.

crates/datafusion_ext/src/vars.rs
```diff
@@ -1,741 +1,222 @@
...
-    pub remote_session_id: SessionVar<Option<Uuid>>,
-    pub user_id: SessionVar<Uuid>,
-    pub user_name: SessionVar<str>,
-    pub database_name: SessionVar<str>,
-    pub max_datasource_count: SessionVar<Option<usize>>,
-    pub memory_limit_bytes: SessionVar<Option<usize>>,
-    pub max_tunnel_count: SessionVar<Option<usize>>,
-    pub max_credentials_count: SessionVar<Option<usize>>,
-    pub is_cloud_instance: SessionVar<bool>,
+    inner: Arc<RwLock<SessionVarsInner>>,
```
universalmind303, line 25: `RwLock` seemed like a better fit than `Mutex` as it's primarily read focused.

crates/datafusion_ext/src/vars/inner.rs
```diff
@@ -0,0 +1,370 @@
...
+use super::constants::*;
+use super::error::VarError;
+use super::value::Value;
+use std::sync::Arc;
+use tracing::error;
+use uuid::Uuid;
+
+/// Variables for a session.
+#[derive(Debug)]
+pub struct SessionVarsInner {
```
universalmind303, line 18: this is mostly unchanged from the original `SessionVars`.

the new `SessionVars` is a wrapper around this that provides some read/write utilities around this inner struct.

crates/sqlexec/src/planner/context_builder.rs
```diff
@@ -126,8 +126,8 @@ impl<'a> PartialContextProvider<'a> {
     ) -> Option<Arc<dyn TableFunc>> {
         let catalog = self.ctx.get_session_catalog();
 
-        let resolve_func = |schema, name| {
-            match catalog.resolve_entry(DEFAULT_CATALOG, schema, name) {
+        let resolve_func = |schema: String, name| {
```
scsmithr, line 129: Could change this into a function to allow something like `fn resolve(schema: impl Into<String>, name: impl Into<String>) -> ...`

scsmithr (approved):
Looks reasonable to me. If tests pass, then good to go!

crates/sqlexec/src/planner/context_builder.rs
universalmind303: I'm going to be working on the catalog next, will likely have to refactor some of these methods anyways.

PR #1585, opened 2023-08-18
refactor: move catalog into a datafusion extension

universalmind303:
So this ended up being a bit trickier than the session vars. 

The context is used within async/await a lot more than the vars were & caused the `parking_lot::RwLock` to not work. I had to swap it out with the `tokio::sync::RwLock` which unfortunately causes all read/write actions to be async. This caused qquite a ripple effect as all dependent function calls in the stack also had to be converted to async/await. 

I also added a small example exec `CreateTableExec`, it's not fully wired up yet, but it does show that we can mutably access the catalog & session vars from the exec now!

you can test it out by swapping out the planner in `sqlexec/src/session.rs:create_physical_plan` for the extension aware planner. 

There's a few small todo's & unwraps left in here, but I think the bulk of the work is ready for review.

### DIFF
--- a/crates/sqlexec/src/metastore/catalog.rs
+++ b/crates/sqlexec/src/metastore/catalog.rs
@@ -7,6 +7,7 @@ use protogen::metastore::types::options::TableOptions;
 use protogen::metastore::types::service::Mutation;
 use std::collections::HashMap;
 use std::sync::Arc;
+use tokio::sync::{RwLock, RwLockReadGuard, RwLockWriteGuard};
 use tracing::debug;
 
 use super::client::{SupervisorClient, WorkerError};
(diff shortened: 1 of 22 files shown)


crates/sqlexec/src/planner/preprocess.rs
```diff
@@ -33,14 +32,15 @@ impl<'a> ast::VisitorMut for CastRegclassReplacer<'a> {
...
                 // TODO
-                if let Some(ent) = catalog.resolve_entry(DEFAULT_CATALOG, &schema, rel) {
-                    // Table found.
-                    return Some(ent.get_meta().id);
-                }
+                todo!("IS IT POSSIBLE TO DO THIS WITHOUT THE CATALOG?")
+                // if let Some(ent) = catalog.resolve_entry(DEFAULT_CATALOG, &schema, rel).await {
+                //     // Table found.
+                //     return Some(ent.get_meta().id);
+                // }
```
scsmithr, line 43: No, this isn't possible.

crates/sqlexec/src/metastore/catalog.rs
```diff
@@ -47,6 +48,98 @@ pub struct SessionCatalog {
     /// Map schema IDs to objects in the schema.
     schema_objects: HashMap<u32, SchemaObjects>,
 }
+pub struct AsyncSessionCatalog {
+    inner: Arc<RwLock<SessionCatalog>>,
+}
```
scsmithr, line 53: The introduction of async to all these methods is going to be very difficult to resolve (and not needed for most this). The only things that need async is `maybe_refresh_state` and `mutate`, and that's just due to the client.

Some changes we can make here:
- Move `metastore_client` to this struct.
- Move implementation of `mutate` and `maybe_refresh_state` to this struct.
  - Scope locking down to the `set_state` and `version` calls.
- Replace tokio lock with parking_lot.

Once that's in, we can remove the asyncs on the functions that don't need them.

scsmithr (commented):
I think it's important t

scsmithr:
I made a change that splits out the metastore client from the session catalog and puts it in a struct called `CatalogMutator`. We can stick that on the datafusion context.

universalmind303:
closed in favor of #1607

PR #1604, opened 2023-08-21
chore: Remove init catalog endpoint for metastore

scsmithr:
Removes needing to "init" a catalog before fetching or mutating it. Since we create catalogs lazily, the only thing "init" was doing was loading it into memory. This change makes it so that that other two endpoints just do that automatically if the catalog isn't found.

These errors should no longer happen:

```
Missing database catalog: cbdfc895-3d01-4771-92b6-9ab1c43a3bb9
```

(because I deleted it)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3313,6 +3313,7 @@ version = "0.4.0"
 dependencies = [
  "async-trait",
  "bytes",
+ "dashmap",
  "datafusion",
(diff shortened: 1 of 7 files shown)


universalmind303 (approved):

greyscaled (commented):
Awesome ✔️ !

Issue #968, opened 2023-05-11
Connection reset every time I try to query view


greyscaled:
## Context

This is exactly what was happening in my terminal:

```
grey_free=> \d
                    List of relations
    Schema     |         Name          | Type  |  Owner
---------------+-----------------------+-------+---------
 glare_catalog | columns               | table | unknown
 glare_catalog | databases             | table | unknown
 glare_catalog | schemas               | table | unknown
 glare_catalog | session_query_metrics | table | unknown
 glare_catalog | tables                | table | unknown
 glare_catalog | views                 | table | unknown
(6 rows)

grey_free=> select * from glare_catalog.views;
grey_free=> select * from public.datasources;
SSL connection has been closed unexpectedly
The connection to the server was lost. Attempting reset: Succeeded.
psql (14.7 (Ubuntu 14.7-0ubuntu0.22.04.1), server 15.1)
WARNING: psql major version 14, server major version 15.
         Some psql features might not work.
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-CHACHA20-POLY1305, bits: 256, compression: off)
grey_free=> select * from public.datasources;
SSL connection has been closed unexpectedly
The connection to the server was lost. Attempting reset: Succeeded.
psql (14.7 (Ubuntu 14.7-0ubuntu0.22.04.1), server 15.1)
WARNING: psql major version 14, server major version 15.
         Some psql features might not work.
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-CHACHA20-POLY1305, bits: 256, compression: off)
grey_free=>
grey_free=> select * from public.datasources;
SSL connection has been closed unexpectedly
The connection to the server was lost. Attempting reset: Succeeded.
psql (14.7 (Ubuntu 14.7-0ubuntu0.22.04.1), server 15.1)
WARNING: psql major version 14, server major version 15.
         Some psql features might not work.
SSL connection (protocol: TLSv1.2, cipher: ECDHE-RSA-CHACHA20-POLY1305, bits: 256, compression: off)
grey_free=> select * from glare_catalog.views;
grey_free=> select * from public.datasources;
SSL connection has been closed unexpectedly
The connection to the server was lost. Attempting reset: Succeeded.
psql (14.7 (Ubuntu 14.7-0ubuntu0.22.04.1), server 15.1)
WARNING: psql major version 14, server major version 15.
         Some psql features might not work.
[...]

PR #1607, opened 2023-08-21
refactor: async catalog

universalmind303:
so some of the new code actually broke our interaction with datafusion.

specifically within the `attach_remote_session` function. this mutably sets the catalog to a new catalog from the remote session. 

datafusion exposes no possible way to mutate the extensions after the state is initialized. This seemed kinda strange to me considering their numerous extension points. _i guess you could completely rebuild the state, but that seems like massive overkill_

this approach patches datafusion with a new function to register an extension on an existing state. (i'll open up a PR with datafusion if this is an acceptable approach. 

see for changes to datafusion. 
https://github.com/apache/arrow-datafusion/compare/28.0.0...universalmind303:arrow-datafusion:28-patch?expand=1

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1570,8 +1570,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "28.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5ddbcb2dda5b5033537457992ebde78938014390b2b19f9f4282e3be0e18b0c3"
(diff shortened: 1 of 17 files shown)


universalmind303:
Apparantly this PR causes a regression in #968 specifically in the `views.slt`. not sure why tho.. 

```

# View/CTE name conflict (#968)
#
# `external_datasources` makes use of a CTE named 'datasources'. Creating a view
# of the same name that referenced it previously resulted in a stack overflow.

statement ok
create view datasources as select * from glare_catalog.external_datasources;

statement ok
select * from datasources;
```

it works fine locally, but causes a SO when running in the SLT.

PR #1613, opened 2023-08-22
feat: DDL execs

universalmind303:
⚠️  this breaks the DDL's when running locally, as we aren't bypassing anything anymore & everything is delegated to the physical execs. but i think this should be fine as long as we finish up the rest of the DDL's today.


scsmithr:
> ⚠️ this breaks the DDL's when running locally, as we aren't bypassing anything anymore & everything is delegated to the physical execs. but i think this should be fine as long as we finish up the rest of the DDL's today.

I'm confident we can get it done.

scsmithr (approved):

crates/sqlexec/src/metastore/catalog.rs
```diff
@@ -436,24 +437,34 @@ impl NamespacedCatalogEntry<'_> {
 #[derive(Debug, Default)]
 pub struct TempObjects {
     /// In-memory (temporary) tables.
-    current_session_tables: HashMap<String, Arc<MemTable>>,
+    current_session_tables: RwLock<HashMap<String, Arc<MemTable>>>,
```
universalmind303, line 440: why not use `DashMap`? it's designed to be a replacement for this. 

> DashMap tries to be very simple to use and to be a direct replacement for RwLock<HashMap<K, V, S>>. To accomplish this, all methods take &self instead of modifying methods taking &mut self. This allows you to put a DashMap in an Arc<T> and share it between threads while being able to modify it.

crates/sqlexec/src/planner/logical_plan/delete.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fn try_decode_extension(_extension: &LogicalPlanExtension) -> Result<Self> {
+        unimplemented!()
+    }
+
+    fn try_encode(&self, _buf: &mut Vec<u8>, _codec: &dyn LogicalExtensionCodec) -> Result<()> {
+        unimplemented!()
+    }
+}
```
universalmind303, line 62: don't we need this implemented for (de)serialization of the node?

crates/sqlexec/src/planner/logical_plan/drop_tables.rs
```diff
@@ -1,7 +1,7 @@
 use super::*;
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct DropTables {
-    pub names: Vec<OwnedTableReference>,
+    pub references: Vec<OwnedFullObjectReference>,
```
universalmind303, line 4: IMO `references` is a little vague. can we have this be `tbl_references`?

crates/sqlexec/src/planner/logical_plan/delete.rs
adhish20: I thought we don't need to serialise and deserialise logical plan anymore?

crates/sqlexec/src/planner/logical_plan/drop_tables.rs
scsmithr: I'm good with renaming.

I made this change to enforce that the execs are fully planned and there's no ambiguity when running them remotely.

As a side note, we should look at making changes to metastore to execute these DDL plans directly instead of converting them to metastore specific messages.

crates/sqlexec/src/metastore/catalog.rs
scsmithr: DashMap would actually incur more overhead here, specifically when returning the names of all tables, as it would need to acquire multiple locks.

I think the rwlock here is fine since these are scoped to a single session, and so concurrent access is unlikely (so much so that I would actually prefer a plain mutex here, but that's minor).

crates/sqlexec/src/planner/logical_plan/drop_tables.rs
universalmind303: > As a side note, we should look at making changes to metastore to execute these DDL plans directly instead of converting them to metastore specific messages.

i like that idea. Now that the execs are fully serializable, it seems unnecessary to convert them to yet another struct just so metastore can handle them.

scsmithr:
Going to go ahead and merge this since SLTs are passing, and we can continue to work off main.

crates/sqlexec/src/planner/logical_plan/delete.rs
scsmithr: I think we'll likely be removing some (or all?) of this trait since we only care able serializing physical plans now.

crates/sqlexec/src/planner/logical_plan/drop_tables.rs
scsmithr: > IMO `references` is a little vague. can we have this be `tbl_references`?

Addressing this in https://github.com/GlareDB/glaredb/pull/1650

PR #1618, opened 2023-08-22
feat: Support RPC Inserts

vrongmeal:


crates/sqlexec/src/planner/physical_plan/remote_insert.rs
```diff
@@ -0,0 +1,128 @@
...
+    )]))
+});
+
+/// Executes an insert on the cached table provider on remote side.
+///
+/// Only valid to execute this on remote side.
+#[derive(Debug, Clone)]
+pub struct RemoteInsertExec {
+    pub provider: ProviderReference,
+    pub input: Arc<ClientExchangeRecvExec>,
```
scsmithr, line 33: This should be `Arc<dyn ExecutionPlan>`. Since it's executing on the remote side, it'll already have the plan ready to go and we don't need to do any client recv/send stuff.

crates/sqlexec/src/planner/physical_plan/remote_insert.rs
```diff
@@ -0,0 +1,128 @@
...
+        datafusion::physical_plan::Partitioning::UnknownPartitioning(1)
+    }
+
+    fn output_ordering(&self) -> Option<&[datafusion::physical_expr::PhysicalSortExpr]> {
+        None
+    }
+
+    fn children(&self) -> Vec<Arc<dyn ExecutionPlan>> {
+        Vec::new()
+    }
```
scsmithr, line 67: Should return `input`.

crates/sqlexec/src/planner/physical_plan/remote_insert.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    fn with_new_children(
+        self: Arc<Self>,
+        _children: Vec<Arc<dyn ExecutionPlan>>,
+    ) -> datafusion::error::Result<Arc<dyn ExecutionPlan>> {
+        Err(DataFusionError::Plan(
+            "Cannot replace children for RemoteInsertExec".to_string(),
+        ))
+    }
```
scsmithr, line 76: Should be able to change `input`.

crates/protogen/src/sqlexec/physical_plan.rs
```diff
@@ -26,9 +26,17 @@ pub struct RemoteScanExec {
     pub limit: Option<u64>,
 }
 
+#[derive(Clone, PartialEq, Message)]
+pub struct RemoteInsertExec {
+    #[prost(bytes, tag = "1")]
+    pub provider_id: Vec<u8>, // UUID
+    #[prost(message, tag = "2")]
+    pub input: Option<ClientExchangeRecvExec>,
```
scsmithr, line 34: We shouldn't be serializing `input` here.

crates/sqlexec/src/extension_codec.rs
```diff
@@ -357,6 +358,36 @@ impl<'a> PhysicalExtensionCodec for GlareDBExtensionCodec<'a> {
...
+                    .table_providers
+                    .expect("remote context should have provider cache")
+                    .get(&provider_id)
+                    .ok_or_else(|| {
+                        DataFusionError::Internal(format!("Missing proivder for id: {provider_id}"))
+                    })?;
+
+                Arc::new(RemoteInsertExec {
+                    provider: ProviderReference::Provider(provider),
+                    input: Arc::new(input),
```
scsmithr, line 388: We'll want to do `input: inputs[0].clone()` here instead of trying to deserialize it.

crates/sqlexec/src/extension_codec.rs
```diff
@@ -399,6 +430,23 @@ impl<'a> PhysicalExtensionCodec for GlareDBExtensionCodec<'a> {
...
+                    ))
+                }
+            };
+
+            proto::ExecutionPlanExtensionType::RemoteInsertExec(proto::RemoteInsertExec {
+                provider_id: id.into_bytes().to_vec(),
+                input: Some(proto::ClientExchangeRecvExec {
+                    broadcast_id: exec.input.broadcast_id.into_bytes().to_vec(),
+                    schema: Some(exec.input.schema.clone().try_into()?),
+                }),
```
scsmithr, line 448: See above, we shouldn't serialize the input here.

crates/sqlexec/src/remote/table.rs
```diff
@@ -86,10 +98,24 @@ impl TableProvider for StubRemoteTableProvider {
...
+        let recv = RemoteExecutionExec::new(
+            self.client.clone(),
+            Arc::new(RemoteInsertExec::from_broadcast(
+                broadcast_id,
+                ProviderReference::RemoteReference(self.provider_id),
+            )),
+        );
+
+        let exec = SendRecvJoinExec::from_send_execs(Arc::new(recv), vec![send]);
+        Ok(Arc::new(exec))
```
scsmithr, line 119: This should all be `Ok(Arc::new(RemoteInsertExec { provider: self.provider.clone(), input }))`.

Only place where we need to handle the client send/recv stuff is in the planner during rewriting.

scsmithr (commented):
I believe the error you're getting is from manually constructing the client send/recv stuff, and it being executed client side during planning. I would have to look at it more deeply to figure our if that really is the case, and where it's happening.

But only the planner should be putting client send/recv execs in the plan. We shouldn't need to worry about it elsewhere. Since this will be executing on the remote side, the remote node will already have the execution plan ready to go. That plan may or may not have client send/recv execs, but inserts shouldn't be aware of that.

PR #1622, opened 2023-08-22
chore: add physical plans for drop schemas, views and tunnel

adhish20:


crates/sqlexec/src/remote/planner.rs
```diff
@@ -145,9 +149,63 @@ impl ExtensionPlanner for DDLExtensionPlanner {
...
+                }
+                .unwrap();
+
+                let exec: DropSchemasExec = DropSchemasExec {
+                    catalog_version: self.catalog_version,
+                    names: drop_schemas_lp
+                        .names
+                        .into_iter()
+                        .map(|n| n.schema_name().to_string())
+                        .collect(),
```
scsmithr, line 165: Why do we need to iterate? Can we not do `drop_schemas_lp.name`?

scsmithr (approved):

crates/sqlexec/src/remote/planner.rs
adhish20: converting SchemaReference to String

PR #1633, opened 2023-08-25
chore: Copy To

### LINKED ISSUE #1600: Copy To physical plan
## Description

This one likely to be tricky.

vrongmeal:
Fixes #1600

### DIFF
--- a/crates/protogen/src/sqlexec/physical_plan.rs
+++ b/crates/protogen/src/sqlexec/physical_plan.rs
@@ -5,7 +5,10 @@ use crate::gen::metastore::catalog::TableEntry;
 use datafusion_proto::protobuf::{LogicalExprNode, Schema};
 use prost::{Message, Oneof};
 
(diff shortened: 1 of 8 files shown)


universalmind303:
I kinda expected that we'd end up with multiple execs for copy_to? Does this handle every scenario, or is it just our existing functionality implemented as an exec? 

If the latter, we should keep #1600 open to handle all the possible scenarios.

scsmithr:
> I kinda expected that we'd end up with multiple execs for copy_to? Does this handle every scenario, or is it just our existing functionality implemented as an exec?
> 
> If the latter, we should keep #1600 open to handle all the possible scenarios.

I believe this is just to get the existing functionality in as an exec.

We will require some changes (possibly at the planner level) so support the remote/local stuff. I think we'll end up wanting to switch out `dest` to be a `DataSink` (datafusion might already have a trait we could use).

scsmithr (approved):

PR #1656, opened 2023-08-28
chore: Add additional tests for asserting command tag output

scsmithr:
This also adds the ability to ignore selected messages from the backed. For example, Postgres will return a NoticeResponse for the query `create schema if not exists text_schema` indicating if the schema was actually created, but we don't yet have this feature. Since notices are purely informational, it's safe for us to ignore those messages.

What this looks like:

```
send
Query {"query": "create schema if not exists test_schema"}
----

until NoticeResponse=ignore
ReadyForQuery
----
CommandComplete {"tag":"CREATE SCHEMA"}
ReadyForQuery {"status":"I"}
```


testdata/pgprototest/ddl.pt
scsmithr, line 1: Renamed file.

PR #1660, opened 2023-08-29
feat: Update catalog with rpc

scsmithr:
This change makes it so that the local session will pull the catalog after it detects the query was a DDL query.

```
GlareDB (v0.4.0)
Type \help for help.
Using in-memory catalog
> \open http:localhost:6542
> create table hello7 (a int);
Table created
> insert into hello7 select 1 as a;
Inserted 1 row
> insert into hello7 select 2 as a;
Inserted 1 row
> select * from hello7;
┌────────────┐
│ $operation │
│ ──         │
│ Utf8       │
╞════════════╡
│ 1          │
│ 2          │
└────────────┘
```

Other changes:

- Reverts some of the execution stream/result stuff. It made working with the results pretty clumsy.
- Adds a quick-fix around querying system tables by wrapping it in a local hint. This was useful for debugging.
- Reinits the datafusion context on client attach. This was done to create a catalog mutator that didn't have a metastore client. Without this change, the catalog would inadvertently revert back to the local session's catalog since the session will attempt to update the catalog from the local cache with the client.

Bugs:

- The above example, the column name is incorrect. I'm not sure why yet, but plan on looking into it in a follow up.
- The memory store not working is very inconvenient for testing. The above example was done by specifying a data dir.

### DIFF
--- a/crates/sqlexec/src/context/remote.rs
+++ b/crates/sqlexec/src/context/remote.rs
@@ -7,12 +7,15 @@ use datafusion::{
 };
 use datafusion_ext::vars::SessionVars;
 use datasources::native::access::NativeTableStorage;
+use protogen::{
+    metastore::types::catalog::CatalogEntry, rpcsrv::types::service::ResolvedTableReference,
+};
 use uuid::Uuid;
 
 use crate::{
     background_jobs::JobRunner,
     dispatch::external::ExternalDispatcher,
-    errors::Result,
+    errors::{ExecError, Result},
     extension_codec::GlareDBExtensionCodec,
     metastore::catalog::{CatalogMutator, SessionCatalog, TempCatalog},
     remote::{provider_cache::ProviderCache, staged_stream::StagedClientStreams},
@@ -80,6 +83,13 @@ impl RemoteSessionContext {
         &self.catalog
     }
 
+    pub async fn refresh_catalog(&mut self) -> Result<()> {
+        self.catalog
+            .maybe_refresh_state(self.catalog_mutator().get_metastore_client(), false)
+            .await?;
+        Ok(())
+    }
+
     /// Returns the extension codec used for serializing and deserializing data
     /// over RPCs.
     pub fn extension_codec(&self) -> GlareDBExtensionCodec<'_> {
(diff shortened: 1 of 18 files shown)


vrongmeal (approved):

vrongmeal:
> This change makes it so that the local session will pull the catalog after it detects the query was a DDL query.

Is this sufficient? Given it's a cloud instance, multiple people might be using it.

vrongmeal:
**Local Instance 1:**

```
> create table hello (a int);
Table created
> insert into hello select 1;
Inserted 1 row
> select * from hello;
┌────────────┐
│ $operation │
│ ──         │
│ Utf8       │
╞════════════╡
│ 1          │
└────────────┘
```

**Local Instance 2:**

```
> alter table hello rename to bye;
Table renamed
> select * from bye;
┌────────────┐
│ $operation │
│ ──         │
│ Utf8       │
╞════════════╡
│ 1          │
└────────────┘
```

**Local Instance 1:**

```
> select * from hello;
┌────────────┐
│ $operation │
│ ──         │
│ Utf8       │
╞════════════╡
│ 1          │
└────────────┘
> select * from bye;
Error: Error during planning: Unable to fetch table provider for 'bye': failed to resolve: failed to find table: bye
```

The above steps are in order.

crates/sqlexec/src/planner/context_builder.rs
```diff
@@ -111,21 +112,32 @@ impl<'a> PartialContextProvider<'a> {
...
-                    // TODO: Probably want to get the fully qualified reference.
                     let meta = ent.get_meta();
                     if !(meta.is_temp || meta.builtin) {
-                        let owned = reference.to_owned_reference();
-                        let prov = client.dispatch_access(owned).await?;
+                        let prov = client
+                            .dispatch_access(ResolvedTableReference::Internal {
+                                table_oid: meta.id,
+                            })
+                            .await?;
```
vrongmeal, line 140: We're sending an external table as `ResolvedTableReference::Internal` but not resolving it as an external table in `context/remote.rs`.

crates/sqlexec/src/context/remote.rs
```diff
@@ -122,23 +132,41 @@ impl RemoteSessionContext {
...
-        let prov: Arc<dyn TableProvider> =
-            if let Some(tbl) = self.catalog.resolve_table(database, schema, name) {
-                self.tables.load_table(tbl).await?.into_table_provider()
-            } else {
+        let prov: Arc<dyn TableProvider> = match table_ref {
+            ResolvedTableReference::Internal { table_oid } => {
+                match self.catalog.get_by_oid(table_oid) {
+                    Some(CatalogEntry::Table(tbl)) => {
+                        self.tables.load_table(tbl).await?.into_table_provider()
+                    }
```
vrongmeal, line 146: Here we assume any `ResolvedTableReference::Internal` is a native table.

vrongmeal (changes requested):
We're not resolving external tables anywhere.

scsmithr:
> > This change makes it so that the local session will pull the catalog after it detects the query was a DDL query.
> 
> Is this sufficient? Given it's a cloud instance, multiple people might be using it.

We can add in a check that if the result is an error, then refresh the catalog. We don't have a lot of detail about why a query would have errored right now, so that check would happen on every error. I think we can look into adding more detail in the response stream when executing a physical plan that would give us more info without having to inspect the error.

crates/sqlexec/src/planner/context_builder.rs
scsmithr: Will fix.

vrongmeal (approved):
Good to go 🚀

PR #1664, opened 2023-08-29
fix(rpc): Use correct schema when serialize create table execs

### LINKED ISSUE #1661: RPC: Incorrect schema used when inserting batches
## Description

```
GlareDB (v0.4.0)
Type \help for help.
Using in-memory catalog
> \open http:localhost:6542
> create table hello7 (a int);
Table created
> insert into hello7 select 1 as a;
Inserted 1 row
> insert into hello7 select 2 as a;
Inserted 1 row
> select * from hello7;
┌────────────┐
│ $operation │
│ ──         │
│ Utf8       │
╞════════════╡
│ 1          │
│ 2          │
└────────────┘
```

Column name should be `a`.

scsmithr:
Closes #1661 

Was using the schema of the _output_ of the create table execs, and the schema for the table itself.

```
> \open http://localhost:6542
> create table hello (a int, b int, c int);
Table created
> insert into hello select 1, 2, 3;
Inserted 1 row
> select * from hello;
┌───────┬───────┬───────┐
│ a     │ b     │ c     │
│ ──    │ ──    │ ──    │
│ Int32 │ Int32 │ Int32 │
╞═══════╪═══════╪═══════╡
│ 1     │ 2     │ 3     │
└───────┴───────┴───────┘
```

### DIFF
--- a/testdata/sqllogictests/rpc.slt
+++ b/testdata/sqllogictests/rpc.slt
@@ -52,15 +52,16 @@ Kathryn  Sweden
 
 # One off tests for #1599
 
-# statement ok
-# create schema hello_world;
+statement ok
+create schema hello_world;
 
-# statement ok
-# drop schema hello_world;
+statement ok
+drop schema hello_world;
 
+# Views don't seem to be working yet.
 # statement ok
 # create view my_view as select 1;
-
+#
 # query I
 # select * from my_view;
 # ----
@@ -77,3 +78,14 @@ select * from test2;
 ----
 1
 
+# Test for #1661
+statement ok
+create table t1661 (a int, b int, c int);
+
+statement ok
+insert into t1661 select 1, 2, 3;
+
+query III
+select * from t1661;
+----
+1 2 3
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -131,8 +131,8 @@ jobs:
 
       - name: SQL Logic Tests (RPC)
         run: |
-          just sql-logic-tests -v 'sqllogictests/rpc' \
-                                  'sqllogictests/time'
+          just sql-logic-tests -v --rpc-test 'sqllogictests/rpc' \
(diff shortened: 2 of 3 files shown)


testdata/sqllogictests/rpc.slt
```diff
@@ -77,3 +78,14 @@ select * from test2;
...
 
+# Test for #1661
+statement ok
+create table t1661 (a int, b int, c int);
+
+statement ok
+insert into t1661 select 1, 2, 3;
+
+query III
+select * from t1661;
```
vrongmeal, line 89: Shouldn't the query here be `select a, b, c from t1661`? Even then Not sure how this test works for #1661 🤔 

Or do I understand the issue incorrectly?

scsmithr: The issue here was we using the schema of the output of the operation (which in this case is a schema with a single column named `$operation`).

To illustrate, no matter which schema I provided during the create table, it was always created with the incorrect schema:

```
> \open http://localhost:6542
> create table hello (a int, b int, c int);
Table created
> insert into hello select 1, 2, 3;
Error: Error during planning: Column count doesn't match insert query!
> insert into hello select 1;
Inserted 1 row
> select * from hello;
┌────────────┐
│ $operation │
│ ──         │
│ Utf8       │
╞════════════╡
│ 1          │
└────────────┘
```

The test case ensuring we can actually insert 3 columns and properly get 3 columns back.

scsmithr: I'm sure we have other tests that would catch this, but there's currently only two files running in CI for the rpc stuff, so just wanted to add it in to catch regressions before we have all the SLTs working for rpc.

vrongmeal: Gotcha! Yeah, this makes sense.

vrongmeal (approved):

PR #1670, opened 2023-08-30
feat: Remote to local copy to

### LINKED ISSUE #1600: Copy To physical plan

vrongmeal:
We rewrite any `CopyTo` plan (in the remote planner) that has a copy to
local destination and replace it with the source query. The rewriter
collects the format and destination of the copy to plan and we wrap the
`SendRecvJoinExec` with a `CopyToExec` for local.

Fixes #1600

### DIFF
--- a/testdata/sqllogictests/functions/parquet_scan.slt
+++ b/testdata/sqllogictests/functions/parquet_scan.slt
@@ -22,12 +22,15 @@ select count(*) from parquet_scan('https://github.com/GlareDB/glaredb/raw/main/t
 #
 # Note that this is a pretty big file, but the limit will be pushed down to the
 # exec, ensuring we don't need to load the whole thing.
-query T
-select length(head) > 1 from parquet_scan(
-  'https://huggingface.co/datasets/allenai/soda/resolve/refs%2Fconvert%2Fparquet/default/train/0000.parquet'
-) limit 1;
-----
-t
+#
+# TODO: Re-enable this test with a new URL that works. This URL returns 404.
+#
+# query T
+# select length(head) > 1 from parquet_scan(
+#   'https://huggingface.co/datasets/allenai/soda/resolve/refs%2Fconvert%2Fparquet/default/train/0000.parquet'
+# ) limit 1;
+# ----
+# t
 
 # Multiple URLs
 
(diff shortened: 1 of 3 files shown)


vrongmeal:
This doesn't work since the logical plan and physical plan both have different schemas which causes the planner to error...
Need to figure that part out.

crates/sqlexec/src/planner/physical_plan/copy_to.rs
```diff
@@ -37,7 +93,11 @@ impl ExecutionPlan for CopyToExec {
     }
 
     fn schema(&self) -> Arc<Schema> {
-        GENERIC_OPERATION_AND_COUNT_PHYSICAL_SCHEMA.clone()
+        if self.is_dummy_query() {
+            self.source.schema()
+        } else {
+            GENERIC_OPERATION_AND_COUNT_PHYSICAL_SCHEMA.clone()
+        }
```
scsmithr, line 100: I believe this should always be the `GENERIC_...` schema as that's what we use to determine that the operation was a COPY.

vrongmeal: Yeah! I'll experiment a bit here. I want to wrap the source query's `SendRecvJoinExec` inside a `CopyToLocal` exec. Because we're using extensions, DF checks that the logical and physical plans have the same schema. We eventually get the same schema, but the DDL planner does not handle it. I'll continue working here once I'm done with values exec.

vrongmeal:
[UPDATE] PR ready for review:

We rewrite any `CopyTo` plan (in the remote planner) that has a copy to
local destination and replace it with the source query. The rewriter
collects the format and destination of the copy to plan and we wrap the
`SendRecvJoinExec` with a `CopyToExec` for local.

vrongmeal:
> I think the rewrite makes sense for now to get this working, but like with the local table stuff, I think it's likely we'll need to move this into physical optimization.

Yup! For now, I just wanted to get this working.

testdata/sqllogictests/functions/parquet_scan.slt
```diff
@@ -22,12 +22,15 @@ select count(*) from parquet_scan('https://github.com/GlareDB/glaredb/raw/main/t
...
-t
+#
+# TODO: Re-enable this test with a new URL that works. This URL returns 404.
+#
+# query T
+# select length(head) > 1 from parquet_scan(
+#   'https://huggingface.co/datasets/allenai/soda/resolve/refs%2Fconvert%2Fparquet/default/train/0000.parquet'
+# ) limit 1;
+# ----
+# t
```
vrongmeal, line 33: Disabled the test for now.

vrongmeal: Opened https://github.com/GlareDB/glaredb/issues/1682

Issue #1665, opened 2023-08-29
RPC: Values exec not serialized


scsmithr:
## Description

Serialization just isn't implemented for it.

Means we can't run queries like `insert into my_table values (1)`.

PR #1673, opened 2023-08-30
fix: Serialize `ValuesExec` for RPC.

vrongmeal:
```
GlareDB (v0.4.0)
Type \help for help.
Using in-memory catalog
>
> values (1), (2), (3);
┌─────────┐
│ column1 │
│ ──      │
│ Int64   │
╞═════════╡
│ 1       │
│ 2       │
│ 3       │
└─────────┘
>
> create table abc (a int, b text);
Table created
>
> insert into abc
::: values (1, 'something'), (2, 'another thing');
Inserted 2 rows
>
> select * from abc;
┌───────┬───────────────┐
│ a     │ b             │
│ ──    │ ──            │
│ Int32 │ Utf8          │
╞═══════╪═══════════════╡
│ 1     │ something     │
│ 2     │ another thing │
└───────┴───────────────┘
```

Fixes: #1665

### DIFF
--- a/testdata/sqllogictests/rpc.slt
+++ b/testdata/sqllogictests/rpc.slt
@@ -89,3 +89,27 @@ query III
 select * from t1661;
 ----
 1 2 3
+
+# Test if "values" clause works.
+# https://github.com/GlareDB/glaredb/pull/1673
+
+query I
+values (1), (2), (3);
+----
+1
+2
+3
+
+statement ok
+create table t1665 (a int, b text);
+
+statement ok
+insert into t1665 values
+  (1, 'a-thing'),
+  (2, 'b-thing');
+
+query TI
+select b, a from t1665;
+----
+a-thing  1
+b-thing  2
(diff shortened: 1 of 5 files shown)


scsmithr (approved):
Very exciting!

Would you be able to add a small test case to rpc.slt (since we're running that in ci with rpc)? Then we can work on getting the other SLTs running!

vrongmeal:
> Would you be able to add a small test case to rpc.slt (since we're running that in ci with rpc)? Then we can work on getting the other SLTs running!

Yup! I'll add the same case as in the PR description.

testdata/sqllogictests/rpc.slt
```diff
@@ -89,3 +89,27 @@ query III
...
+statement ok
+insert into t1665 values
+  (1, 'a-thing'),
+  (2, 'b-thing');
+
+query TI
+select b, a from t1665;
+----
+a-thing  1
+b-thing  2
```
vrongmeal, line 115: @scsmithr Added a couple of basic test cases.

Issue #1587, opened 2023-08-19
Stack overflow when querying view that contains a view (on mac)


scsmithr:
## Description

Specifically this test case:

```
# View/CTE name conflict (#968)
#
# `external_datasources` makes use of a CTE named 'datasources'. Creating a view
# of the same name that referenced it previously resulted in a stack overflow.

statement ok
create view datasources as select * from glare_catalog.external_datasources;

statement ok
select * from datasources;
```

PR #1681, opened 2023-08-30
feat: remote execution of table funcs via dispatcher

universalmind303:
So this moves all of the tablefunc exec logic into the dispatcher (similar to how we are doing with tables). 

The functions are a bit more dynamic than the tables and as such, we sometimes need to determine the execution preference at runtime. 

**Ideally I wanted to do this during physical planning**. 

However, the logical planner needs the schema, so we needed to move the logic further upstream and determine if the function should run remote or locally before we get to the LogicalPlan. Otherwise we end up performing schema scans locally, and table scans remotely which causes the state to get out of sync.

--- 
Implicitly we know that: 
- some functions will always to run remotely when possible (s3, postgres, bigquery,...)
- some always locally (local filesystem)
- some determined at runtime
---

## So in order to determine the logic prior to planning, we had to do 2 things. 

- First we add some metadata to the function entries specifying their "runtime_preference" (Local, Remote, Unspecified)

- Next, we look at that metadata during dispatch.
  - we can easily rule out (Local & Remote preferences)
  - `Unspecified` requires us to inspect the args to determine the true destination at runtime.  

If at this point, we still don't know where to execute it, it should error as it would indicate a bug in our resolution logic. 

We could probably expand this metadata hint to other catalog entries as well.

### DIFF
--- a/crates/protogen/proto/rpcsrv/service.proto
+++ b/crates/protogen/proto/rpcsrv/service.proto
@@ -94,6 +94,10 @@ message FetchCatalogResponse { metastore.catalog.CatalogState catalog = 1; }
 message DispatchAccessRequest {
   bytes session_id = 1;
   ResolvedTableReference table_ref = 2;
+  repeated bytes args = 3;
+  // gRPC doesn't really provide a way for Option<Vec<T>> or Option<Map<K,V>>
+  // so we use a map and just check if it's empty or not in rust.
+  map < string, bytes > options = 4;
 }
 
 // Execute a physical plan and get the stream.
--- a/crates/datafusion_ext/src/functions.rs
+++ b/crates/datafusion_ext/src/functions.rs
@@ -2,23 +2,27 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::sync::Arc;
 
-use async_trait::async_trait;
-use datafusion::common::OwnedTableReference;
-use datafusion::datasource::DefaultTableSource;
-use datafusion::datasource::TableProvider;
-use datafusion::execution::context::SessionState;
-use datafusion::logical_expr::{LogicalPlan, LogicalPlanBuilder};
-use decimal::Decimal128;
-use protogen::metastore::types::catalog::{CredentialsEntry, DatabaseEntry};
-
 use crate::errors::{ExtensionError, Result};
 use crate::vars::SessionVars;
+use async_trait::async_trait;
+use datafusion::datasource::TableProvider;
+use datafusion::execution::context::SessionState;
+use decimal::Decimal128;
+use protogen::metastore::types::catalog::{CredentialsEntry, DatabaseEntry, RuntimePreference};
+use protogen::rpcsrv::types::func_param_value::{
+    FuncParamValue as ProtoFuncParamValue, FuncParamValueArrayVariant,
+    FuncParamValueEnum as ProtoFuncParamValueEnum,
+};
 
 #[async_trait]
 pub trait TableFunc: Sync + Send {
     /// The name for this table function. This name will be used when looking up
     /// function implementations.
     fn name(&self) -> &str;
+    fn runtime_preference(&self) -> RuntimePreference;
+    fn detect_runtime(&self, _args: &[FuncParamValue]) -> Result<RuntimePreference> {
+        Ok(self.runtime_preference())
+    }
 
     /// Return a table provider using the provided args.
     async fn create_provider(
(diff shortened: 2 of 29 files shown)


universalmind303:
looks like generate_series is still having some issues when connected to remote
```
> select * from generate_series(1, 4);
Error: Execution error: failed to execute physical plan on remote service: Internal error: Unsupported plan and extension codec failed with [This feature is not implemented: encoding not implemented for physical plan: LazyMemTableExec { .. }]. Plan: LazyMemTableExec { .. }. This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker
```

universalmind303:
looks like some other SLT issues with `views`

```
>  thread 'tokio-runtime-worker' has overflowed its stack
fatal runtime error: stack overflow
```
https://github.com/GlareDB/glaredb/actions/runs/6030332033/job/16361899055?pr=1681

crates/protogen/proto/rpcsrv/service.proto
```diff
@@ -94,6 +94,10 @@ message FetchCatalogResponse { metastore.catalog.CatalogState catalog = 1; }
 message DispatchAccessRequest {
   bytes session_id = 1;
   ResolvedTableReference table_ref = 2;
+  repeated bytes args = 3;
+  // gRPC doesn't really provide a way for Option<Vec<T>> or Option<Map<K,V>>
+  // so we use a map and just check if it's empty or not in rust.
+  map < string, bytes > options = 4;
```
scsmithr, line 100: I would just mention these are protobuf encoded.

scsmithr (commented):
Lots going on here. I'm good if we merge this in, get it deployed in a qa, and see what works and what breaks.

Currently looking at the stack overflow. Trying to figure out if that's from this change, and if we should just bump the stack size (hopefully temporarily).

scsmithr:
Failing on this test case:

```
# View/CTE name conflict (#968)
#
# `external_datasources` makes use of a CTE named 'datasources'. Creating a view
# of the same name that referenced it previously resulted in a stack overflow.

statement ok
create view datasources as select * from glare_catalog.external_datasources;

statement ok
select * from datasources;
```

The changes you made don't impact the correctness of this. I would just bump the stack size to 4MB here: https://github.com/glaredb/glaredb/blob/4e33b6bad0d1787c65ec1de87bdeb774d8a51421/crates/testing/src/slt/cli.rs#L109. That would only bump it for the slts.

I did some digging, and the above test case will cause the stack to grow beyond 2MB when running through pgsrv, and I could reproduce by spinning up the server, connecting via psql, and running those queries.

However, when running in release mode, the stack size never grows beyond 300KB, and so we shouldn't need to worry about this impacting real usage, at least for now (there's likely some optimization to shrink stack usage during planning).

universalmind303:
> However, when running in release mode, the stack size never grows beyond 300KB, and so we shouldn't need to worry about this impacting real usage, at least for now (there's likely some optimization to shrink stack usage during planning).


This very well could be caused by datafusion's LogicalPlan implementation. Those views likely cause a lot of plans to be created under the hood. I believe one of the driving forces for polars to use arena allocation for the logical plan was to prevent this exact issue. 

> The changes you made don't impact the correctness of this. I would just bump the stack size to 4MB here: https://github.com/glaredb/glaredb/blob/4e33b6bad0d1787c65ec1de87bdeb774d8a51421/crates/testing/src/slt/cli.rs#L109. That would only bump it for the slts.

Wonder if we could just remove the debug symbols when running compiling & running tests in CI.

scsmithr:
> > However, when running in release mode, the stack size never grows beyond 300KB, and so we shouldn't need to worry about this impacting real usage, at least for now (there's likely some optimization to shrink stack usage during planning).
> 
> This very well could be caused by datafusion's LogicalPlan implementation. Those views likely cause a lot of plans to be created under the hood. I believe one of the driving forces for polars to use arena allocation for the logical plan was to prevent this exact issue.

I know there's been some issues previously on datafusion about stack overflow due to the logical plans. I've mostly seen it with the protobuf serialization/deserialization tests, but this matches what I observed above.

> > The changes you made don't impact the correctness of this. I would just bump the stack size to 4MB here: https://github.com/glaredb/glaredb/blob/4e33b6bad0d1787c65ec1de87bdeb774d8a51421/crates/testing/src/slt/cli.rs#L109. That would only bump it for the slts.
> 
> Wonder if we could just remove the debug symbols when compiling & running tests in CI.

I would want those changes to apply locally as well. I've run into the stack overflow before (https://github.com/GlareDB/glaredb/issues/1587) and frequently run the SLTs to make sure I know what I'm breaking.

I'm also not sure if disabling debug info is enough here. I created this profile:

```
[profile.test-limit-debug]
inherits = "test"
debug = 0
```

And run with that during slts and still get an overflow.

scsmithr (approved):

PR #1683, opened 2023-08-31
fix: Return only "NULL" when value is null in RPC SLT

### LINKED ISSUE #1679: RPC: Incorrect null formatting in response
## Description

```
  2023-08-30T19:15:49.372963Z ERROR testing::slt::cli: Error while running test `sqllogictests/simple`, error: test fail: query result mismatch:
[SQL] select null;
[Diff] (-expected|+actual)
-   NULL
+   NULL (empty)
at /Users/sean/Code/github.com/glaredb/glaredb/testdata/sqllogictests/simple.slt:8
```

vrongmeal:
Fixes #1679

### DIFF
--- a/crates/testing/src/slt/test.rs
+++ b/crates/testing/src/slt/test.rs
@@ -319,26 +319,31 @@ impl AsyncDB for TestClient {
                                 if num_columns == 0 {
                                     num_columns = batch.num_columns();
                                 }
+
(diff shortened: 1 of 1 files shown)


vrongmeal:
Sean, just tagged you for notifying. Merging this in. Minor change in RPC test to fix null formatting.

`sqllogictests/simple` can't still be run because of lack of insert into temp table over RPC.

PR #1692, opened 2023-08-31
fix: DDL for temp tables

### LINKED ISSUE #1676: RPC: Unable to insert into temp table
## Description

Working through some slts.

```
> \open http://localhost:6542
> CREATE TEMP TABLE t1 (a INTEGER, b INTEGER, c INTEGER);
Table created
> INSERT INTO t1 VALUES (1,2,3);
Error: Error during planning: Unable to fetch table provider for 't1': failed to resolve: failed to find table: t1
```

vrongmeal:
Create and Insert.

TODO: Delete temp table.

```
> create temp table abc (a int);
Table created
> select * from abc;
┌───────┐
│ a     │
│ ──    │
│ Int32 │
╞═══════╡
└───────┘
> insert into abc values (1), (2);
Inserted 2 rows
> select * from abc;
┌───────┐
│ a     │
│ ──    │
│ Int32 │
╞═══════╡
│ 1     │
│ 2     │
└───────┘
> create temp table def (a int, b text) as values (1, 'vaibhav');
Table created
> select * from def;
┌───────┬─────────┐
│ a     │ b       │
│ ──    │ ──      │
│ Int32 │ Utf8    │
╞═══════╪═════════╡
│ 1     │ vaibhav │
└───────┴─────────┘
```

Fixes #1676


vrongmeal:
Added `sqllogictests/simple` to RPC as well. Works locally.

crates/sqlexec/src/remote/rewriter.rs
```diff
@@ -118,15 +153,40 @@ impl TreeNodeRewriter for CopyRemoteToLocalRewriter {
...
+                        produce_one_row: false,
+                        schema: create_temp_table.schema.clone(),
+                    })))
+            } else if let Some(_drop_tables) = ext.node.as_any().downcast_ref::<DropTables>() {
+                // TODO: Figure out how to seperate temp tables and others.
+                // Send the other tables to remote and keep the remote tables
+                // locally. Create a "combined" plan where both remote and local
+                // tables are deleted (if required).
+                Ok(LogicalPlan::Extension(ext))
+            } else if let Some(copy_to) = ext.node.as_any().downcast_ref::<CopyTo>() {
```
universalmind303, line 183: we should use `ExtensionType` here instead of inline downcasting. It'll make it a bit more readable as well as properly handling all discriminants. 


```rust 
if let LogicalPlan::Extension(ext) = node {
  let ext_type = ext.node.name().parse::<ExtensionType>().unwrap();
  match ext_type {
    ExtensionType::Insert => {
      let insert_into = Insert::try_downcast_extension(ext)?;
      // ..

vrongmeal: Yeah, there's a TODO at the top for this. I'll make this change as well.

scsmithr (approved):

PR #1693, opened 2023-08-31
fix: rpc create table subqueries

### LINKED ISSUE #1677: RPC: Unable to CREATE TABLE AS when source contains unserializable table source
## Description

```
> create table t1 as select * from generate_series(1, 15, 2);
Error: Execution error: failed to execute physical plan on remote service: Internal error: Unsupported plan and extension codec failed with [This feature is not implemented: encoding not implemented for physical plan: LazyMemTableExec { .. }]. Plan: LazyMemTableExec { .. }. This was likely caused by a bug in DataFusion's code and we would welcome that you file an bug report in our issue tracker
```

universalmind303:
closes #1677

### DIFF
--- a/crates/datafusion_ext/src/functions.rs
+++ b/crates/datafusion_ext/src/functions.rs
@@ -20,7 +20,11 @@ pub trait TableFunc: Sync + Send {
     /// function implementations.
     fn name(&self) -> &str;
     fn runtime_preference(&self) -> RuntimePreference;
-    fn detect_runtime(&self, _args: &[FuncParamValue]) -> Result<RuntimePreference> {
+    fn detect_runtime(
+        &self,
+        _args: &[FuncParamValue],
+        _parent: &RuntimePreference,
+    ) -> Result<RuntimePreference> {
         Ok(self.runtime_preference())
     }
 
--- a/crates/datafusion_ext/src/functions.rs
+++ b/crates/datafusion_ext/src/functions.rs
@@ -23,7 +23,7 @@ pub trait TableFunc: Sync + Send {
     fn detect_runtime(
         &self,
         _args: &[FuncParamValue],
-        _parent: &RuntimePreference,
+        _parent: RuntimePreference,
     ) -> Result<RuntimePreference> {
         Ok(self.runtime_preference())
     }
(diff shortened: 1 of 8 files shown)


crates/protogen/src/metastore/types/catalog.rs
```diff
@@ -494,47 +494,38 @@ impl From<FunctionType> for catalog::function_entry::FunctionType {
         }
     }
 }
+
```
universalmind303, line 497: i was getting some inconsistencies when converting this at runtime, so i modified it to be a bit more explicitly linked to the proto definition.

crates/protogen/src/metastore/types/catalog.rs
```diff
@@ -494,47 +494,38 @@ impl From<FunctionType> for catalog::function_entry::FunctionType {
...
-                RuntimePreference::Unspecified
-            }
-            catalog::function_entry::RuntimePreference::Local => RuntimePreference::Local,
-            catalog::function_entry::RuntimePreference::Remote => RuntimePreference::Remote,
-        })
+impl From<catalog::function_entry::RuntimePreference> for RuntimePreference {
+    fn from(value: catalog::function_entry::RuntimePreference) -> Self {
+        // This is safe because the enum values should be identical.
+        // If they're not, then we need to update the proto repr
+        unsafe { std::mem::transmute(value) }
```
scsmithr, line 522: I'd prefer sticking the the match to be more explicit since this is a write once thing. Also the match would actually compile down to nothing.

crates/datafusion_ext/src/functions.rs
```diff
@@ -20,7 +20,11 @@ pub trait TableFunc: Sync + Send {
     /// function implementations.
     fn name(&self) -> &str;
     fn runtime_preference(&self) -> RuntimePreference;
-    fn detect_runtime(&self, _args: &[FuncParamValue]) -> Result<RuntimePreference> {
+    fn detect_runtime(
+        &self,
+        _args: &[FuncParamValue],
+        _parent: &RuntimePreference,
```
scsmithr, line 26:
```suggestion
        _parent: RuntimePreference,
```

scsmithr (approved):

crates/datafusion_ext/src/functions.rs
universalmind303: why would we want an owned version here?

scsmithr: RuntimePreference is Copy (i32), so really comes down to passing a reference to an i32, or passing i32 itself. I normally just pass values when something is Copy.

PR #1715, opened 2023-09-05
feat: add further optimizations for hybrid unions

universalmind303:
groups related children in unions 

for example: takes a query 

```sql
select * from local where c = 1
union (select * from local where c = 2)
union (select * from remote where c = 3)
```

and rewrites the initial plan of 

```mermaid
graph TD

UnionExec --> FilterExec1 & FilterExec2 & FilterExec3
  FilterExec1 --> RemoteExec1
  FilterExec2 --> LocalExec1
  FilterExec3 --> LocalExec2
subgraph Remote
  RemoteExec1 --> EmptyExec1["EmptyExec (test_schema)"]
end

subgraph Local1[Local]
  LocalExec1 --> EmptyExec2["EmptyExec (test_schema)"]
end

subgraph Local2[Local]
  LocalExec2 --> EmptyExec3["EmptyExec (test_schema)"]
end
```

to 
```mermaid
graph TD
  UnionExec --> RemoteExec & LocalExec
  subgraph Local
  LocalExec -->
  UnionExecLocal[UnionExec] --> LocalExec1 & LocalExec2
  LocalExec1 --> EmptyExec2["EmptyExec (test_schema)"]
  LocalExec2 --> EmptyExec3["EmptyExec (test_schema)"]
end 
subgraph Remote 
  RemoteExec --> FilterExec --> EmptyExec1["EmptyExec (test_schema)"]
end
```

### DIFF
--- a/crates/datafusion_ext/src/runtime/group_pull_up.rs
+++ b/crates/datafusion_ext/src/runtime/group_pull_up.rs
@@ -82,7 +82,6 @@ impl PhysicalOptimizerRule for RuntimeGroupPullUp {
     ) -> Result<Arc<dyn ExecutionPlan>> {
         plan.transform_up(&|plan| {
             if !can_pull_through_node(plan.as_ref()) {
-                println!("can't pull through node");
                 return Ok(Transformed::No(plan));
             }
 
--- a/crates/datafusion_ext/src/runtime/group_pull_up.rs
+++ b/crates/datafusion_ext/src/runtime/group_pull_up.rs
@@ -1,6 +1,6 @@
 use datafusion::common::tree_node::{Transformed, TreeNode};
 use datafusion::config::ConfigOptions;
-use datafusion::error::Result;
+use datafusion::error::{DataFusionError, Result};
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
 use datafusion::physical_plan::aggregates::AggregateExec;
 use datafusion::physical_plan::filter::FilterExec;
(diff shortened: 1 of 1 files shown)


crates/datafusion_ext/src/runtime/group_pull_up.rs
```diff
@@ -23,6 +24,54 @@ impl RuntimeGroupPullUp {
...
+        for child in it {
+            match child.preference {
+                RuntimePreference::Remote => remote.push(child),
+                RuntimePreference::Local => local.push(child),
+                RuntimePreference::Unspecified => todo!("unspecified runtime preference"),
+            }
+        }
+
+        // if the plan only contains one runtime, we can just return it as is.
+        if (local.len() | remote.len()) == 0 {
```
scsmithr, line 62: I'd rewrite this to `local.is_empty() || remote.is_empty()` for clarity.

crates/datafusion_ext/src/runtime/group_pull_up.rs
```diff
@@ -23,6 +24,54 @@ impl RuntimeGroupPullUp {
...
+
+        let it = children.into_iter();
+        let mut remote = Vec::new();
+        let mut local = Vec::new();
+
+        for child in it {
+            match child.preference {
+                RuntimePreference::Remote => remote.push(child),
+                RuntimePreference::Local => local.push(child),
+                RuntimePreference::Unspecified => todo!("unspecified runtime preference"),
```
scsmithr, line 57: I think we should return the untransformed plan in this case instead of panicking. Views are unspecified, so I believe a union between a view and something else would trigger this.

I left views as "unspecified" since the grouping for a view depends on its content (which itself will have the appropriate runtime preference).

crates/datafusion_ext/src/runtime/group_pull_up.rs
universalmind303: so the logic for this was actually incorrect, the previous tests actually caught it! should be all fixed now!

Issue #1655, opened 2023-08-28
better startup message when connecting to non cloud rpc 


universalmind303:
## Description

```
> just run local --ignore-rpc-auth -c http://0.0.0.0:6789   
Connected to Cloud deployment: unknown
```

instead of `Connected to Cloud deployment: unknown`, it should say something more meaningful like `Connected to Remote Deployment: http://0.0.0.0:6789`

PR #1717, opened 2023-09-05
feat: add better startup messages to cli

universalmind303:
Mostly addresses #1655, but also did some refactoring. 

I kept finding myself getting lost a bit when navigating `glaredb` crate, so this should help separate the actual logic from the structs & datatypes. 

WRT the refactor, it mostly moves the cli subcommand fields out of the enum variants & into their own structs. 

### Before
image

### After
image

### Before
image

### After
image


crates/glaredb/src/local.rs
```diff
@@ -158,22 +75,29 @@ impl LocalSession {
         .await?;
 
         let sess = if let Some(url) = opts.cloud_url.clone() {
-            let exec_client = if opts.ignore_rpc_auth {
-                RemoteClient::connect(url).await?
+            let (exec_client, info_msg) = if opts.ignore_rpc_auth {
+                let u = &url.to_string();
+                (
+                    RemoteClient::connect(url).await?,
+                    format!("Connected to remote GlareDB server: {}", u.bright_cyan()),
```
scsmithr, line 82: Very nitpicky, bright cyan is hard to read on light themes (blasphemy I know). Could we make this and the one below just cyan?

scsmithr (approved):
lgtm, much cleaner

PR #1746, opened 2023-09-11
feat: tls for rpc-srv

adhish20:


adhish20:
To test:
1. generate the key files by running ./scripts/gen-certs.sh
2. Run server : just run rpc-proxy --rpc-server-cert scripts/client.crt --rpc-server-key scripts/client.key --cloud-api-addr https://qa.glaredb.com --cloud-auth-code test
3. just run local -c glaredb://test:adhish@org.0.0.0.0:6444/abc --domain glaredb.com --ca-cert-path scripts/server.crt

adhish20:
the cert names in the testing are weird because I didn't wanted to break pgsrv tests

crates/glaredb/src/args/local.rs
```diff
@@ -57,6 +57,14 @@ pub struct LocalClientOpts {
     /// Max number of rows to display.
     #[arg(long)]
     pub max_rows: Option<usize>,
+
+    /// Path to CA certificate for TLS (required for TLS protocol).
+    #[arg(long)]
+    pub ca_cert_path: Option<String>,
```
tychoish, line 63: this is for the client (I am, admittedly new so "client for what?" might be a thing...  shouldn't/wouldn't things fall back to use system certificates, and using this for specific certificates? (presumably in non-testing environments you'd use letsencrypt or something similar.)

In general having the commands/operations have "tls by default, have a flag to disable it" is a good idea.

tychoish (commented):
I'm just a guy reading through things, and I *definitely* lack all context, so take all of these comments with a very large grain of salt. 

> the cert names in the testing are weird because I didn't wanted to break pgsrv tests

do they need to be the same certificate? 

speaking of tests? 😂

crates/glaredb/src/args/local.rs
```diff
@@ -57,6 +57,14 @@ pub struct LocalClientOpts {
...
+
+    /// Domain name against which to verify the server’s TLS certificate (required for TLS protocol).
+    #[arg(long)]
+    pub domain: Option<String>,
```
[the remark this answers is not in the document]
adhish20: this is done by tonic by default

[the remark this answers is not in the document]
adhish20: this is only client side verification (which is required by tonic). We are only doing this to encrypt the connection.

crates/glaredb/src/args/local.rs
adhish20: updated the help line to mention that this is for rpcsrv proxy TLS

greyscaled:
@tychoish some additional context here: https://github.com/GlareDB/cloud/issues/2164

adhish20:
> I'm just a guy reading through things, and I _definitely_ lack all context, so take all of these comments with a very large grain of salt.
> 
> > the cert names in the testing are weird because I didn't wanted to break pgsrv tests
> 
> do they need to be the same certificate?
> 
> speaking of tests? 😂

we can have separate cert generators for them for the actual tests, these commands are to test that tis is working fine locally

scripts/gen-certs.sh
```diff
@@ -10,3 +10,7 @@ openssl req -new -x509 -days 365 -nodes -text \
         -out server.crt \
         -keyout server.key \
         -subj "/CN=glaredb.com"
+
+openssl req -noenc -newkey rsa -keyout client.key -out client.csr -subj '/CN=glaredb.com' -addext subjectAltName=DNS:glaredb.com
+
+openssl x509 -req -in client.csr -CA server.crt -CAkey server.key -days 365 -out client.crt -copy_extensions copy
```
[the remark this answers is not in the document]
greyscaled: **Drive by here**:

Would it make sense to gen these in memory in the tests using weak params with the available interface https://docs.rs/openssl/latest/openssl/pkey/index.html and https://docs.rs/openssl/latest/openssl/x509/index.html ?

For example:

```
openssl genpkey  -algorithm RSA -pkeyopt rsa_keygen_bits:512
```

crates/glaredb/src/args/mod.rs
```diff
@@ -49,6 +49,14 @@ pub struct RpcProxyArgs {
...
     #[clap(long)]
     pub cloud_auth_code: String,
+
+    /// Path to server cert to use for TLS (required for TLS protocol).
+    #[clap(long)]
+    pub rpc_server_cert: Option<String>,
+
+    /// Path to server key to use for TLS (required for TLS protocol).
+    #[clap(long)]
+    pub rpc_server_key: Option<String>,
```
universalmind303, line 59: could we have aliases for these, 
```sh
./glaredb proxy \
  --rpc-server-cert="path/to/cert" \
  --rpc-server-key="path/to/key"
 ``` 
 
 seems pretty long to me. 
 
 `--cert="cert"` and `--key="key"` seems pretty straightforward to me.

File: crates/glaredb/src/args/mod.rs [hunk 5]
```diff
@@ -49,6 +49,14 @@ pub struct RpcProxyArgs {
     /// Authorization code for communicating with Cloud.
     #[clap(long)]
     pub cloud_auth_code: String,
+
+    /// Path to server cert to use for TLS (required for TLS protocol).
+    #[clap(long)]
+    pub rpc_server_cert: Option<String>,
```
universalmind303, line 55: If this is required to be a path, we should use `PathBuf` instead of `String`. 

To me, string implies that they could inline the cert `--rpc-server-cert=$MY_CERT` _(which we could add in a later PR if someone requests the functionality)_

I'm pretty sure clap will parse PathBuf's by default.

tychoish:
> we can have separate cert generators for them for the actual tests, these commands are to test that tis is working fine locally

This makes sense, I was more saying that we should have some kind of CI-able test of this rather than relying on local/manual testing. (sorry for the shorthand) 

I realize that this is (mostly) for ourselves, and likely to get used mostly in scripts and config files, [which is why i think I tend toward the longer flag names are a bit better, though I don't think its particularly material], but for service-type things, config files seem a bit easier to manage anyway, (I also feel like its easier to add a new flag than it is to remove an old one, so I'd tend to avoid aliases but...

scripts/gen-certs.sh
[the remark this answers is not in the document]
greyscaled: I think we can just remove this after the first pass, as it helped to get things going, but isn't apart of any broader test.

Furthermore `gen-certs` gets called as part of building our image:

https://github.com/GlareDB/glaredb/blob/9e0e4b3dd8b196d151c5515cb58fe91273fa4f91/.github/workflows/image.yaml#L43-L44

https://github.com/GlareDB/glaredb/blob/9e0e4b3dd8b196d151c5515cb58fe91273fa4f91/Dockerfile#L12-L13

and I think having cert files in `scripts/` is confusing (even if gitignored)

crates/glaredb/src/rpc_proxy.rs
```diff
@@ -28,11 +33,33 @@ impl RpcProxy {
...
-            .add_service(ExecutionServiceServer::new(self.handler))
-            .serve(addr)
-            .await?;
+        let mut server = Server::builder().trace_fn(|_| debug_span!("rpc_proxy_service_request"));
+
+        match (rpc_server_cert, rpc_server_key) {
+            (Some(cert), Some(key)) => {
+                let cert = std::fs::read_to_string(cert)?;
+                let key = std::fs::read_to_string(key)?;
+                let identity = Identity::from_pem(cert, key);
```
tychoish, line 42: looks like here's the place where we're pulling information (like domain?) out of the server's certificate, I'm inclined to think that the client certs would work the same way.

adhish20:
@greyscaled @tychoish 
Added a config file path that defaults to "/etc/glaredb.conf" which is supposed to be TOML file.
Also added a param to disable tls on the server side. If this looks good, I will do the same for client.
```
[rpc_tls]
server_cert_path = '/Users/adhish/GlareDB/glaredb/server.pem'
server_key_path = '/Users/adhish/GlareDB/glaredb/server.key'
```

crates/glaredb/src/rpc_proxy.rs
adhish20: Identity just stores the cert and the key. Doesn't provide any other information.

crates/glaredb/src/args/local.rs
```diff
@@ -58,13 +58,13 @@ pub struct LocalClientOpts {
...
 
-    /// Path to CA certificate for rpcsrv proxy TLS (required for TLS protocol).
-    #[arg(long)]
-    pub ca_cert_path: Option<String>,
-
     /// Domain name against which to verify the rpcsrv proxy server’s TLS certificate (required for TLS protocol).
-    #[arg(long)]
-    pub domain: Option<String>,
+    #[arg(long, value_parser, default_value = "/etc/glaredb.conf", hide = true)]
+    pub tls_conf_path: PathBuf,
```
tychoish, line 63: I'd call this `conf_path`

crates/glaredb/src/rpc_proxy.rs
tychoish: https://docs.rs/x509-parser/latest/x509_parser/certificate/struct.X509Certificate.html <- looks like this will pull the info out of the certificate.

tychoish (commented):
in my ideal world, we'd be able to specify default values for this that are reasonable, and then fall back _either_ to a config file or command-line arguments.

ok, having sort of caused an interlude, I think the config file approach is something we should not do as a rush job, and consider what we want more broadly rather than complicate our interface surface on a ticket here, if that makes sense. How does the following sound:

- we do keep the `disable_tls` option (and hide it by default.)
- we add a (hidden) flag for (only) the path to the public certificate (for the client).
- we add a flag to the server (probably for just the key; also hidden) to the server side.
- using the x509 library we can pull the domain out of the parsed certificate. (if this seems reasonable?) 

I think to make this easier to work out, also: 
- if @greyscaled can get you a certificate (that we won't be using, just for testing, so we can pull the cert off a test pod and then rotate) to test against, 
- you can check using the "real" cert (I think it's reasonable to add a shell script and commit unused certs in a `testdata` folder (or similar) that just runs glaredb client/servers to see if it works by eyeball at least right now.)

(everything else about this seems fine to me 😂) 
- 
-

crates/rpcsrv/src/proxy.rs
```diff
@@ -76,11 +77,26 @@ impl<A: ProxyAuthenticator> RpcProxyHandler<A> {
...
+            let url = format!("https://{}:{}", key.ip, key.port);
+            let ca = std::fs::read_to_string(
+                meta.get("ca_cert_path")
+                    .map(|s| s.to_str())
+                    .transpose()?
+                    .unwrap(),
+            )
+            .unwrap();
+            let tls_conf = ClientTlsConfig::new()
+                .ca_certificate(Certificate::from_pem(ca))
```
tychoish, line 91: does this mean we're reading the certificate from the file each time a new request is made?

this feels wrong to me (because we'd just read the same thing again and again) but the truth is that cert-manager can replace/rotate the certificate on a running pod, so maybe this is the correct. Maybe there should be a small ticket or a comment to say "cache this in the service struct, and reread occassionally" (the other model is to use a file system watcher to get notified of the events, but that's worse) 

anyway, I'm just thinking outloud; this seems reasonable and I don't think we should change it now, but a comment might be good.

PR #1755, opened 2023-09-12
wip: serializable udfs

universalmind303:
This copies a lot of code from upstream datafusion's `DefaultPlanner` as we've reached the point that we need a fully custom physical planner & can no longer rely on rewriter & optimizer rules. 

The planner is mostly copied from `DefaultPlanner`, with the stuff from `RemotePlanner` added in.


crates/sqlexec/src/functions.rs
```diff
@@ -1,206 +1,205 @@
...
+}
 
-    fn try_from_name_implicit_pg_catalog(name: &str) -> Option<BuiltinScalarFunction> {
-        match name {
-            "current_schemas" => Some(BuiltinScalarFunction::CurrentSchemas),
-            _ => None,
-        }
+impl BuiltinScalarFunction {
+    pub fn find_function(name: &str) -> Option<Self> {
+        Self::from_str(name).ok()
```
scsmithr, line 115: I believe this is causing a test failure where we can't use the full qualified name (`pg_catalog.current_schemas`). Previously we just split on '.' to get the schema, and I'd be good with that here too.

scsmithr: Oh nvm, these should be in the pg funcs:

```
    /// SQL function `current_user`
    ///
    /// `current_user()` -> `String`
    /// ```sql
    /// select current_user();
    /// ```
    CurrentUser,
    /// SQL function `current_role`
    ///
    /// `current_role()` -> `String`
    /// ```sql
    /// select current_role();
    /// ```
    CurrentRole,
    /// SQL function `user`
    ///
    /// `user()` -> `String`
    /// ```sql
    /// select user();
    /// ```
    User,
    /// SQL function `current_schema`
    ///
    /// `current_schema()` -> `String`
    /// ```sql
    /// select current_schema();
    /// ```
    CurrentSchema,
    /// SQL function `current_database`
    ///
    /// `current_database()` -> `String`
    /// ```sql
    /// select current_database();
    /// ```
    CurrentDatabase,
    /// SQL function `current_catalog`
    ///
    /// `current_catalog()` -> `String`
    /// ```sql
    /// select current_catalog();
    /// ```
    CurrentCatalog,
```

universalmind303: Ok I think i got all of the pg specific ones moved into `BuiltinPostgresFunctions`

scsmithr (approved):
ship it

PR #1756, opened 2023-09-13
fix: serialize AnalyzeExec

### LINKED ISSUE #1752: RPC: AnalyzeExec not serialized
## Description

> Execution error: failed to execute physical plan on remote service: Internal error: Unsupported plan and extension codec failed with [This feature is not implemented: encoding not implemented for physical plan: AnalyzeExec verbose=false

For `explain analyze ...` queries

adhish20:
Fixes #1752

### DIFF
--- a/crates/protogen/src/sqlexec/physical_plan.rs
+++ b/crates/protogen/src/sqlexec/physical_plan.rs
@@ -293,10 +293,18 @@ pub struct InterleaveExec {}
 pub struct RuntimeGroupExec {}
 
 #[derive(Clone, PartialEq, Message)]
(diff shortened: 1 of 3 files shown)


adhish20:
should we submit a change to data fusion to make the verbose param pub in AnalyzeExec? Currently, I hardcoded it to true

scsmithr (commented):
> should we submit a change to data fusion to make the verbose param pub in AnalyzeExec? Currently, I hardcoded it to true

Yes

Can we also add an slt test to ensure that this doesn't error with rpc?

Issue #1648, opened 2023-08-26
Chore: Explicitly set Rust version in CI


scsmithr:
## Description

To avoid it getting changed out from under us and possibly causing failures due to new clippy lints being added.

PR #1763, opened 2023-09-14
ci: pin rust version in CI

tychoish:
Closes: https://github.com/GlareDB/glaredb/issues/1648

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -21,7 +21,8 @@ jobs:
 
       - name: Rust toolchain
         uses: dtolnay/rust-toolchain@1.72.0
-        components: clippy,rustfmt
+        with: 
+          components: clippy,rustfmt
 
       - name: Cache
         uses: Swatinem/rust-cache@v2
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -20,7 +20,8 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
+        uses: dtolnay/rust-toolchain@1.72.0
+        components: clippy,rstfmt
```
scsmithr, line 24: This needs to be:
```
with:
  components: clippy,rustfmt
```

tychoish:
```suggestion
        components: clippy,rustfmt
```

tychoish: The thing that is confusing is how this passed at all.

tychoish: > This actual "Build and Test" workflow isn't running,

ah that's confusing that syntax failures in the yaml file don't result in failed tasks?

thanks for the eyeballs: I would have actually looked if the task had had something that wasn't a check mark next to it?

scsmithr (approved):

PR #1766, opened 2023-09-15
ci: fanout CI steps for parallel build

tychoish:
My CI build ran for 11 minutes. That seems long, let's see what this is like.

### DIFF
--- a/scripts/protocol-test.sh
+++ b/scripts/protocol-test.sh
@@ -11,6 +11,9 @@ export RUST_BACKTRACE=1
 
 run_id=${RANDOM}
 
+# build once...
+just build || true
+
 # Start up GlareDB.
 glaredb_log_file="/tmp/glaredb.log-${run_id}"
 nohup cargo run --bin glaredb -- -v server --user glaredb --password dummy > "${glaredb_log_file}" 2>&1 &
--- a/scripts/protocol-test.sh
+++ b/scripts/protocol-test.sh
@@ -11,9 +11,6 @@ export RUST_BACKTRACE=1
 
 run_id=${RANDOM}
 
-# build once...
-just build || true
-
 # Start up GlareDB.
 glaredb_log_file="/tmp/glaredb.log-${run_id}"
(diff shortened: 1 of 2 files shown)


scripts/protocol-test.sh
```diff
@@ -30,7 +30,7 @@ cargo run --bin pgprototest -- \
     -v || ret=$?
 
 # Kill GlareDB.
-kill "${glaredb_pid}"
+kill "${glaredb_pid}" || true
```
tychoish, line 33: failing to kill the process here (if it say, doesn't start up?) means that we don't end up 

we can imagine better implementations of this test: we kill glaredb after we check if the test has failed and report its output, but to do that in shell would involve a lot of `trap` and signal stuff, which I've (delightfully) not had write in a long time, and seems like overkill. I can't imagine a case where we fail to kill glaredb here and where the test would pass and the system would be operable (we start glare as a different user? we could probably SIGKILL here to rule out problems with the signal handler in glare as being a cause for a failed kill), so this seems both safe and was the thing that helped me debug a failed test.

tychoish:
The biggest win of this is that you get feedback early and easy about what passed and what didn't pass in your build: I think this makes it easier to assess (both as a developer and a potential reviewer) what the state of a branch is: is everything broken or is there just a lint failure?

The first results begin returning in 4-5 minutes, which is much less than the previous 11+ minute runtime, and it's more clear how you might add in more tests: (and unblocking this removes any disintegration for people to add more tests. 

The cons, are that there's a lot of redundancy in the build (still) and the total amount of compute time is more than it would be otherwise. I suspect that there is some missing caching? We can probably counteract the increased cost, caching not withstanding somewhat by using smaller instances for everything except the integration tests. Probably not worth thinking about too much. 

I also don't have the appropriate access to modify the required tasks for the build, which I think is a blocker.

tychoish:
I'm going to push experiments to this branch so we can compare timing using different cache strategies.

PR #1767, opened 2023-09-15
feat: opt-in tls for rpc

greyscaled:
## Summary

Adds `--diable-tls` (true by default) to glaredb local and glaredb rpcsrv.

### Details

This is opt-in while we're still working on the feature and ensuring end-to-end experience
works as expected. At some point in the near future, it will be on by default and opt-out
will be flag for local use. Note that opting out will only work when running both the rpcsrv
and local with opt-out.

Resolves: #1784


adhish20:
Also need to update rpcsrv/src/proxy.rs connect() where the metadata vars that got removed in this change are used.

crates/glaredb/src/local.rs
```diff
@@ -73,24 +74,13 @@ impl LocalSession {
...
-                            return Err(anyhow!(
-                                "Specify ca-cert-path and domain in --args for TLS"
-                            ));
-                        }
-                    }
-                };
+                let url: ProxyDestination = url.try_into()?;
+
+                if !opts.disable_tls {
+                    let ca_cert = reqwest::get("https://TODO").await?.text().await?;
```
tychoish, line 80: I thought we were going to hand the cert back in the handshake with cloud rather as a new field in an existing API rather than a new one.

crates/sqlexec/src/remote/client.rs
```diff
@@ -173,15 +157,18 @@ impl RemoteClient {
     }
 
     /// Connect to a proxy destination.
-    pub async fn connect_with_proxy_destination(dst: ProxyDestination) -> Result<Self> {
-        Self::connect_with_proxy_auth_params(dst.dst.to_string(), dst.params, dst.tls_conf).await
+    pub async fn connect_with_proxy_destination(
+        dst: ProxyDestination,
+        ca_info: Option<CAInfo>,
```
tychoish, line 162: just thinking about all these API changes, but does it make sense to put the CAInfo in the ProxyDestination struct?

crates/glaredb/src/local.rs
```diff
@@ -73,25 +74,23 @@ impl LocalSession {
...
-                let client = RemoteClient::connect_with_proxy_destination(url).await?;
+
+                let mut ca_info = None;
+                if !opts.disable_tls {
+                    url.dst
+                        .set_scheme("https")
+                        .expect("not able to convert http to https");
+
+                    ca_info = Some(
+                        reqwest::get(format!("{}/api/internal/databases/authenticate", url.dst))
```
tychoish, line 86: also, is there a pooled client that we could use?

py-glaredb/src/lib.rs
```diff
@@ -124,6 +124,7 @@ fn connect(
         let mut session = if let Some(url) = conf.cloud_url.clone() {
             let exec_client = RemoteClient::connect_with_proxy_destination(
                 url.try_into().map_err(PyGlareDbError::from)?,
+                false,
```
greyscaled, line 127: I need to review this

crates/glaredb/src/args/local.rs
```diff
@@ -58,17 +58,13 @@ pub struct LocalClientOpts {
...
-    pub domain: Option<String>,
-
-    /// Path to CA certificate for rpcsrv proxy TLS (required for TLS protocol).
-    #[clap(long, default_value = "true", hide = true)]
-    pub disable_tls: bool,
+    /// Enable RPC TLS
+    ///
+    /// (Internal)
+    ///
+    /// /// In the future, we'll swap this for disable_tls
```
greyscaled, line 65:
```suggestion
    /// In the future, we'll swap this for disable_tls
```

crates/glaredb/src/local.rs
greyscaled: I don't have a good answer here, might appreciate @vrongmeal 's take

crates/glaredb/src/local.rs
```diff
@@ -73,25 +73,9 @@ impl LocalSession {
...
-                            ));
-                        }
-                    }
-                };
-
-                let mut url: ProxyDestination = url.try_into()?;
-                url = url.with_tls(tls_conf);
-                let client = RemoteClient::connect_with_proxy_destination(url).await?;
+                let client =
+                    RemoteClient::connect_with_proxy_destination(url.try_into()?, !opts.enable_tls)
```
greyscaled, line 77: the `!` here is so that when we change the args from `enable` -> `disable`, we don't need to change further logic inside `connect_with_proxy_destination`

scsmithr: I'm fine with this. We can change in the future.

greyscaled: I'm now realizing I should've done this approach with the server side too...oops

crates/glaredb/src/rpc_proxy.rs
```diff
@@ -1,12 +1,14 @@
...
 use protogen::gen::rpcsrv::service::execution_service_server::ExecutionServiceServer;
 use proxyutil::cloudauth::CloudAuthenticator;
 use rpcsrv::proxy::RpcProxyHandler;
 use std::net::SocketAddr;
-use std::path::PathBuf;
 use tonic::transport::{Identity, Server, ServerTlsConfig};
 use tracing::{debug_span, info};
 
+const CERT_PATH: &str = "/etc/certs/tls.crt";
+const CERT_KEY_PATH: &str = "/etc/certs/tls.key";
```
scsmithr, line 10: Do we know if these are always available? And can we leave a comment letting us know where they're coming from (cert manager?)?

crates/sqlexec/src/remote/client.rs
```diff
@@ -173,15 +157,18 @@ impl RemoteClient {
     }
 
     /// Connect to a proxy destination.
-    pub async fn connect_with_proxy_destination(dst: ProxyDestination) -> Result<Self> {
-        Self::connect_with_proxy_auth_params(dst.dst.to_string(), dst.params, dst.tls_conf).await
+    pub async fn connect_with_proxy_destination(
```
tychoish, line 160: my inclination would be to do the query for the cert in the main() if we're running glaredb in a mode where it _might_ do hybrid exec. (i.e. when we construct the remote client) so that, in the case that this errors, we would get that error very early rather than late.

crates/sqlexec/src/remote/client.rs
```diff
@@ -193,14 +186,33 @@ impl RemoteClient {
         }
 
         let mut dst: Endpoint = dst.try_into()?;
-        if let Some(tls_conf) = tls_conf {
-            metadata.insert(CA_CERT_PATH, tls_conf.ca_cert_path.parse()?);
-            metadata.insert(DOMAIN, tls_conf.domain.parse()?);
-            let ca = std::fs::read_to_string(tls_conf.ca_cert_path)?;
+
+        let mut api_url = "https://console.glaredb.com/api/internal/authenticate/client";
+        let host = dst.uri().host().expect("invalid host");
```
scsmithr, line 191:
```suggestion
        let host = dst.uri().host()?;
```

To not panic here

crates/sqlexec/src/remote/client.rs
```diff
@@ -193,14 +180,33 @@ impl RemoteClient {
...
+
+        if !disable_tls {
+            let mut body = HashMap::new();
+            body.insert("user", params.user);
+            body.insert("password", params.password);
+            body.insert("org_name", params.org);
+            body.insert("db_name", params.db_name);
+
+            let client = reqwest::Client::new();
+            let res = client
```
tychoish, line 204: client creation is expensive (ish) and part of my "can we do this closer to the startup or when we construct the remote client". I'm not sure how re-connections are handled, but 

another option would be to pass the credential config in as a future, if  you didn't want to resolve the certs until you try to connect. This way (I think) subsequent attempts to resolve the future would get an effectively cached value, and you would end up doing the request only once.

crates/sqlexec/src/remote/client.rs
```diff
@@ -193,14 +186,33 @@ impl RemoteClient {
...
-        if let Some(tls_conf) = tls_conf {
-            metadata.insert(CA_CERT_PATH, tls_conf.ca_cert_path.parse()?);
-            metadata.insert(DOMAIN, tls_conf.domain.parse()?);
-            let ca = std::fs::read_to_string(tls_conf.ca_cert_path)?;
+
+        let mut api_url = "https://console.glaredb.com/api/internal/authenticate/client";
+        let host = dst.uri().host().expect("invalid host");
+        if host.contains("qa.glaredb.com") {
+            api_url = "https://qa.glaredb.com/api/internal/authenticate/client";
+        }
```
scsmithr, line 194: This could be written like:

```rust
let host = dst.uri().host()?;
let api_url = if host.contains("qa.glaredb.com") {
    "https://qa.glaredb.com/api/internal/authenticate/client"
} else {
    "https://console.glaredb.com/api/internal/authenticate/client"
}

```

crates/sqlexec/src/remote/client.rs
```diff
@@ -173,15 +157,24 @@ impl RemoteClient {
...
+        Self::connect_with_proxy_auth_params(dst.dst.to_string(), dst.params, disable_tls).await
     }
 
     /// Connect to a destination with the provided authentication params.
     async fn connect_with_proxy_auth_params<'a>(
         dst: impl TryInto<Endpoint, Error = tonic::transport::Error>,
         params: ProxyAuthParams,
-        tls_conf: Option<TlsConfig>,
+        disable_tls: bool,
```
scsmithr, line 177: I would prefer we change this to `enable_tls`. We have a few unneeded `!`s right now.

crates/glaredb/src/args/local.rs
```diff
@@ -58,17 +58,13 @@ pub struct LocalClientOpts {
...
-    /// Path to CA certificate for rpcsrv proxy TLS (required for TLS protocol).
-    #[clap(long, default_value = "true", hide = true)]
-    pub disable_tls: bool,
+    /// Enable RPC TLS
+    ///
+    /// (Internal)
+    ///
+    /// In the future, we'll swap this for disable_tls
+    #[clap(long, hide = true)]
+    pub enable_tls: bool,
```
tychoish, line 67: I would probably be inclined to make this (and below) still be disable_tls, but have it default to true, so that you can enable it by doing `--disable_tls=false` (which is I think the state that it was in, and what made it opt-in before?)

crates/sqlexec/src/remote/client.rs
tychoish: I think there's decent cause to do the resolution of the cert during client creation not during connection (which would bring back tls config) and if the Option<> was populated it'd be enabled and if it weren't it would be disabled (right?)

crates/sqlexec/src/remote/client.rs
greyscaled: The reason for doing so here is relative to the common call from Python and Local. I can take a look at lifting it up, but I think it means changing the Python logic and I wasn't wanting to do that

crates/glaredb/src/local.rs
```diff
@@ -73,25 +73,18 @@ impl LocalSession {
...
-                };
+                let mut dst: ProxyDestination = url.try_into()?;
+
+                // Review: I had wanted to do this in
+                // RemoteClient::connect_with_proxy_auth_params
+                if opts.enable_tls {
+                    dst.dst
+                        .set_scheme("https")
+                        .expect("failed to upgrade scheme from http to https");
+                }
```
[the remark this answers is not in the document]
greyscaled: > this URL is the cloud api for the proxy, or is this the API URL that we're getting the cert from?

I feel confused by the former, so I'll try to answer:

This is for calling rpc proxy (ie: https://remote.glared.com:...), and is performed by tonic:

https://github.com/GlareDB/glaredb/blob/grey/tls/crates/sqlexec/src/remote/client.rs#L219

To see it's exact shape, see here:

https://github.com/GlareDB/glaredb/blob/8dddb8262d75cf1943b1033a3306fc2aa30834b6/crates/sqlexec/src/remote/client.rs#L416-L434

crates/sqlexec/src/remote/client.rs
```diff
@@ -193,14 +180,33 @@ impl RemoteClient {
         }
 
         let mut dst: Endpoint = dst.try_into()?;
-        if let Some(tls_conf) = tls_conf {
-            metadata.insert(CA_CERT_PATH, tls_conf.ca_cert_path.parse()?);
-            metadata.insert(DOMAIN, tls_conf.domain.parse()?);
-            let ca = std::fs::read_to_string(tls_conf.ca_cert_path)?;
+
+        let mut api_url = "https://console.glaredb.com/api/internal/authenticate/client";
```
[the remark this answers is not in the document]
greyscaled: I need more context here - the reason we can't pass it down from args or otherwise, easily, is because it would require an end user to set it.

The logic I added (which I admit looks a bit strange) is to construct the API URL from the connection string they use.,

so for example a user will do:

```
\open glaredb://user:pass@org.remote.glaredb.com:6443/deployment
```

and from that we know to use the `console.glaredb.com/api`.

crates/glaredb/src/rpc_proxy.rs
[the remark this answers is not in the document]
greyscaled: I absolutely agree

crates/glaredb/src/rpc_proxy.rs
```diff
@@ -37,28 +33,22 @@ impl RpcProxy {
...
                 .serve(addr)
                 .await?
-        } else if let (Some(server_cert_path), Some(server_key_path)) =
-            (server_cert_path, server_key_path)
-        {
-            let cert = std::fs::read_to_string(server_cert_path)?;
-            let key = std::fs::read_to_string(server_key_path)?;
+        } else {
+            let cert = std::fs::read_to_string(CERT_PATH)?;
+            let key = std::fs::read_to_string(CERT_KEY_PATH)?;
```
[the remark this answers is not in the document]
greyscaled: I'm also in favor

crates/glaredb/src/args/local.rs
greyscaled: Yes good point: context for this change was that when I set the default value, I couldn't override it.

I'll show you what I mean:

```shell
$ cargo run --bin glaredb -- --enable-tls=false
   Compiling sqlexec v0.5.0 (/home/greyb/Repos/glaredb/crates/sqlexec)
   Compiling pgsrv v0.5.0 (/home/greyb/Repos/glaredb/crates/pgsrv)
   Compiling rpcsrv v0.5.0 (/home/greyb/Repos/glaredb/crates/rpcsrv)
   Compiling glaredb v0.5.0 (/home/greyb/Repos/glaredb/crates/glaredb)
    Finished dev [unoptimized + debuginfo] target(s) in 4m 57s
     Running `target/debug/glaredb --enable-tls=false`
error: unexpected value 'false' for '--enable-tls' found; no more were expected

Usage: glaredb --enable-tls

For more information, try '--help'
```

crates/sqlexec/src/remote/client.rs
```diff
@@ -46,30 +47,11 @@ pub struct ProxyAuthParams {
     pub compute_engine: Option<String>,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct TlsConfig {
```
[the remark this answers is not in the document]
greyscaled: I think improving the DX here would be great; I might need help to accomplish it, as I'm sorting of hitting my "first ever time coding in Rust" limit 😂 .

In other words; the things you're asking for and spotting I agree with, but feel like iterating on (if it's acceptable to do so), or maybe get some help on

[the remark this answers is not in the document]
greyscaled: > set the values to the defaults we expect

What would be the default for `ca_cert`?

crates/glaredb/src/args/local.rs
```diff
@@ -58,17 +58,13 @@ pub struct LocalClientOpts {
...
-    /// Path to CA certificate for rpcsrv proxy TLS (required for TLS protocol).
-    #[clap(long, default_value = "true", hide = true)]
-    pub disable_tls: bool,
+    /// Enable RPC TLS
+    ///
+    /// (Internal)
+    ///
+    /// /// In the future, we'll swap this for disable_tls
+    #[clap(long, hide = true)]
+    pub enable_tls: bool,
```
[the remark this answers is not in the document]
greyscaled: See here: https://github.com/GlareDB/glaredb/pull/1767#discussion_r1334879492

crates/sqlexec/src/remote/client.rs
greyscaled: Fair, I am slightly worried about how that logic gets shared between Local and Python.

Is this something we can follow-up optimize (unless my worry is unfounded)?

crates/glaredb/src/local.rs
vrongmeal: We don't need to worry about a pooled client for an HTTP request here. Seems like this was removed.

crates/glaredb/src/local.rs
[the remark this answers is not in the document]
tychoish: I think part of my confusion is that I typically don't think about grpc as being HTTPS (even though it is; I think it will do http, but it downcasts to json, which you often want to avoid)

crates/sqlexec/src/remote/client.rs
tychoish: makes sense. I think it probably still makes sense, but definitely can save that kind of refactoring for later.

crates/sqlexec/src/remote/client.rs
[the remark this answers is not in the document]
tychoish: I think it'd be better to have it be settable via commandline args and then overridable with defaults if people don't set it, rather than hardcode our cloud deployment at this point that feels deep in the middle of the code.

crates/glaredb/src/args/local.rs
greyscaled: I might be able to get this: https://github.com/clap-rs/clap/issues/1649#issuecomment-1156638220

crates/glaredb/src/local.rs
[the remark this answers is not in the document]
greyscaled: My understanding is that we need to change the scheme from `http` -> `https` here, and the assumption is based purely on the tonic example:

https://github.com/hyperium/tonic/blob/d5c14fa20623a189771d993cb019697565393df6/examples/src/tls/client.rs#L18-L19

crates/glaredb/src/args/local.rs
```diff
@@ -58,13 +58,20 @@ pub struct LocalClientOpts {
...
+    /// Note: in the future, this will be 'on' by default
+    #[clap(long, default_value="true", action = clap::ArgAction::Set, hide = true)]
+    pub disable_tls: bool,
+
+    /// Address of the GlareDB cloud server.
+    ///
+    /// (Internal)
+    ///
+    #[clap(long, default_value = "https://console.glaredb.com", hide = true)]
+    pub cloud_api_addr: String,
```
tychoish, line 74: I'd take out `api` of this var.

tychoish (approved):
let's add the TODOs to the epic?

crates/sqlexec/src/remote/client.rs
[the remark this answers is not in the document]
greyscaled: I used an arg to accomplish this (cloud_addr)

crates/sqlexec/src/remote/client.rs
greyscaled: FYI: I merged without this change, because it required updating error/signature. If I was more familiar with Rust and the cb I would've been more able to make that change.

greyscaled:
> let's add the TODOs to the epic?

Issues created and added.

Issues now on the epci:
- https://github.com/GlareDB/glaredb/issues/1828
- https://github.com/GlareDB/glaredb/issues/1829

Additional issues:
- https://github.com/GlareDB/glaredb/issues/1830
- https://github.com/GlareDB/glaredb/issues/1831

PR #1781, opened 2023-09-19
feat: Runtime group exec pull up in EXPLAIN statements

### LINKED ISSUE #1754: RuntimeGroupExec not pulled up through FilterExec
## Description

For this query:

```
select count(*) as query_count,
       max(e.elapsed_compute_ns) as max_elapsed
  from bq_segment.glaredb_prod.execution_metric e 
  inner join pg_prod.public.users u on e.user_id = u.id
  inner join users u2 on u.email = u2.email
  where e.timestamp::timestamp > now() - interval '3 day'
```

The `RemoteScanExec` is for the bigquery table `bq_segment.glaredb_prod.execution_metric`.

```
│               │                                 FilterExec: timestamp@1 > 1694262538161859386                                                           │
│               │                                   RuntimeGroupExec: runtime_preference=remote                                                           │
│               │                                     RepartitionExec: partitioning=RoundRobinBatch(8), input_partitions=1                                │
│               │                                       RemoteScanExec: projection=[elapsed_compute_ns, timestamp, user_id] 
```

gruuya:
Instead of performing a "manual" physical optimization for `RuntimeGroupExec`, which in case of `EXPLAIN` doesn't work since the entire input is stringified by that point, rely on DataFusion's default physical planner handling of `EXPLAIN`s. To achieve this simply add the corresponding optimizer to the internal session state while creating it.

Closes #1754

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ To see all options use `--help`:
 
 ### Hybrid Execution
 
-1. Sign up at <https://console.cloud.com> for a **free** fully-managed
+1. Sign up at <https://console.glaredb.com> for a **free** fully-managed
    deployment of GlareDB
 2. Copy the connection string from GlareDB Cloud, for example:
 
--- a/crates/sqlexec/src/context/local.rs
+++ b/crates/sqlexec/src/context/local.rs
@@ -30,6 +30,7 @@ use std::slice;
 use std::sync::Arc;
 use tokio_postgres::types::Type as PgType;
 
+use datafusion_ext::runtime::group_pull_up::RuntimeGroupPullUp;
 use uuid::Uuid;
 
 use super::{new_datafusion_runtime_env, new_datafusion_session_config_opts};
@@ -81,7 +82,8 @@ impl LocalSessionContext {
             .with_extension(Arc::new(catalog_mutator))
             .with_extension(Arc::new(native_tables.clone()))
             .with_extension(Arc::new(TempCatalog::default()));
-        let state = SessionState::with_config_rt(conf, Arc::new(runtime));
+        let state = SessionState::with_config_rt(conf, Arc::new(runtime))
+            .add_physical_optimizer_rule(Arc::new(RuntimeGroupPullUp {}));
 
         let df_ctx = DfSessionContext::with_state(state);
         df_ctx.register_variable(datafusion::variable::VarType::UserDefined, Arc::new(vars));
--- a/testdata/sqllogictests/explain.slt
+++ b/testdata/sqllogictests/explain.slt
@@ -5,3 +5,28 @@ explain select 1;
 
 statement ok
 explain analyze select 1;
+
+# Test for #1754
+# Ensure `RuntimeGroupExec` is pulled as far up as possible in the `EXPLAIN`ed output
+statement ok
+create temp table t1754 as values (1, 'one'), (2, 'two'), (3, 'three');
+
+query TT
+explain
+select * from t1754
+where column1 != 2
+order by column2
+limit 1;
+----
+logical_plan Limit: skip=0, fetch=1
+  Sort: t1754.column2 ASC NULLS LAST, fetch=1
+    Filter: t1754.column1 != Int64(2)
+      TableScan: t1754 projection=[column1, column2]
+physical_plan RuntimeGroupExec: runtime_preference=local
+  GlobalLimitExec: skip=0, fetch=1
+    SortPreservingMergeExec: [column2@1 ASC NULLS LAST], fetch=1
+      SortExec: fetch=1, expr=[column2@1 ASC NULLS LAST]
+        CoalesceBatchesExec: target_batch_size=8192
+          FilterExec: column1@0 != 2
+            RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
+              MemoryExec: partitions=1, partition_sizes=[2]
--- a/crates/datafusion_ext/src/runtime/group_pull_up.rs
+++ b/crates/datafusion_ext/src/runtime/group_pull_up.rs
@@ -3,17 +3,18 @@ use datafusion::config::ConfigOptions;
 use datafusion::error::{DataFusionError, Result};
 use datafusion::physical_optimizer::PhysicalOptimizerRule;
 use datafusion::physical_plan::aggregates::AggregateExec;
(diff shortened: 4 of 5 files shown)


crates/sqlexec/src/context/local.rs
```diff
@@ -81,7 +82,8 @@ impl LocalSessionContext {
             .with_extension(Arc::new(catalog_mutator))
             .with_extension(Arc::new(native_tables.clone()))
             .with_extension(Arc::new(TempCatalog::default()));
-        let state = SessionState::with_config_rt(conf, Arc::new(runtime));
+        let state = SessionState::with_config_rt(conf, Arc::new(runtime))
+            .add_physical_optimizer_rule(Arc::new(RuntimeGroupPullUp {}));
```
gruuya, line 86: I didn't do this for `RemoteSessionContext` since from what I can see it only handles the actual execution of the physical plans, as opposed to creating/optimizing them like in the case of a `LocalSessionContext`.

On a related note, when testing this on the qa instance I've noticed that `SessionState` seems to be serialized from the remote and sent back to the local machine? This is because I don't see `RuntimeGroupPullUp` in the available optimizers inside of `Session::create_physical_plan`, even though it is there when the `Session` is created. If this returned `SessionState` originates from this part of the logic then the full end-to-end test can be done once this change is deployed to the remote GlareDB instance in the cloud. Otherwise I'd need to make an equivalent change to `RemoteSessionContext` as well.

testdata/sqllogictests/explain.slt
```diff
@@ -5,3 +5,28 @@ explain select 1;
...
+    Filter: t1754.column1 != Int64(2)
+      TableScan: t1754 projection=[column1, column2]
+physical_plan RuntimeGroupExec: runtime_preference=local
+  GlobalLimitExec: skip=0, fetch=1
+    SortPreservingMergeExec: [column2@1 ASC NULLS LAST], fetch=1
+      SortExec: fetch=1, expr=[column2@1 ASC NULLS LAST]
+        CoalesceBatchesExec: target_batch_size=8192
+          FilterExec: column1@0 != 2
+            RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
+              MemoryExec: partitions=1, partition_sizes=[2]
```
gruuya, line 32: This is a Draft PR because I'm debugging a seemingly faulty test comparison/failure
```sql
  2023-09-19T11:23:43.570162Z ERROR testing::slt::cli: Error while running test `sqllogictests/explain`, error: test fail: query result mismatch:
[SQL] explain
select * from t1754
where column1 != 2
order by column2
limit 1;
[Diff] (-expected|+actual)
    logical_plan Limit: skip=0, fetch=1
      Sort: t1754.column2 ASC NULLS LAST, fetch=1
        Filter: t1754.column1 != Int64(2)
          TableScan: t1754 projection=[column1, column2]
    physical_plan RuntimeGroupExec: runtime_preference=local
      GlobalLimitExec: skip=0, fetch=1
        SortPreservingMergeExec: [column2@1 ASC NULLS LAST], fetch=1
          SortExec: fetch=1, expr=[column2@1 ASC NULLS LAST]
            CoalesceBatchesExec: target_batch_size=8192
              FilterExec: column1@0 != 2
                RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
-                 MemoryExec: partitions=1, partition_sizes=[2]
+                 MemoryExec: partitions=1, partition_sizes=[2]
at /Users/markogrujic/GlareDB/glaredb/testdata/sqllogictests/explain.slt:14

    at crates/testing/src/slt/cli.rs:320 on main ThreadId(1)

Error: Test failures
error: test failed, to rerun pass `-p testing --test sqllogictests`
```
I did try a combination of tabs/spaces/newlines but to no avail.

crates/sqlexec/src/context/local.rs
scsmithr: > I didn't do this for RemoteSessionContext since from what I can see it only handles the actual execution of the physical plans, as opposed to creating/optimizing them like in the case of a LocalSessionContext.

Yep, we're trying to keep the planning/optimization on the local side.

> On a related note, when testing this on the qa instance I've noticed that SessionState seems to be serialized from the remote and sent back to the local machine?

So you're probably hitting this issue because we're creating a new session state when we "attach" a remote session:

https://github.com/glaredb/glaredb/blob/244817e6394f31eccb1de98b1830a39b2c91c14c/crates/sqlexec/src/context/local.rs#L115-L131

It's not the cleanest right now, but was the easiest thing to do. I'm open to alternatives to creating the session states/contexts that make these things less likely to trip someone up.

gruuya: Oh makes sense. Yeah, I think that's fine for the time being, I just need to add our optimizer rule there too.

testdata/sqllogictests/explain.slt
scsmithr: Yeah this is a bit weird, tried some whitespace changes locally too and didn't work.

Also what's the plan for the repartition exec and different number of threads?

```
-               RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
-                 MemoryExec: partitions=1, partition_sizes=[2] 
+               RepartitionExec: partitioning=RoundRobinBatch(8), input_partitions=1
+                 MemoryExec: partitions=1, partition_sizes=[2]
```

gruuya: For now I omitted this test, hope to get one added in some other PR.

README.md
```diff
@@ -77,7 +77,7 @@ To see all options use `--help`:
 
 ### Hybrid Execution
 
-1. Sign up at <https://console.cloud.com> for a **free** fully-managed
+1. Sign up at <https://console.glaredb.com> for a **free** fully-managed
```
scsmithr, line 80: thank you 🙏

scsmithr (approved):

testdata/sqllogictests/explain.slt
vrongmeal: @gruuya can we create an issue for the whitespace thing? I'll have to dig a bit deeper into it.

gruuya: Certainly, will open one.

gruuya: Opened #1793 to address this faulty test.

Issue #1776, opened 2023-09-18
Perf: use range requests when possible for parquet files


universalmind303:
We can use range requests to optimize our queries on remote datasets. 

## Metadata only count
```sql
select count(*) from 'https://datasets.clickhouse.com/hits_compatible/hits.parquet';
```

We can fulfill this entire query using only the metadata, but instead we try to pull the entire 13GB file into memory, then perform the count. 

## column selection 
```sql
select station_id, name from 'https://localhost/bikeshare_stations.parquet
```
Given the metadata from the footer, we can issue a request for just `station_id` and `name` columns without having to pull in the entire dataset into memory. drastically cutting 

## limit 
```sql
select * from 'https://localhost/bikeshare_stations.parquet limit 1
```
the `RowGroups` in the metadata contain a field `num_rows`. Given the `num_rows` field, we can safely determine how many row_groups we need to pull in. 

## parallelize over the row groups.
instead of fetching everything in a single request, we could fetch the metadata & issue N requests based on N row_groups. This results in more network requests, but should be faster as it'd be less IO constrained. 

I think we'd probably want some kind of threshold on here though, like if the file is < `SOME_THRESHOLD` we don't parallelize, but anything > `SOME_THRESHOLD`, we use parallelization.

universalmind303:
Thinking more about this more.. and I keep coming back to wanting/needing a custom physical planner. I think we _could_ do this with a rewriter rule or physical optimizer rule but it just seems inefficient because we can create that physical node pretty easily when converting the logical plan. 

@scsmithr  @gruuya @vrongmeal any thoughts?

vrongmeal:
(conversation on slack):

cory:
> so for the metadata only count, i'd imagine instead of rewriting an existing physical plan, we could check the logical plan during physical planning. If it meets this "fast path" criteria
> ```
> LogicalPlan::Aggregate{scan: LogicalPlan::Scan(parquet_object_store)), expr: AggregateExpr(Count(1))) => do fast path
> LogicalPlan::Aggregate(other) => not fast path
> ```
> so pretty much my idea is that the fast path criteria can easily be determined up front without having to retroactively apply it. (also avoiding a bunch of dyn casting by checking the physical plan nodes)
> i think the use of a custom planner would also make the DDLRewriter obsolete as we could do that all in a single pass too.


Vaibhav:
[...]

PR #1789, opened 2023-09-19
feat: enable file statistics

universalmind303:
now we can get some metadata only queries like this

```
> explain select count(*) from 'hits.parquet';
┌───────────────┬───────────────────────────────────────────────────────────────────────────────┐
│ plan_type     │ plan                                                                          │
│ ──            │ ──                                                                            │
│ Utf8          │ Utf8                                                                          │
╞═══════════════╪═══════════════════════════════════════════════════════════════════════════════╡
│ logical_plan  │ Aggregate: groupBy=[[]], aggr=[[COUNT(UInt8(1))]]↵                            │
│               │   TableScan: /Users/xxxxxxxxxxxxx/Downloads/hits.parquet projection=[WatchID] │
│ physical_plan │ ProjectionExec: expr=[99997497 as COUNT(UInt8(1))]↵                           │
│               │   EmptyExec: produce_one_row=true↵                                            │
│               │                                                                               │
└───────────────┴───────────────────────────────────────────────────────────────────────────────┘

### CODE REVIEW

File: crates/datasources/src/object_store/http.rs [hunk 1]
```diff
@@ -82,4 +90,80 @@ impl ObjStoreAccess for HttpStoreAccess {
             e_tag: None,
         })
     }
+
+    async fn into_table_provider(
```
universalmind303, line 94: Http needs a special implementation as it doesn't support globbing.

so we need to create a separate provider for each location.. 

Now that we are using the file statistics, we cant just throw them into a file group like we were before
but instead actually need to do a Union
otherwise queries like `count(*)` will only use the statistics from the first location.

universalmind303: we already had the `MultiSourceTableProvider` but it was in `sqlbuitlins` so i just moved it into `datasources

crates/datasources/src/object_store/mod.rs
```diff
@@ -35,6 +37,74 @@ pub mod http;
 pub mod local;
 pub mod s3;
 
+pub struct MultiSourceTableProvider {
```
universalmind303, line 40: this was just moved from `sqlbuiltins`

tychoish (commented):
the code looks reasonable enough to me 😀 

I think regression tests for:
- ndjson extension checking
- ensure that we do metadata only scans for counts when we expect? (it's easy to regress on these things and we don't know catch it, but users _often_ do.) 

Other things:
- tighter commit message on the PR (I think either the conventional thing, or `<component>: <user impact>`) if you plan to squash. (also you/we'd called this "metadata-only operations" or something when we talked yesterday, which is a great way to talk about this. 
- when using glaredb on local files, there's also (maybe) some metadata in files that we could use for quick counts/etc?

universalmind303:
> ensure that we do metadata only scans for counts when we expect? (it's easy to regress on these things and we don't know catch it, but users often do.)

so this is actually just plugging in to the object store & datafusions native handling, we were just missing out on a lot of these optimizations because we were not fetching the file statistics. For fileformats other than parquet, the `infer_file_statistics` is a noop & just returns the default values. 

> tighter commit message on the PR (I think either the conventional thing, or <component>: <user impact>) if you plan to squash. (also you/we'd called this "metadata-only operations" or something when we talked yesterday, which is a great way to talk about this.

yeah i'm terrible at commit messages. I prefer doing squash commits & using the PR title as the commit. That way you don't even have to think about it.

tychoish:
> yeah i'm terrible at commit messages. I prefer doing squash commits & using the PR title as the commit. That way you don't even have to think about it.

I definitely agree and am a squash/merge kind of guy. I thought your ndjson and `use file statistics` commits were better than the PR title. Anyway, I think as a (very loose) guideline PR titles that:
- have the kind of commit (fix/feature/etc. I actually care least about this, but it seems like other folks do it, so...) 
- the major subsystem
- what the impact to users is.

Sorry for perhaps being too terse earlier, I was thinking about having this optimization in local parquet files? It's perhaps ok if we handle this as part of future work.

universalmind303:
partially closes #1776

We can still make some further optimizations such as row group parallelization & column parallelization

universalmind303:
we should follow up this PR with #1791

universalmind303:
dropping this in here to show the perf gains
this branch: 125ms
main: 459ms

```
glaredb  universalmind303/file-statistics ❯ timeit {./target/release/glaredb -q "select count(*) from 'https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2023-01.parquet'"}
┌─────────────────┐
│ COUNT(UInt8(1)) │
│              ── │
│           Int64 │
╞═════════════════╡
│         3066766 │
└─────────────────┘
125ms 269µs 291ns
glaredb  universalmind303/file-statistics ❯ timeit {./glaredb-main -q "select count(*) from 'https://d37ci6vzurychx.cloudfront.net/trip-data/yellow_tripdata_2023-01.parquet'"}
┌─────────────────┐
│ COUNT(UInt8(1)) │
│              ── │
│           Int64 │
╞═════════════════╡
│         3066766 │
└─────────────────┘
459ms 394µs 167ns
```

PR #1794, opened 2023-09-20
fix: Trim whitespaces when checking SLT output

### LINKED ISSUE #1793: Chore: Flaky SLT verification logic
## Description

While trying to add tests for #1781, I ran into an odd issue whereby a SLT test errors out, but there is no apparent difference between the expected and actual output:

```diff
 2023-09-20T04:59:52.295968Z ERROR testing::slt::cli: Error while running test `sqllogictests/explain`, error: test fail: query result mismatch:
[SQL] explain
select * from t1754
where column1 != 2
order by column2
limit 1;
[Diff] (-expected|+actual)
    logical_plan Limit: skip=0, fetch=1
      Sort: t1754.column2 ASC NULLS LAST, fetch=1
        Filter: t1754.column1 != Int64(2)
          TableScan: t1754 projection=[column1, column2]
    physical_plan RuntimeGroupExec: runtime_preference=local
      GlobalLimitExec: skip=0, fetch=1
        SortPreservingMergeExec: [column2@1 ASC NULLS LAST], fetch=1
          SortExec: fetch=1, expr=[column2@1 ASC NULLS LAST]
            CoalesceBatchesExec: target_batch_size=8192
              FilterExec: column1@0 != 2
                RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
-                 MemoryExec: partitions=1, partition_sizes=[2]
+                 MemoryExec: partitions=1, partition_sizes=[2]
at /Users/markogrujic/GlareDB/glaredb/testdata/sqllogictests/explain.slt:14
```

I tried a variety of tabs/spaces/newlines, but it didn't help. Also looking at the invisible characters explicitly shows no difference as well.

vrongmeal:
The explain test failed due to not matching a new line character at the
end. It's much simpler to omit any leading or trailing whitespaces from
a row given the nature of SLTs.

Fixes #1793

### DIFF
--- a/testdata/sqllogictests/explain.slt
+++ b/testdata/sqllogictests/explain.slt
@@ -5,3 +5,34 @@ explain select 1;
 
 statement ok
 explain analyze select 1;
+
+# Test for #1754
+# Ensure `RuntimeGroupExec` is pulled as far up as possible in the `EXPLAIN`ed output
+#
+# Omitting this test for now.
+# See https://github.com/GlareDB/glaredb/pull/1781#discussion_r1330072240
+
+halt
+
+statement ok
+create temp table t1754 as values (1, 'one'), (2, 'two'), (3, 'three');
+
+query TT
+explain
+select * from t1754
+where column1 != 2
+order by column2
+limit 1;
+----
+logical_plan Limit: skip=0, fetch=1
+  Sort: t1754.column2 ASC NULLS LAST, fetch=1
+    Filter: t1754.column1 != Int64(2)
+      TableScan: t1754 projection=[column1, column2]
+physical_plan RuntimeGroupExec: runtime_preference=local
+  GlobalLimitExec: skip=0, fetch=1
+    SortPreservingMergeExec: [column2@1 ASC NULLS LAST], fetch=1
+      SortExec: fetch=1, expr=[column2@1 ASC NULLS LAST]
+        CoalesceBatchesExec: target_batch_size=8192
+          FilterExec: column1@0 != 2
+            RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
+              MemoryExec: partitions=1, partition_sizes=[2]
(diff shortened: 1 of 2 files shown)


testdata/sqllogictests/explain.slt
```diff
@@ -5,3 +5,34 @@ explain select 1;
...
+  Sort: t1754.column2 ASC NULLS LAST, fetch=1
+    Filter: t1754.column1 != Int64(2)
+      TableScan: t1754 projection=[column1, column2]
+physical_plan RuntimeGroupExec: runtime_preference=local
+  GlobalLimitExec: skip=0, fetch=1
+    SortPreservingMergeExec: [column2@1 ASC NULLS LAST], fetch=1
+      SortExec: fetch=1, expr=[column2@1 ASC NULLS LAST]
+        CoalesceBatchesExec: target_batch_size=8192
+          FilterExec: column1@0 != 2
+            RepartitionExec: partitioning=RoundRobinBatch(10), input_partitions=1
```
gruuya, line 37: One concern here is that this may fail for people with different number of cores, check this run for instance (not sure how it passed here): https://github.com/GlareDB/glaredb/actions/runs/6235138391/job/16923733930

I'm ok with omitting this particular test from the pr, so that I can find some generic-enough variant at some later point.

vrongmeal: I'll just let this test be there. We can edit it later when you find a better alternative.

scsmithr: Could we do something like what we do with replacing environment variables in queries? So we would have something like:

```
RepartitionExec: partitioning=RoundRobinBatch(${NUM_CPUS}), input_partitions=1
```

and we set `NUM_CPUS` in the test runner.

gruuya: Yeah, that sounds like a good solution.

tychoish (approved):

PR #1796, opened 2023-09-20
ci: improve labeling of slt tests

tychoish:
Was working through some stuff with Grey, and realize that the labels
and divisions between steps here were confusing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,8 +104,30 @@ jobs:
           just python build
           just python test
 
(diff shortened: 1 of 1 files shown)


tychoish:
Ok after poking at this, my "oh it'll be easy to split this into separate steps" turns out to actually be something of a nightmare. This is bigger than I wanted to go for this ticket, but will linger in the back of my mind, for refactoring at some point (mostly so  that the integration tests are a bit easier to orchestrate locally.

PR #1797, opened 2023-09-21
feat: Generic external Delta and Iceberg tables

### LINKED ISSUE #1780: Question, does glaredb support external table using delta table format ?
is this planned to be supported ?

`con.execute(f'''CREATE or replace external table xxx AS SELECT * FROM delta_scan('/lakehouse/default/Tables/xxxx') ''')`

currently only views a support, just to be very clear, I expect external table to support extra statistics :)

gruuya:
The PR introduces a `StorageOptions` metastore options struct, which is aimed to be flexible enough to capture config options from various object stores for the `object_store` crate, as well as differing use cases in a very compact way (basically a map).

Building on this, the PR also add support for:
- Delta Lake databases in something other than S3 (though with Unity catalog still required); still need to add tests for this.
- External Delta tables in arbitrary object stores (covered with tests).
- External Iceberg tables in arbitrary object stores (covered with tests yet).

Closes #1780


crates/sqlexec/src/parser/options.rs
```diff
@@ -202,6 +203,18 @@ impl fmt::Display for StmtOptions {
...
 }
 
+impl TryFrom<&mut StmtOptions> for StorageOptions {
+    type Error = ParserError;
+
+    fn try_from(value: &mut StmtOptions) -> Result<Self, Self::Error> {
+        let mut inner = BTreeMap::new();
+        for (key, value) in value.m.iter() {
+            inner.insert(key.clone(), value.clone().parse_opt()?);
+        }
```
gruuya, line 213: We could also perform validation here that the key is one of:
- [Azure options](https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html#variants)
- [S3 options](https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html#variants)
- [Google options](https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html#variants)

crates/arrow_util/src/pretty.rs
gruuya, line 1: Changes in this file are by clippy.

crates/protogen/src/metastore/types/options.rs
```diff
@@ -424,6 +420,33 @@ impl From<DeltaLakeUnityCatalog> for options::DeltaLakeUnityCatalog {
...
+// parameters across different object stores, as well as support different optional parameters in
+// differing use-cases (e.g. creating a Delta database vs a Delta table).
+// List of supported config options in `object_store` crate by object store type:
+// - [Azure options](https://docs.rs/object_store/latest/object_store/azure/enum.AzureConfigKey.html#variants)
+// - [S3 options](https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html#variants)
+// - [Google options](https://docs.rs/object_store/latest/object_store/gcp/enum.GoogleConfigKey.html#variants)
+#[derive(Debug, Default, Clone, Arbitrary, PartialEq, Eq, Hash)]
+pub struct StorageOptions {
+    pub inner: BTreeMap<String, String>,
+}
```
gruuya, line 435: This is the core of the proposed implementation.

For a bit of context, I initially went with hard-coded structs to try and capture the various storage options in various situations. That turned out to be too verbose and repetitive for my liking, with having to cover a lot of different dimensions (external DBs vs tables, Delta vs Iceberg, GCS vs S3 vs ...). 

So I ended up going with this solution instead, which has significantly less complexity I believe. Arguably we lose some amount of transparency and validation at each step in the pipeline, but the counter-argument would be that we should do validation at initial setup anyway (i.e. when creating the external DB/table in the first place).

In addition, there's a great deal of overlap with how we treat regular files (`TableOptions::S3`, `TableOptions::Gcs`) on object stores and the proposed Delta/Iceberg storage options handling. The present implementation could conceivably be used to consolidate those as well (though I didn't want to try that without a discussion, or in this PR for that matter).

scsmithr (approved):
This looks great, cleaned up some of the existing code.

Also really appreciate the added tests for thsi.

crates/protogen/src/metastore/types/options.rs
scsmithr: I agree that this is much simpler than explicitly listing out the options. We originally did that because the first implementation was for GCS, and all that needed was the service account and location, and then we had S3, which now needed more stuff. So really the code we ended up with was an evolution of adding stuff over time and not planning it up front.

I also agree there is a good bit of overlap with table options. I had a TODO or comment (or even slack message) saying that I wanted to try to consolidate the individual object store files with the data lake stuff. I would be in favor of moving towards that.

gruuya: Got it, makes sense.

crates/sqlexec/src/parser/options.rs
tychoish: It would be a programmer error if we had another type though, correct?

crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -228,18 +226,21 @@ impl<'a> SessionPlanner<'a> {
                     other => return Err(internal!("Unknown catalog type: {}", other)),
                 };
 
+                let mut storage_options = StorageOptions::try_from(m)?;
+                if let Some(creds) = creds_options {
+                    storage_options_with_credentials(&mut storage_options, creds);
```
tychoish, line 231: shouldn't we return storage_options rather than pass the return by value? I might be new to the idiom, but this always looks harder?

also similar idiom/style question: do we want function names to look like verb-phrases? (e.g. get_storage..., or resolve_storage?)

crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -497,6 +498,28 @@ impl<'a> SessionPlanner<'a> {
...
             }
+            TableOptions::DELTA | TableOptions::ICEBERG => {
+                let location = m.remove_required("location")?;
+
+                let mut storage_options = StorageOptions::try_from(m)?;
+                if let Some(creds) = creds_options {
+                    storage_options_with_credentials(&mut storage_options, creds);
+                }
+
+                // TODO: Perform a HEAD request to validate the provided config options?
```
tychoish, line 509: I sort of feel like we should do this lazily (e.g. not check until it gets used to avoid having an error on startup that would cause it to get stuck), but maybe provide an option to validate the catalog which would just connect to everything as another operation/report.

universalmind303 (approved):

vrongmeal (approved):

crates/datasources/src/lake/mod.rs
```diff
@@ -23,83 +23,51 @@ pub enum LakeStorageOptionsError {
...
-                    "google_service_account_key".to_string(),
-                    creds.service_account_key,
-                );
-            }
-            Self::Local => (),
-        }
-        opts
-    }
+/// Create an object store from the provided storage options.
+pub fn storage_options_into_object_store(
```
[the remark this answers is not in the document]
gruuya: > "into" in the function name was confusing given the "from" in the comment

In general true, but in Rust specifically there is a meaningful/tight relationship between the two: https://doc.rust-lang.org/rust-by-example/conversion/from_into.html

The function is converting the dictionary of object store config options into a object store implementation itself.

crates/protogen/src/metastore/types/options.rs
```diff
@@ -424,6 +420,33 @@ impl From<DeltaLakeUnityCatalog> for options::DeltaLakeUnityCatalog {
     }
 }
 
+// Options for a generic `ObjectStore`; to make them as versatile and compact as possible it's just
+// a wrapper for a map, like in `delta-rs`, except here it's a `BTreeMap` instead of a `HashMap`,
+// since the former is Hash unlike the latter. This enables us to capture a variety of different
```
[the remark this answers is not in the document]
gruuya: Actually, in Rust saying "A is B" usually means that the struct/enum A implements the trait B. So in this case it is not about space efficiency, but rather trait limitations of a `HashMap`. So we can only use `BTreeMap` here since parent structs also need the `Hash`.

crates/sqlexec/src/parser/options.rs
gruuya: Not necessarily; it would for example allow us to propagate some other props that could be internally useful for us, but not for building the object store itself (also the key is actually just a `String` here, I'm just considering whether to validate the possible values of that string).

crates/sqlexec/src/planner/session_planner.rs
gruuya: The goal here is to "hydrate" the supplied options with additional options extracted from the provided credentials object. The name is a bit cumbersome I admit; but still I feel it's better to update the options in place rather than create a new map and re-populate all the options again.

crates/sqlexec/src/planner/session_planner.rs
gruuya: Yeah, makes sense. In any case, I'll leave it for another PR as this one is getting quite large.

crates/protogen/src/metastore/types/options.rs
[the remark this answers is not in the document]
tychoish: your explanation makes sense, it just seems like: 

> Options for a generic `ObjectStore`; to make them as versatile and compact as possible

seemed to indicate that it was about space compactness, so the comment as a whole really makes things _less_ clear. and not more clear.

If it's just about trait requirements, then it's fine to just do it as is and we probably don't need a comment at all because the compiler enforces the right thing anyway and the comment just makes a normal thing look weird.

crates/sqlexec/src/planner/session_planner.rs
tychoish: yeah, that makes sense. I guess I'd be inclined to do that kind of thing in a method on the options rather than a function but it doesn't matter much.

crates/sqlexec/src/planner/session_planner.rs
tychoish: sounds good.... does it make sesne file an issue for this kind of thing? I always feel like code TODOs make bad project management tools 😂

tychoish (approved):
I don't want to block this at all; I think there are a couple of little follow ups (either filing issues for cleanup tasks or clarifying a comment) that should probably be done _before_ we merge this, but these are _super minor_ things mostly as a kindness to our future selves 😂

crates/sqlexec/src/planner/session_planner.rs
gruuya: Fair point, opened https://github.com/GlareDB/glaredb/issues/1817 to track that.

crates/sqlexec/src/planner/session_planner.rs
gruuya: > I guess I'd be inclined to do that kind of thing in a method on the options

True, that is a better solution. However this comment
https://github.com/GlareDB/glaredb/blob/fdbf9118078a97dbc8836836f2f75f22ef5c9b55/crates/protogen/src/lib.rs#L5
dissuaded me from putting that there, even though (alongside with `storage_options_into_object_store`) it fits better there.

tychoish: You're probably more in touch with the idiom on this kind of thing, and there's often a very legitimate reason to avoid collecting "model" logic in the these packages that can't depend on anything else internally. 

At the same time, the thing that happens with a strict "don't put logic here" rule, is that you end up with all the logic in different places and that's bad too. I wonder if: 
- we can put logic here without dependencies on other package for small things
- we can import/re-export these models via other packages and put the logic in _those_ packages just for clarity?
- decide that it's fine as it is and that the confusion (potential) is ok for the medium term, and ignore.

crates/sqlexec/src/planner/session_planner.rs
gruuya: Follow-up PR for that issue: https://github.com/GlareDB/glaredb/pull/1819

PR #1819, opened 2023-09-25
feat: Implement Delta/Iceberg external table validation during initial setup

### LINKED ISSUE #1817: Validate storage options when creating a Delta/Iceberg table in an object store
## Description

Could be done via a HEAD request, however the `location` option provided in such a case is a directory path, and not a concrete object, so this won't work.

On the other hand `list`-ing from the table root dir could be an overkill/expensive.

gruuya:
Align handling of Delta and Iceberg external tables with other variants by eagerly validating the object store connectivity.

Closes #1817

### DIFF
--- a/crates/sqlexec/src/planner/errors.rs
+++ b/crates/sqlexec/src/planner/errors.rs
@@ -94,6 +94,20 @@ impl From<crate::errors::ExecError> for PlanError {
     }
 }
 
+macro_rules! impl_from_dispatch_variant {
+    ($FromType:ty) => {
+        impl From<$FromType> for PlanError {
+            #[inline]
+            fn from(err: $FromType) -> Self {
+                PlanError::Dispatch(crate::dispatch::DispatchError::from(err))
+            }
+        }
+    };
+}
+impl_from_dispatch_variant!(datasources::lake::delta::errors::DeltaError);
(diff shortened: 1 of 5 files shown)


crates/sqlexec/src/planner/errors.rs
```diff
@@ -94,6 +94,20 @@ impl From<crate::errors::ExecError> for PlanError {
...
+            fn from(err: $FromType) -> Self {
+                PlanError::Dispatch(crate::dispatch::DispatchError::from(err))
+            }
+        }
+    };
+}
+impl_from_dispatch_variant!(datasources::lake::delta::errors::DeltaError);
+impl_from_dispatch_variant!(datasources::lake::iceberg::errors::IcebergError);
+impl_from_dispatch_variant!(datasources::object_store::errors::ObjectStoreSourceError);
+
```
gruuya, line 110: Feels like there might be a cleaner way to do this, but no luck with a generic from along the lines 
```rust
impl<E: Into<crate::dispatch::DispatchError>> From<E> for PlanError {
    fn from(err: E) -> Self {
        PlanError::Dispatch(crate::dispatch::DispatchError::from(err))
    }
}
```

Since this leads to "conflicting implementations of trait `From<DispatchError>` for type `PlanError`"

scsmithr (approved):

Issue #1805, opened 2023-09-22
tpch q1 fails converting to polars


universalmind303:
## Description
trying to run some of the TPCH benchmarks & came across this issue

```py
import glaredb

con1 = glaredb.connect()
con1.execute("create temp table if not exists lineitem as select * from './tpch/tables_scale_1/lineitem.parquet'")

query_str = """
select
    l_returnflag,
    l_linestatus,
    sum(l_quantity) as sum_qty,
    sum(l_extendedprice) as sum_base_price,
    sum(l_extendedprice * (1 - l_discount)) as sum_disc_price,
    sum(l_extendedprice * (1 - l_discount) * (1 + l_tax)) as sum_charge,
    avg(l_quantity) as avg_qty,
    avg(l_extendedprice) as avg_price,
    avg(l_discount) as avg_disc,
    count(*) as count_order
from
    lineitem
where
    l_shipdate <= '1998-09-02'
group by
    l_returnflag,
    l_linestatus
order by
    l_returnflag,
    l_linestatus;
"""


df = con1.sql(query_str).to_polars()
print(df)
```

```
thread 'thread 'thread 'glaredb-python-thread-1glaredb-python-thread-7glaredb-python-thread-0' panicked at '' panicked at '' panicked at 'assertion failed: `(left == right)`
  left: `LargeUtf8`,
 right: `Utf8`: StringArray expects DataType::Utf8assertion failed: `(left == right)`
  left: `LargeUtf8`,
 right: `Utf8`: StringArray expects DataType::Utf8assertion failed: `(left == right)`
  left: `LargeUtf8`,
  ```

PR #1821, opened 2023-09-25
perf: additional tpch benchmarks

universalmind303:
this adds a bit more robust tpch benchmark that tests glaredb against some other systems.
Currently it compares against polars and duckdb. I'd like to add datafusion and spark, but this PR is already HUGE.

Most of this code is taken from the [polars tpch benchmarks](https://github.com/pola-rs/tpch)

still todo
- [ ] validate results of queries
- [ ] plot the output to something more meaningful than a csv file


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "benchmarks/tpch/tpch-dbgen"]
```
universalmind303, line 1: I wanted to avoid using submodules, but C libraries are kinda a PITA to work with, so this seemed to be the only manageable way of importing this without copy/pasting the entire repo into our benchmarks.

py-glaredb/Cargo.toml
```diff
@@ -29,3 +29,7 @@ uuid = "1.4.1"
 async-trait = "0.1.72"
 anyhow = "1.0.72"
 once_cell = "1.18.0"
+
+[profile.release]
+codegen-units = 1
+lto = "fat"
```
scsmithr, line 35: oh boy

.gitmodules
scsmithr: I think that's fine. If you're not running benchmarks, then no need to init the submodule.

My concern with this however is generating the the data can be pretty slow. I didn't read through all of this to see if we're downloading data we've generated or not, so unsure if this is valid concern.

benchmarks/tpch/utils.py
```diff
@@ -0,0 +1,96 @@
...
+INCLUDE_IO = bool(os.environ.get("INCLUDE_IO", False))
+
+
+def _scan_ds(con, path: str, name):
+    path = f"{path}.{FILE_TYPE}"
+
+    q = f"create temp table {name} as select * from parquet_scan('{path}');"
+
+    con.execute(q)
+    return name
```
scsmithr, line 27: Would we also be able to have runs that use `create table ...`? We could show both, but the queries using delta would be more representative of typical workloads. We'd also need to make sure that `con` is initialized with a database backed by a directory.

py-glaredb/Cargo.toml
universalmind303: I can remove this if you want, kinda forgot it was in this PR. I wanted to run the benches with the fastest & smallest possible binary.

scsmithr: Did you see a difference?

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,52 @@
...
+
+gen_scale_factor n: requirements
+	make -C tpch-dbgen all
+	cd tpch-dbgen && ./dbgen -vf -s {{n}} && cd ..
+	mkdir -p "tables_scale_{{n}}"
+	mv tpch-dbgen/*.tbl tables_scale_{{n}}/
+	{{VENV_BIN}}/python prepare_files.py {{n}} 
+	rm -rf tables_scale_{{n}}/*.tbl
+
+@run_private name $SCALE_FACTOR="1":
```
tychoish, line 34: looks like just has a https://github.com/casey/just#private-recipes private recipe capability, maybe better than this form?

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,52 @@
...
+	{{VENV_BIN}}/python -m {{name}}_queries.executor 
+
+run_glaredb $SCALE_FACTOR="1": 
+	just run_private "glaredb" $SCALE_FACTOR
+run_polars $SCALE_FACTOR="1": 
+	just run_private "polars" $SCALE_FACTOR
+run_duckdb $SCALE_FACTOR="1": 
+	just run_private "duckdb" $SCALE_FACTOR
```
tychoish, line 42: seems like we could just put the {{name}} variable in the `run` target and this would just cut out a layer of process nesting?

py-glaredb/Cargo.toml
universalmind303: I didn't compare them, but we probably should...

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,52 @@
...
+
+run n_runs="1" sf="1": requirements
+  #!/usr/bin/env bash
+  n=0
+  while [ $n -lt {{n_runs}}  ]; do
```
tychoish, line 47: We're running these tests in gpd,gpd,gpd order rather than ggg,ppp,ddd order. does this matter? Do they do something to flush ram buffers between runs?

I'd be inclined, also when running in CI to have each target 

I think I'd be inclined to have a "special" target named "all" that ran all three suites, and otherwise just have `run` with `suite` and `scale` all do the same thing.

benchmarks/tpch/polars_queries/q16.py
```diff
@@ -0,0 +1,43 @@
...
+import polars_queries.utils as polars_utils
+import polars as pl
+from linetimer import CodeTimer, linetimer
+
+Q_NUM = 16
+pl.Config().set_fmt_float("full")
+
+
+@linetimer(name=f"Overall execution of polars Query {Q_NUM}", unit="ms")
+def q():
```
tychoish, line 13: why not call these functions `main`? 

^ wrote the above before I realized that this is all imported queries. 

do we need to commit these or can we checkthem out as part of the setup from the polars repo?

.gitmodules
universalmind303: > My concern with this however is generating the the data can be pretty slow. I didn't read through all of this to see if we're downloading data we've generated or not, so unsure if this is valid concern.

generating the data is pretty slow, especially on larger scale factors, but this seemed like the most reproducible way without having to manage the datasets in s3 or gcs. They are all created on the spot.

benchmarks/tpch/glaredb_queries/q1.py
gruuya, line 1: This file seems to be empty?

Also I'm potentially missing some discussion, but why aren't glaredb and duckdb re-using the existing sql queries (instead those were removed and the queries have been duplicated into the respective duckdb/glaredb modules)?

Otherwise great stuff, very excited to see some numbers from these!

benchmarks/tpch/utils.py
universalmind303: added a flag to use tables instead of temp tables

benchmarks/tpch/glaredb_queries/q1.py
universalmind303: q1 was empty because of this issue. 
https://github.com/GlareDB/glaredb/issues/1805
I was using polars to validate the data, but since i removed that logic, we could probably add it back in.

tychoish:
@universalmind303 other question maybe post/link to an example output file for the test data? 

- do we want to have python lint/fmt targets?

benchmarks/tpch/glaredb_queries/q1.py
universalmind303: > Also I'm potentially missing some discussion, but why aren't glaredb and duckdb re-using the existing sql queries (instead those were removed and the queries have been duplicated into the respective duckdb/glaredb modules)?

so there are some _slight_ differences in the SQL syntax between glaredb & duckdb. Instead of conditional logic for each query that deviates, it seemed to make the most sense to have them declare their own query.

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,52 @@
...
+	rm -r .venv
+
+clean-tables:
+	rm -r tables_scale_*
+
+clean: clean-tpch-dbgen clean-venv
+
+gen_scale_factor n: requirements
+	make -C tpch-dbgen all
+	cd tpch-dbgen && ./dbgen -vf -s {{n}} && cd ..
```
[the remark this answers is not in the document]
universalmind303: not sure i follow?

benchmarks/tpch/justfile
universalmind303: lol whoops. I thought i was using private recipes. they are `_` not `@`

benchmarks/tpch/prepare_files.py
```diff
@@ -0,0 +1,116 @@
+import sys
+
+import polars as pl
+
+scale_fac = int(sys.argv[1])
+
+h_nation = """n_nationkey
```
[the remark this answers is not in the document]
universalmind303: yeah, its some eval stuff happening here. 

This file was one that was taken from the polars TPCH runner

```
    df = pl.read_csv(
        f"tables_scale_{scale_fac}/{name}.tbl",
        has_header=False,
        separator="|",
        try_parse_dates=True,
        new_columns=eval(f"h_{name}"), # this line here
    )
```

benchmarks/tpch/justfile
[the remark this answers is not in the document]
tychoish: Oh this is just shell scripting nit/nonsense (pushd/popd are like CD but they have a stack of your previous directory changes so you don't have to pair the pop case.

`pushd tpch-dbgen; ./dbgen -vfs {{n}}; popd` 

(I don't think the `cd` can fail, because if it did (permissions non-existantce, line 27 would also fail.) 

I presume that just has the same kind of make semantics with each shell line being run in a different bash session?

benchmarks/tpch/justfile
tychoish: `@` apparently follows make's silence thing, but putting it on the target is kinda boss. 😂

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,52 @@
+set fallback
+VENV := ".venv"
+VENV_BIN := ".venv/bin"
+
+
+@venv:
+	@python3 -m pip install virtualenv > /dev/null
```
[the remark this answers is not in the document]
universalmind303: > other justfiles use a couple of spaces rather than an apparent tab.

my editor did the tabs. no standard fmtter for justfiles, so it didnt auto fix. 

> also here and 12,13, redirecting stdout to /dev/null just seems like it would be confusing? can we use a quiet mode instead?

the quiet mode doesn't seem to work here & dumps out a wall of text, so like in `py-glaredb/justfile` we just redirect to /dev/null to make it actually quiet.

.gitmodules
tychoish: the options seem like:
- submodules may be fine, actually
- `git clone` in a justfile target. (basically a sub-module except there aren't top-level references for it, and may be easier to test out?) 
- write a script to generate and upload files to GCS, and run it in CI when one of the input files changes.

benchmarks/tpch/polars_queries/q16.py
universalmind303: > do we need to commit these or can we checkthem out as part of the setup from the polars repo?

not sure I follow? 

are you suggesting importing them from the https://github.com/pola-rs/tpch repo?

benchmarks/tpch/README.md
```diff
@@ -0,0 +1,42 @@
+# TPCH Benchmarks for GlareDB and similar systems
```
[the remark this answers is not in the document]
universalmind303: > we shouldn't commit the pyc or __pycache__ files.

accidental.

universalmind303:
> do we want to have python lint/fmt targets?

yeah we should probably add a linter at some point. I'd like to avoid adding any more to this pretty big pr, so i'd prefer doing it in a separate PR.

benchmarks/tpch/justfile
tychoish: my first inclination also with this was that "we could surely do something with xargs", and it looks like `seq | xargs -I{} <cmd>` is the best option, and while I think this is fine in practice it is a bit weird. 

There's also `gnu parallel` which isn't installed everywhere, and is named oddly for given that we'd run it serially, just as a fancy loop, but that works, also I have learned that `zsh` has a `repeat` helper, because of course.

benchmarks/tpch/README.md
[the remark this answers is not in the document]
universalmind303: > make have a queries directory` and have duckdb, polars, and glaredb there.

I'm not a python expert, so there's probably a better way of organizing, but doing `queries/glaredb` messes with the modules & a flatter structure was a bit easier.

.gitmodules
universalmind303: > write a script to generate and upload files to GCS, and run it in CI when one of the input files changes.

I wasn't expecting us to run these benches in the CI, but instead ad-hoc

tychoish: how long do the benchmarks take to run? 

I think, even if we don't run them on every commit or even every day (as in a timer, that we use the CI infrastructure as a (potentially) more stable way to run these? There's going to be a lot of variability between peoples machines and setups, and if we detect a regression it's easier to find if we can do the binary search in parallel, at least?

scsmithr: I opened up https://github.com/GlareDB/glaredb/issues/1822 thinking we could just run them on every merge into main, with a workload size of one (and each merge would queue up benchmark runs). I think this would get us the data we want at a pretty granular level without blocking anyone's workflow.

benchmarks/tpch/glaredb_queries/q1.py
```diff
@@ -0,0 +1,44 @@
+import utils
+from datetime import datetime
+import glaredb
+from linetimer import CodeTimer, linetimer
+
+
+Q_NUM = 2
```
scsmithr, line 7:
```suggestion
Q_NUM = 1
```

scsmithr (approved):

.gitmodules
universalmind303: > I opened up https://github.com/GlareDB/glaredb/issues/1822 thinking we could just run them on every merge into main.

That sounds reasonable. scale of 1 should be pretty quick to generate & run. (<5 min).

benchmarks/tpch/README.md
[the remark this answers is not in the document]
tychoish: Oh, I didn't realize that we needed to we were importing the glare/duck/polars queries as modules. That changes things. 

I think the difference between `import queries.polars` and `import polars_queries` doesn't matter too much (modulo _maybe_ an extra `__init__.py`.

Having said that, I think the thing that confused me is that the `queries` directory, to be consistent with this naming scheme maybe ought to be `sql_queries`

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,52 @@
...
+## Set up virtual environment and install requirements
+@requirements: venv
+	{{VENV_BIN}}/python -m pip install --upgrade pip > /dev/null
+	{{VENV_BIN}}/pip install -r requirements.txt > /dev/null
+	
+clean-tpch-dbgen:
+	make -C tpch-dbgen clean
+
+clean-venv:
+	rm -r .venv
```
tychoish:
```suggestion
	rm -r {{VENV}} 
```

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,48 @@
...
+clean-tables:
+  rm -r tables_scale/
+
+clean: clean-tpch-dbgen clean-venv
+
+gen_scale_factor n: requirements
+  make -C tpch-dbgen all
+  cd tpch-dbgen && ./dbgen -vf -s {{n}} && cd ..
+  mkdir -p "tables_scale/{{n}}"
+  mv tpch-dbgen/*.tbl tables_scale/{{n}}/
```
tychoish, line 30: should this be a copy?

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,48 @@
...
+  rm -r {{VENV}}
+
+clean-tables:
+  rm -r tables_scale/
+
+clean: clean-tpch-dbgen clean-venv
+
+gen_scale_factor n: requirements
+  make -C tpch-dbgen all
+  cd tpch-dbgen && ./dbgen -vf -s {{n}} && cd ..
```
tychoish, line 28:
```suggestion
  pushd tpch-dbgen && ./dbgen -vf -s {{n}} && popd
```
https://just.systems/man/en/chapter_45.html

changing back isn't needed but makes it a bit more edit proof if you convert it to a bash script.

benchmarks/tpch/justfile
tychoish: should answer why interleave each execution environment rather than run one n times, and then run another?

benchmarks/tpch/justfile
```diff
@@ -0,0 +1,48 @@
...
+  #!/usr/bin/env bash
+  n=0
+  while [ $n -lt {{n_runs}}  ]; do
+    just _run_private glaredb {{SCALE_FACTOR}}
+    just _run_private polars {{SCALE_FACTOR}}
+    just _run_private duckdb {{SCALE_FACTOR}}
+    ((n++))
+  done
+
+q n system $SCALE_FACTOR="1":
```
tychoish, line 47: slightly longer name might be more clear `q`

tychoish (commented):
tried to answer questions,

benchmarks/tpch/justfile
universalmind303: imo `q` was pretty clear as most tpch benchmarks are `q1` `q2`... `just q 1 glaredb` seems to read well within the context of tpch.

benchmarks/tpch/prepare_files.py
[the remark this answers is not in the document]
universalmind303: >Still, why would you do it this way, over even just putting the lists of strings in as lists of strings?


Most of this code is taken from the [polars tpch benchmarks](https://github.com/pola-rs/tpch), so i didnt spend too much time refactoring as it worked as is.

benchmarks/tpch/glaredb_queries/q1.py
gruuya: Got it, thanks. 

Still after a quick glance it seems that a majority of the queries are actually identical, so they could be extracted away (I think only a handful are different)?

If we plan on adding other systems in the future (e.g. DataFusion), they could also re-use those as well. Also they should probably match one or the other variant of the queries that are currently slightly different (i.e. I assume DF query syntax matches GlareDB).

gruuya (approved):

PR #1824, opened 2023-09-25
chore(dx): misc cleanup justfile and add .ignore file for ripgrep

tychoish:
While reading through #1821, realized a couple of things that could be
cleaner.

- The `.ignore` file means that ripgrep will show us `.github` which
  is _quite_ confusing when that doesn't happen.

- We were double-ignoring shell-line echoing in the just file, and had
  a quirky way of installing virtualenv that was always noisy if you
  have virtualenv installed at the system level. This should work for
  everyone.


py-glaredb/justfile
```diff
@@ -2,15 +2,14 @@ set fallback
...
-	@python3 -m pip install virtualenv > /dev/null
-	@python3 -m virtualenv {{VENV}} > /dev/null
+	python3 -c "import virtualenv" || python3 -m pip --quiet install virtualenv
+	python3 -m virtualenv {{VENV}} --quiet
 
 ## Set up virtual environment and install requirements
-@requirements: venv
-	{{VENV_BIN}}/python -m pip install --upgrade pip > /dev/null
-	{{VENV_BIN}}/pip install -r requirements.txt > /dev/null
+@requirements: {{VENV}}
```
universalmind303, line 10: shouldn't this still be `venv` specifying that the other command `@venv` is a dependency of this one?

I'm actually surprised `just` doesn't complain with it like this as there is no command `.venv`

universalmind303 (approved):

PR #1832, opened 2023-09-26
fix(tls): NotValidForName

greyscaled:
Currently receiving an error:

`error: InvalidCertificate(NotValidForName)`

This attempts to fix that.

### DIFF
--- a/crates/glaredb/src/rpc_proxy.rs
+++ b/crates/glaredb/src/rpc_proxy.rs
@@ -47,7 +47,7 @@ impl RpcProxy {
             let cert = std::fs::read_to_string(CERT_PATH)?;
             let key = std::fs::read_to_string(CERT_KEY_PATH)?;
 
-            let identity = Identity::from_pem(cert, key);
+            let identity = Identity::from_pem(&cert, &key);
 
             server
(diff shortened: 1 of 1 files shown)


crates/glaredb/src/rpc_proxy.rs
```diff
@@ -47,7 +47,7 @@ impl RpcProxy {
             let cert = std::fs::read_to_string(CERT_PATH)?;
             let key = std::fs::read_to_string(CERT_KEY_PATH)?;
 
-            let identity = Identity::from_pem(cert, key);
+            let identity = Identity::from_pem(&cert, &key);
```
greyscaled, line 50: I think this was an "oops" on my part, see:

https://docs.rs/tonic/latest/tonic/transport/index.html#server

scsmithr: This won't fix the issue.

The function takes `AsRef<[u8]>` which means `cert` and `&cert` are both valid (since they implement `AsRef`).

https://docs.rs/tonic/latest/tonic/transport/struct.Identity.html#method.from_pem

greyscaled: Likely wildcard then, just wanted to cover my bases. I'll close.

Issue #1829, opened 2023-09-26
(python): kwarg for disable-tls


greyscaled:
https://github.com/GlareDB/glaredb/blob/d95ae28f6b53f10dd5072a12e6cff7b4bb96f730/py-glaredb/src/lib.rs#L127-L130

- implement missing kwarg
- turn on by default (see #1828)

PR #1837, opened 2023-09-27
chore(py-glaredb): keyword args for connect

greyscaled:
This alters the pyFunction `connect` to have two additional optional arguments `disable_tls` and `cloud_addr`. These arguments match those that are defined in `LocalClientOpts`.

### Before

```
glaredb.connect([data_dir_or_cloud_url], [spill_path])
```

### After

`data_dir_or_cloud_url` is positional, and all other args are keyword with defaults and variable in order

```
glaredb.connect([data_dir_or_cloud_url],
	[spill_path],
	[disable_tls],
	[cloud_addr]
)
```

#### Local

```py
import glaredb

conn = glaredb.connect()
conn.sql("SELECT 'hello local';").show()
┌─────────────────────┐
│ Utf8("hello local") │
│ ──                  │
│ Utf8                │
╞═════════════════════╡
│ hello local         │
└─────────────────────┘
```

#### Hybrid (using keyword args in any order)

```py
import glaredb

cloud_url = "<user>:<pass>@<org>.remote.qa.glaredb.com:6443/<db_name>"

conn = glaredb.connect(cloud_url, cloud_addr="https://qa.glaredb.com", disable_tls=False)
conn.sql("SELECT 'hello remote';").show()
┌──────────────────────┐
│ Utf8("hello remote") │
│ ──                   │
│ Utf8                 │
╞══════════════════════╡
│ hello remote         │
└──────────────────────┘
```

Resolves: #1829

### DIFF
--- a/py-glaredb/src/lib.rs
+++ b/py-glaredb/src/lib.rs
@@ -80,7 +80,7 @@ impl From<Option<String>> for PythonSessionConf {
 }
 
 #[pyfunction]
-#[pyo3(signature = (data_dir_or_cloud_url = None, spill_path = None, disable_tls = true, cloud_addr = String::from("https://console.glaredb.com")))]
+#[pyo3(signature = (data_dir_or_cloud_url = None, /, *, spill_path = None, disable_tls = true, cloud_addr = String::from("https://console.glaredb.com")))]
 fn connect(
     py: Python,
     data_dir_or_cloud_url: Option<String>,
--- a/py-glaredb/src/lib.rs
+++ b/py-glaredb/src/lib.rs
@@ -80,7 +80,7 @@ impl From<Option<String>> for PythonSessionConf {
 }
 
 #[pyfunction]
-#[pyo3(signature = (data_dir_or_cloud_url, spill_path = None, disable_tls = true, cloud_addr = String::from("https://console.glaredb.com")))]
+#[pyo3(signature = (data_dir_or_cloud_url = None, spill_path = None, disable_tls = true, cloud_addr = String::from("https://console.glaredb.com")))]
 fn connect(
     py: Python,
     data_dir_or_cloud_url: Option<String>,
--- a/py-glaredb/src/lib.rs
+++ b/py-glaredb/src/lib.rs
@@ -86,6 +86,8 @@ fn connect(
     py: Python,
     data_dir_or_cloud_url: Option<String>,
     spill_path: Option<String>,
+    disable_tls: Option<bool>,
+    cloud_addr: Option<String>,
 ) -> PyResult<LocalSession> {
     wait_for_future(py, async move {
         let conf = PythonSessionConf::from(data_dir_or_cloud_url);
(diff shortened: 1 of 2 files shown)


py-glaredb/src/lib.rs
```diff
@@ -86,6 +86,8 @@ fn connect(
     py: Python,
     data_dir_or_cloud_url: Option<String>,
     spill_path: Option<String>,
+    disable_tls: Option<bool>,
+    cloud_addr: Option<String>,
```
scsmithr, line 90: Ideally these (and spill path) would be kwargs. Otherwise this will get very unwieldy.

https://pyo3.rs/main/function/signature#using-pyo3signature--

greyscaled: I'll make that improvement, thanks

greyscaled: IMO I like what I got it. Do you see any benefit of `kwarg` over this?

```rust
#[pyo3(signature = (data_dir_or_cloud_url = None, /, *, spill_path = None, disable_tls = true, cloud_addr = String::from("https://console.glaredb.com")))]
fn connect(
    py: Python,
    data_dir_or_cloud_url: Option<String>,
    spill_path: Option<String>,
    disable_tls: bool,
    cloud_addr: String,
) -> PyResult<LocalSession> { }
```

- `/` defines the end of position args
- '*` defines the beginning of keyword-only args (it's a separate comma because you can't do `*arg=val`)

So this means:

- `data_dir_or_cloud_url` is always the first (positional) arg
- order after that doesn't matter
- all these extra opts have default values and are hard to misconfigure, because by the nature of using a keyword, you have to provide a value!

LMK if you disagree on this. My thinking is that if we did

```rust
#[pyo3(signature = (data_dir_or_cloud_url = None, /, **kwargs))]
fn connect(
    py: Python,
    data_dir_or_cloud_url: Option<String>,
    opts: Option<&PyDict>,
) -> PyResult<LocalSession> { }
```

There would be additional plumbing on `opts`

greyscaled: Example use:

```py
import glaredb

cloud_url = "<user>:<pass>@<org>.remote.qa.glaredb.com:6443/<db_name>"

conn = glaredb.connect(cloud_url, cloud_addr="https://qa.glaredb.com", disable_tls=False)
```

scsmithr: This looks reasonable to me, pinging @universalmind303 to see if he has opinions here.

universalmind303 (approved):
This seems reasonable to me.

A question not really related to this PR, but why do we need cloud_url and cloud_addr? couldn't we infer the addr from the url?

greyscaled:
> why do we need cloud_url and cloud_addr? couldn't we infer the addr from the url?

This is a good question and something I originally inferred in the implementation on https://github.com/GlareDB/glaredb/pull/1767

For example, see here:

https://github.com/GlareDB/glaredb/pull/1767/commits/5e44f62faf4a3f507d8772136ad9207e98ad3eb5

I admit the code looks a little..strange:

```rust
let api_url = if host.contains("qa.glaredb.com") {
            "https://qa.glaredb.com/api/internal/authenticate/client"
        } else {
            "https://console.glaredb.com/api/internal/authenticate/client"
        };
```

and it's mostly to do with the asymmetric resolution:

```
     remote.glaredb.com --> console.glaredb.com
qa.remote.glaredb.com -->          qa.glaredb.com
```

IMO if we used `remote.console.glaredb.com` this would be easier to map to in a way that didn't feel a little weird.

Regardless, one of the review comments was to pass it in as an arg, which is what we do for pgsrv and rpcsrv:

https://github.com/GlareDB/glaredb/blob/f1a540af361ac837c422bfa9619c0100d46ce0e3/crates/glaredb/src/args/mod.rs#L40-L83

I'd be down to revisit/rethink this to clean it up. Appreciate the question!

PR #1846, opened 2023-09-28
chore: Add github action to run benchmarks

### LINKED ISSUE #1822: Chore: Action for running benchmarks
## Description

The idea is to add it as a github runner, then have `runs-on: ubuntu-benchmarks-8-core` for a benchmark step in github actions.

This would give us relatively consistent results across runs. To reduce contention on this vm, we should probably only have benchmarks run on merges to main (and workflow_dispatch).

vrongmeal:
Runs the benchmarks with different scale factors (currently only set to 1) 3 times.

Fixes #1822


vrongmeal:
Here are a few TODOs:

- [ ] Save/Update each benchmark run
- [ ] Fix failing benchmarks
- [ ] Update benchmarks to use current build, and not release (cc: @universalmind303)

.github/workflows/ci.yaml
```diff
@@ -97,13 +119,45 @@ jobs:
...
 
+  benchmarks:
+    name: Benchmarks (Scale Factor = ${{ matrix.scale_factor }})
+    runs-on: ubuntu-latest-8-cores
+    needs: ["build-python"]
+    # CI on push is only triggered for main branch.
+    if: github.event_name == 'push'
+    strategy:
+      # We want all the benchmark jobs to run one by one.
+      max-parallel: 1
```
tychoish, line 137: I would have assumed that "parallel" in this case means "should we run more than one worker at a time" not "should we run more than one job on the same worker"

.github/workflows/ci.yaml
```diff
@@ -97,13 +119,45 @@ jobs:
...
 
       - name: Python Bindings Tests
         run: |
-          just python build
           just python test
 
+  benchmarks:
+    name: Benchmarks (Scale Factor = ${{ matrix.scale_factor }})
+    runs-on: ubuntu-latest-8-cores
+    needs: ["build-python"]
```
tychoish, line 132: we could make sure that this doesn't run unless all the other tests pass? 

adding extra dependencies would make things slower but no one is really waiting on this, so maybe it'd be better to not have to sometimes throw out data later if failed?

these are legit questions, I think both answers have merit (and it might make sense to do development with fewer dependencies and then add  them right before we're ready to merge), but should definitely answer them.

.github/workflows/ci.yaml
```diff
@@ -97,13 +119,45 @@ jobs:
...
             target/
+            .venv/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Python Bindings Tests
         run: |
-          just python build
           just python test
 
+  benchmarks:
```
tychoish, line 129: I'm not sure where we landed on the "do we run with this on every commit with a small scale factor on every commit and a larger matrix every ~day or n commits?

tychoish: oh I see what happens here! perf.

.github/workflows/ci.yaml
```diff
@@ -83,10 +84,31 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+      - uses: extractions/setup-just@v1
+      - uses: actions/cache@v3
+        name: Cache
+        with:
+          path: |
+            ~/.cargo/
+            target/
+            .venv/
```
tychoish, line 101: I think we want to have two caches: one with the rust dependencies and one with the python virtualenv? I'm a little worried that we'll end up with a messy cache situation...

.github/workflows/ci.yaml
```diff
@@ -97,13 +119,45 @@ jobs:
...
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - name: Run benchmarks
+        run: |
+          just bench tpch gen_scale_factor ${{ matrix.scale_factor }}
+          # Run the benchmarks 3 times.
+          just bench tpch run 3 ${{ matrix.scale_factor }}
```
tychoish, line 159: why 3?t

tychoish (commented):
>  Save/Update each benchmark run

could you say more about this?  is this the "data collection" aspect?

vrongmeal:
> > Save/Update each benchmark run
> 
> could you say more about this? is this the "data collection" aspect?

Yeah. It's regarding data collection. We have a `timings.csv` generated here. Would want to at-least save it somewhere.

.github/workflows/ci.yaml
vrongmeal: I am assuming we'd have only one worker. Also, IIRC, Github doesn't give us any control over running the next parallel job in a different runner so IG it's for the best.

.github/workflows/ci.yaml
vrongmeal: That makes sense if we're running benchmarks on each commit. These are only run if there's a merge to the main (which should only happen if the tests pass). That's a whole different conversation, though.

For now, this should work fine. We'll have to add a new job for running micro-benchmarks for each commit (when we do).

.github/workflows/ci.yaml
vrongmeal: This job currently runs benchmarks on each merge to `main`. We can go ahead and expand on this regarding what we prefer. I just wanted a job that works first.

.github/workflows/ci.yaml
vrongmeal: Yeah, I need to figure this out. It's quite messy, agreed. Also, this doesn't work so have to fix it here.

.github/workflows/ci.yaml
vrongmeal: TBH, I had this in the matrix but later realized we'd probably end up fixing a number. I don't have a genuine reason for `3`, but it seemed like a decent number for getting the average timing (when we analyze the data). We may want to change this based on the scale factor. Any thoughts? I'm thinking of more runs for a smaller scale factor. But this is just because we can, I don't think that would be required.

vrongmeal:
We have a working CI, with the following issues:

1. DuckDB query 13 errors:
```
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/runner/work/glaredb/glaredb/benchmarks/tpch/duckdb_queries/q13.py", line 42, in <module>
    q()
  File "/home/runner/work/glaredb/glaredb/benchmarks/tpch/duckdb_queries/q13.py", line 33, in q
    utils.get_customer_ds()
  File "/home/runner/work/glaredb/glaredb/benchmarks/tpch/duckdb_queries/utils.py", line 44, in get_customer_ds
    return _scan_ds(join(base_dir, "customer"))
  File "/home/runner/work/glaredb/glaredb/benchmarks/tpch/duckdb_queries/utils.py", line 31, in _scan_ds
    duckdb.sql(q)
duckdb.CatalogException: Catalog Error: Table with name "_home_runner_work_glaredb_glaredb_benchmarks_tpch_tables_scale_1_customer_parquet" already exists!
```

2. We want to save the `timings.csv`. I am thinking of opening a new issue for saving this data on GlareDB in a format that we can analyze easily, i.e., saving average of all the runs in one CI job with current timestamp and commit hash, something like:
```
timestamp,commit,system,query,time
```

tychoish:
> `timestamp,commit,system,query,time`

- commit changes on PRs, and might be confusing, might need more metadata, and also some information about which system's being used.
- time as observed where?
- I'd label time as duration (and decide on units that make sense.)
- parquet isn't a terrible format for this kind of thing.

tychoish:
> `GlareDB CI / Benchmarks (Scale Factor = ${{ matrix.scale_factor }}) (pull_request) Skipped`

seems wrong

.github/workflows/ci.yaml
tychoish: Oh, I can see arguments for 4 or 5 too, just wanted to know why we picked this :)

.github/workflows/ci.yaml
tychoish: I think we have more workers. 

Why do you think we want only one benchmarks to run one-by-one?

vrongmeal: My assumption here was we'll move to a self-hosted runner so would want to run it one-by-one. If we have more, we would want to modify this accordingly.

vrongmeal:
> > `GlareDB CI / Benchmarks (Scale Factor = ${{ matrix.scale_factor }}) (pull_request) Skipped`
> 
> seems wrong

Yeah, this is because of the `if` statement in the job. I'll move these benchmarks to another workflow file (the one with image builds since that's when we run these.

> commit changes on PRs, and might be confusing, might need more metadata, and also some information about which system's being used.

Yeah, `commit` would work for merges on `main` but we'll need more information going ahead.

> time as observed where?
> I'd label time as duration (and decide on units that make sense.)

Agreed.

> parquet isn't a terrible format for this kind of thing.

Yup, just wanted to mention the schema for the sake of comment so did it in CSV format.

tychoish:
> Yeah, commit would work for merges on main but we'll need more information going ahead.

yeah. I think we need (actions_build_id, execution_number, commit) (or maybe not commit at that point) either we just normalize it and be ok with it (parquet should compress this well, I hope), or we can have a file/table/etc. that has all the metadata, and that we link to the data collected in a run. 

I imagine also that we'll end up wanting to collect something more about the variables under test (currently `system`) as in: scale_factor, num_cores, repeat_id, db_under_test (e.g. for the other engines?) 

(and just to be clear we should collect timestamps when the test finishes running, say,) 

> Yup, just wanted to mention the schema for the sake of comment so did it in CSV format.

Totally makes sense, just wanted to be sure 😅

tychoish (approved):

PR #1848, opened 2023-09-28
chore: add question template & contact link to discord server

universalmind303:
Adds a "Question" template as well as a external link in the issues tab to our community discord similar to below

image


tychoish (commented):
I like it! Should we also link to the docs?

PR #1853, opened 2023-09-28
fix: Coalesce partitions in `RemoteScanExec`

### LINKED ISSUE #1812: RPC: Large snowflake table test failure
## Description

```
  2023-09-23T17:32:25.407456Z ERROR testing::slt::cli: Error while running test `sqllogictests_snowflake/large_table`, error: test fail: query result mismatch:
[SQL] SELECT count(*) FROM large_table;
[Diff] (-expected|+actual)
-   1847746
+   0
at /home/runner/work/glaredb/glaredb/testdata/sqllogictests_datasources_common/include/large_table.slti:10
```

vrongmeal:
Since we're executing and collecting the complete stream in `RemoteScanExec`, we need to coalesce all the partitions before executing else we end up running only the first partition.

Fixes #1812

### DIFF
--- a/crates/sqlexec/src/planner/physical_plan/remote_scan.rs
+++ b/crates/sqlexec/src/planner/physical_plan/remote_scan.rs
@@ -3,6 +3,7 @@ use datafusion::datasource::TableProvider;
 use datafusion::error::{DataFusionError, Result as DataFusionResult};
 use datafusion::execution::context::SessionState;
 use datafusion::execution::TaskContext;
+use datafusion::physical_plan::coalesce_partitions::CoalescePartitionsExec;
 use datafusion::physical_plan::display::ProjectSchemaDisplay;
 use datafusion::physical_plan::expressions::PhysicalSortExpr;
 use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
@@ -88,6 +89,12 @@ impl ExecutionPlan for RemoteScanExec {
         partition: usize,
         context: Arc<TaskContext>,
     ) -> DataFusionResult<SendableRecordBatchStream> {
+        if partition != 0 {
+            return Err(DataFusionError::Execution(
+                "RemoteScanExec only supports 1 partition".to_string(),
+            ));
+        }
+
         let provider = match &self.provider {
             ProviderReference::Provider(p) => p.clone(),
             ProviderReference::RemoteReference(_) => {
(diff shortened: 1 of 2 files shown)


crates/sqlexec/src/planner/physical_plan/remote_scan.rs
```diff
@@ -115,7 +122,12 @@ impl ExecutionPlan for RemoteScanExec {
                 .scan(&state, projection.as_ref(), &filters, limit)
                 .await?;
 
-            plan.execute(partition, context)
+            // NOTE: RemoteScanExec can only have 1 partition since we don't
+            // know about the paritions until we "execute" the remote table
+            // which happens, unfortunately, during execution here (see above).
+            // Hence, to execute the complete plan we need to coalesce the plan.
```
tychoish, line 128: is there a way to, eventually get around this?

tychoish (approved):

crates/sqlexec/src/planner/physical_plan/remote_scan.rs
scsmithr: With a bit of refactoring, I'm sure we could. Right now we're storing the `TableProvider` and so we don't have any info about the partitioning. Which leads to this above:

```
    fn output_partitioning(&self) -> Partitioning {
        Partitioning::UnknownPartitioning(1)
    }
```

The coalesce makes this correct.

What we can look into is store an `ExecutionPlan` which would let us get the partitioning info (and would avoid some of the above work for running the scan).

scsmithr (approved):

crates/sqlexec/src/planner/physical_plan/remote_scan.rs
vrongmeal: > What we can look into is store an ExecutionPlan which would let us get the partitioning info (and would avoid some of the above work for running the scan).

This would require an additional RPC call, I was under the assumption that this was removed intentionally. I remember implementing a store for execution plans which was later removed. Is it because we were on our way to serialize all execs? (IIRC data source execs require some extra effort to serialize)

scsmithr: There might've been an implementation of `StubRemoteTableProvider::scan` that did make a grpc call to create the plan, but I don't remember what it looked like. I think that is something we could add back in with relative ease. The response would just include the partitioning/execution plan id and we just wrap that in its own execution plan.

PR #1854, opened 2023-09-28
fix: create or replace

### LINKED ISSUE #1851: `CREATE OR REPLACE TABLE` doesn't replace table
## Description

It errors

```
> create or replace table hello as select 1;
Table created
> create or replace table hello as select 1;
Error: Execution error: failed to create table in catalog: Failed Metastore request: Duplicate name: hello
```

universalmind303:
closes #1851


testdata/sqllogictests/create_table.slt
```diff
@@ -9,6 +9,9 @@ set search_path to create_table_schema;
 statement ok
 create table t1 (a int, b int);
 
+statement ok
+create or replace table t1 (a int, b int);
+
```
scsmithr, line 14: Could we add a test that adds a new column for the replace? The scenario I want to see tested is:
- Create a table with 2 columns
- Replace with a table with 3 columns
- Do inserts with 3 columns work?
- What happens to existing table data?

vrongmeal (commented):
We need more tests here:
1. Replacing the existing table (and data) with a different schema like @scsmithr mentioned.
2. Tests for `OR REPLACE` external tables (can use `debug` data source here).

testdata/sqllogictests/create_table.slt
universalmind303: hmm, yeah definitely a few things need to happen in the delta writer too. Currently this just orphans the files, but we'd instead want to update the existing delta table using delta commands.

universalmind303:
looks like we may need to upstream some work to delta-rs. overwriting tables is not yet implemented. 

```
 SaveMode::Overwrite => {
                    todo!("Overwriting on create not yet implemented. Use 'write' operation instead.")
                }
```

we might be able to use `write` but i'm not sure how that will affect the delta logs & if it'll give us an accurate repr.

tychoish (commented):
seems good to me just have a few questions about the code.

Cargo.toml
```diff
@@ -22,8 +22,12 @@ serde_json = "1.0.104"
 chrono = "0.4.31"
 
 [workspace.dependencies.deltalake]
-git = "https://github.com/delta-io/delta-rs.git"
-rev = "94cb9b39fa62dcd7657752fcda141409e8dc697b"
+# git = "https://github.com/delta-io/delta-rs.git"
+# rev = "94cb9b39fa62dcd7657752fcda141409e8dc697b"
+
+git = "https://github.com/GlareDB/delta-rs.git"
+branch = "universalmind303/create-or-replace"
```
universalmind303, line 29: currently pointed to my branch until PR is merged

crates/datasources/src/native/access.rs
```diff
@@ -78,36 +79,40 @@ impl NativeTableStorage {
...
-        }
+        let tbl = {
+            let mut builder = CreateBuilder::new()
+                .with_table_name(&table.meta.name)
+                .with_object_store(delta_store);
 
-        // TODO: Partitioning
+            if or_replace {
+                builder = builder.with_save_mode(SaveMode::Overwrite);
+            }
```
scsmithr, line 92: Seems like we'll want an else here to set save mode to `ErrorIfExists`

crates/datasources/src/native/access.rs
```diff
@@ -125,7 +130,7 @@ impl NativeTableStorage {
     }
 
     pub async fn delete_table(&self, table: &TableEntry) -> Result<()> {
-        let prefix = format!("databases/{}/tables/{}", self.db_id, table.meta.id);
+        let prefix = format!("databases/{}/tables/{}", self.db_id, table.meta.name);
```
scsmithr, line 133:
```suggestion
        let prefix = format!("databases/{}/tables/{}", self.db_id, table.meta.id);
```

To avoid breakage on table rename.

scsmithr (commented):
We'll need to change back the paths to use ids instead of names.

crates/datasources/src/native/access.rs
universalmind303: oh i was just using this for testing, forgot to revert.

crates/datasources/src/native/access.rs
universalmind303: the default for `CreateBuilder` is `SaveMode::ErrorIfExists` so no need to specify it.

crates/sqlexec/src/planner/physical_plan/create_temp_table.rs
```diff
@@ -94,6 +96,7 @@ async fn create_temp_table(
     if temp_objects
         .resolve_temp_table(&plan.tbl_reference.name)
         .is_some()
+        && !plan.or_replace
```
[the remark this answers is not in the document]
universalmind303: logic is added for both native & temp tables, but not external tables.

scsmithr (approved):

universalmind303:
> We'll need to change back the paths to use ids instead of names.

just realized that reverting this caused some unwanted side effects.

It's now creating 2 catalog entries instead of just the 1. I'd expect it to use the same entry `20000` instead of creating a new entry `20001`

image

scsmithr:
> > We'll need to change back the paths to use ids instead of names.
> 
> just realized that reverting this caused some unwanted side effects.
> 
> It's now creating 2 catalog entries instead of just the 1. I'd expect it to use the same entry `20000` instead of creating a new entry `20001`
> image

The logic is wrong, copy pasted:

```rust
            Mutation::CreateTable(create_table) => {
                validate_object_name(&create_table.name)?;

                let schema_id = self.get_schema_id(&create_table.schema)?;

                // Create new entry
                let oid = self.next_oid();
                let ent = TableEntry {
                    meta: EntryMeta {
                        entry_type: EntryType::Table,
                        id: oid,
                        parent: schema_id,
                        name: create_table.name.clone(),
                        builtin: false,
                        external: false,
                        is_temp: false,
                    },
                    options: TableOptions::Internal(create_table.options),
                    tunnel_id: None,
                };

                let policy = if create_table.or_replace {
                    CreatePolicy::CreateOrReplace
                } else if create_table.if_not_exists {
                    CreatePolicy::CreateIfNotExists
                } else {
                    CreatePolicy::Create
                };

                self.try_insert_table_namespace(CatalogEntry::Table(ent), schema_id, oid, policy)?;
            }
```

We shouldn't be calling `self.next_oid()` if we're replacing the table. We should be using the same oid. Seems like create view also has this bug.

universalmind303:
> We shouldn't be calling self.next_oid() if we're replacing the table. We should be using the same oid. Seems like create view also has this bug.

so i did notice this too, I was initially trying to refactor as little as possible and just decremented it for the `replace` scenario, but I agree it makes more sense to just never increment it in the first place. 

Will update.

universalmind303:
@scsmithr, I fixed the oid logic for view & table replace. It'll use the original oid if there was one

```
> create table t1 (a int, b int);
Table created
> insert into t1 values (1, 2);
Inserted 1 row
> create view v1 as select * from t1;
View created
> create or replace view v1 as select * from t1;
View created
> select * from glare_catalog.views where builtin = false;
┌────────┬──────────────┬────────────┬─────────────┬───────────┬─────────┬──────────────────┐
│    oid │ database_oid │ schema_oid │ schema_name │ view_name │ builtin │ sql              │
│     ── │           ── │         ── │ ──          │ ──        │ ──      │ ──               │
│ UInt32 │       UInt32 │     UInt32 │ Utf8        │ Utf8      │ Boolean │ Utf8             │
╞════════╪══════════════╪════════════╪═════════════╪═══════════╪═════════╪══════════════════╡
│  20001 │        16384 │      16386 │ public      │ v1        │ false   │ SELECT * FROM t1 │
└────────┴──────────────┴────────────┴─────────────┴───────────┴─────────┴──────────────────┘
> 
```

previously it was creating 2 rows.

scsmithr (approved):

tychoish (approved):

PR #1868, opened 2023-10-02
chore: Move batch stream proto def to a new common file

scsmithr:
Moves some types around and renames some things for clarity.

This addes a new `rpcsrv/common.proto` for types that will be shared between internal and external apis.

I also went ahead and started replacing `broadcast_id` with `work_id` as I believe it's clearer. I originally wrote a lot of the code around the concept of "broadcasting" batches, but I would say what we're currently doing (and will do with distribted exec) isn't really broadcasting, it just some piece of work tied to a stream.


crates/protogen/proto/rpcsrv/common.proto
```diff
@@ -0,0 +1,31 @@
...
+  // relationship between local and remote sessions.
+  bytes session_id = 1;
+  // Unique work id.
+  bytes work_id = 2;
+
+  // IPC encoded batches.
+  bytes arrow_ipc = 3;
+
+  // TODO: Error flag to indicate if the stream errored on the client side. We
+  // don't need the error itself, we just need to know that we did error.
```
tychoish, line 24: would think that we'd want to use this flag to cancel other execution operations, somehow?

crates/protogen/proto/rpcsrv/common.proto
```diff
@@ -0,0 +1,31 @@
...
+syntax = "proto3";
+
+package rpcsrv.common;
+
+/// A single batch as the result of query execution.
+message ExecutionResultBatch {
+  // Session id.
+  //
+  // TODO: This will be changed to database_id when we remove the 1:1
+  // relationship between local and remote sessions.
```
tychoish, line 15: should we change or add a new number for this one for deprecation purposes?

tychoish (approved):
just comments, mostly about comments

crates/protogen/proto/rpcsrv/common.proto
```diff
@@ -0,0 +1,31 @@
+// Types shared between our "external" and "internal" apis.
+//
+// See
+// <https://www.notion.so/glaredb/Distributed-Exec-f1e40e1dac174762b4f0fe4f6ac03ac9?pvs=4#c986d64150ac4b079d431d3d36f25f0f>
```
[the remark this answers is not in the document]
scsmithr: Typically yes, but there's no other context around why this is split into its own file, so figured to add it in for quick reference. Once we have the proper internal/external definitions, we can remove this since the context will be clearer.

crates/protogen/proto/rpcsrv/common.proto
scsmithr: I'm inclined to say no since to the client, whether we use a session id or database id here doesn't matter as it's just some identifier for the remote session. This (and `work_id`) are to identify the session+stream this batch is for.

crates/protogen/proto/rpcsrv/common.proto
scsmithr: We want this to be self-contained as it's just one batch on a stream. Currently when we error (e.g. fail to encode batch), we just drop the stream, and the receiver side exits. But the receiver side does not know why we exited. Could have been a normal completion of the stream, or it could have been because of an error. 

This bool would just let us differentiate between these cases:

- Stream completes normally
- Stream sender errored (this bool), e.g. failed to encode batch (sender would already have the error)
- Stream receiver errored, e.g. failed to decode batch.

The additional bool to indicate a stream send error would just let us short circuit and do some early cleanup.

gruuya (approved):

PR #1872, opened 2023-10-02
chore: Python prereleases

scsmithr:


tychoish (approved):

scsmithr:
Let's hold off on merging this since it would require a slight addendum to how we release (explicitly setting the version to a dev version after release).

tychoish:
I think #2280 should probably take care of some of this (or we can file follow on work for this,)

I think we might want to have different release versions for the bindings?

In any case, this PR is probably closable?

scsmithr:
> I think #2280 should probably take care of some of this (or we can file follow on work for this,)
> 
> I think we might want to have different release versions for the bindings?
> 
> In any case, this PR is probably closable?

It's a bit more complicated than that. The versions we upload to pypi need to have that specific suffix to be marked as pre-release which is inferred from what's in the cargo.toml file. So there'd be extra steps around having to bump that version manually.

The goal here was to see if we could have automated pre-releases, but I think that might be too much to do right now.

PR #1874, opened 2023-10-03
feat: Extend SLTs to excercise remote data storage functionality

gruuya:
The simplest way to go about this seemed to be:
- extending the server command to also accept the storage config options (as the local command)
- add a script to spin up a MinIO server and create a bucket in it
- Add a relevant invocation to the `ci.yaml` file to trigger all that

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -292,4 +292,6 @@ jobs:
               --option access_key_id=$MINIO_ACCESS_KEY \
               --option secret_access_key=$MINIO_SECRET_KEY \
               --option bucket=$MINIO_BUCKET \
-              'sqllogictests_native/*' 
+              'sqllogictests/*' \
+              'sqllogictests_native/*'
+              
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -177,6 +177,9 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           SNOWFLAKE_USERNAME: ${{ secrets.SNOWFLAKE_USERNAME }}
           SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+          MINIO_ACCESS_KEY: glaredb
+          MINIO_SECRET_KEY: glaredb_test
+          MINIO_BUCKET: glaredb-test-bucket
         run: |
           # Prepare SLT (Snowflake)
           export PATH="$HOME/bin:$PATH"
@@ -225,6 +228,9 @@ jobs:
 
           # Prepare SLT (MongoDB)
           export MONGO_CONN_STRING=$(./scripts/create-test-mongo-db.sh)
+          
+          # Prepare SLT (MinIO)
+          ./scripts/create-test-minio-store.sh
 
           echo "-------------------------------- WITHOUT TUNNEL TEST --------------------------------"
           just sql-logic-tests -v --exclude '*/tunnels/ssh'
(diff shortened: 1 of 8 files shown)


.github/workflows/ci.yaml
```diff
@@ -279,3 +285,11 @@ jobs:
...
           just sql-logic-tests --rpc-test --exclude '*/tunnels/ssh' 'sqllogictests_mysql/*'
           just sql-logic-tests --rpc-test --exclude '*/tunnels/ssh' 'sqllogictests_postgres/*'
+          
+          echo "-------------------------- REMOTE DATA STORAGE TESTS --------------------------------"
+          # Test using a remote object store for storing databases and catalog; For now only on MinIO (S3)
+          just sql-logic-tests --location http://localhost:9000 \
+              --option access_key_id=$MINIO_ACCESS_KEY \
+              --option secret_access_key=$MINIO_SECRET_KEY \
+              --option bucket=$MINIO_BUCKET \
+              'sqllogictests_native/*'
```
gruuya, line 295: We probably want to include some more tests that are heavy on DDL & DML statement, so as to exercise the remote object store as much as possible. What would be some good paths to add here?

crates/glaredb/src/args/server.rs
```diff
@@ -43,6 +43,9 @@ pub struct ServerArgs {
     #[clap(short, long, hide = true, value_parser)]
     pub service_account_path: Option<String>,
 
+    #[clap(flatten)]
+    pub storage_config: StorageConfigArgs,
```
gruuya, line 47: `hide` doesn't play nicely with `flatten` it seems, so this will appear on `help` invocations.

crates/bench_runner/src/main.rs
```diff
@@ -69,6 +69,8 @@ fn main() -> Result<()> {
             None,
             None,
             None,
+            Default::default(),
+            None,
```
scsmithr, line 73: This function getting intense with all the Options.

crates/glaredb/src/args/server.rs
scsmithr: Not the worst thing. I started adding in `hide` just so users wouldn't be confused about options that they shouldn't be setting (e.g. segment keys).

scripts/create-test-minio-store.sh
```diff
@@ -0,0 +1,44 @@
...
+
+# Configure mc to connect to our above container as host
+mc config host add glaredb_minio http://localhost:9000 $MINIO_ACCESS_KEY $MINIO_SECRET_KEY
+
+# Remove the bucket if it already exists
+mc rm -r --force glaredb_minio/"$MINIO_BUCKET"
+
+# Finally create the test bucket
+mc mb glaredb_minio/"$MINIO_BUCKET"
+EOF
```
scsmithr, line 44: Crazy shell skills

scsmithr (approved):

crates/bench_runner/src/main.rs
gruuya: Yeah; I think there's a remote chance we can deprecate/consolidate `data_dir` and `service_account_key` into storage config so we'll be back to previous situation.

.github/workflows/ci.yaml
vrongmeal: I think we should run the `sqllogictests/*` tests as well using this. Most of the DDL/DML tests lie there!

vrongmeal (approved):
This generally looks good. Just one comment about adding more tests (`sqllogictests/*`) to this and then we should be good to merge!

PS: How do we feel about segregating the different SLTs into their own jobs in CI? @scsmithr @tychoish

scripts/create-test-minio-store.sh
tychoish: what's `mc` and does writing to stdin of `/bin/sh` respect a shebang? (I have no idea!) It's probably the case that you're not calling `mc` from within the mc` shell....

tychoish (approved):
I think this is fine to merge in its current form; I think in the medium term we could/should do a little cleanup with our integration test orchestration, but we don't need to block this PR on that (though @vrongmeal or @gruuya can/should coordinate on some of those details.)

scripts/create-test-minio-store.sh
gruuya: Good catch, that's just a leftover from my previous attempt to get this working. Removing now.

.github/workflows/ci.yaml
gruuya: Great, thanks for the info!

Issue #1143, opened 2023-06-19
[Epic] Support data source inserts


scsmithr:
Currently everything is read only.

To start, we can create the execution plans. Follow on work may be needed for if we want to mark data sources read-only, but this will likely coincide with some sort of permissions system.

# Tasks

- [x] #2010

# Data Sources

- [x] #1883
- [x] #1889
- [x] #1961
- [ ] inserts into BigQuery
- [ ] Inserts into Snowflake
- [ ] Delta lake
- [ ] Iceberg
- [ ] Object store (we may be able to use datafusion's stuff for this)

# Questions

- scope: share code between `COPY TO` and `INSERT INTO` and figure out how much work.
- identify limits for limits with number of values. 
- [X] read only mode

greyscaled:
I think we hold off on permissions, or just error on inserts if we have credentials that don't have insert permissions. Will want to document this behavior in glaredb docs.

greyscaled:
@vrongmeal might be working on postgres soon

greyscaled:
We're holding off on this for now for other priorities. We can revisit, and start with Pg as POC.

We will likely wait for a request.

greyscaled:
moving from 0.6.0 to https://github.com/GlareDB/glaredb/milestone/25

greyscaled:
Question (as a user/dogfooder): I am using a Postgres data source that I explicitly want to be read-only. It's critical that I do not accidentally perform any mutations (insert, update, delete, drop, alter etc etc): I simply want to _only_ select from this data source.

How can I best achieve that? And how might we consider this problem longer-term?

tychoish:
I think we should handle insert-into-object store (e.g. globbed file datasources) as part of a different epic. 

@vrongmeal and I have been talking about doing delta/iceberg as part of the current push on table-like data formats. in the near term. 

This leaves bigquery and snowflake, mostly. I'm not sure exactly what the right answer is for bigquery (aside from just writing to files in GCS?) and snowflake, which we could probably do relatively soon. 

Otherwise we're continuing to make progress here.

PR #1877, opened 2023-10-03
feat: Support `INSERT INTO` for external sources.

vrongmeal:
Lays down the framework for inserting into external sources.

Supporting inserts *should* be as simple as implementing the `insert_into` method for table providers.

This PR implements `insert_into` for Postgres datasource.

Fixes: #1883

Epic: #1143

### DIFF
--- a/crates/datasources/src/postgres/mod.rs
+++ b/crates/datasources/src/postgres/mod.rs
@@ -1,5 +1,6 @@
 pub mod errors;
 
+mod query_exec;
 mod tls;
 
 use crate::common::ssh::session::SshTunnelSession;
@@ -19,12 +20,14 @@ use datafusion::execution::context::SessionState;
 use datafusion::execution::context::TaskContext;
 use datafusion::logical_expr::{Expr, TableProviderFilterPushDown, TableType};
 use datafusion::physical_expr::PhysicalSortExpr;
+use datafusion::physical_plan::memory::MemoryExec;
 use datafusion::physical_plan::metrics::ExecutionPlanMetricsSet;
 use datafusion::physical_plan::metrics::MetricsSet;
 use datafusion::physical_plan::{
-    DisplayAs, DisplayFormatType, ExecutionPlan, Partitioning, RecordBatchStream,
+    execute_stream, DisplayAs, DisplayFormatType, ExecutionPlan, Partitioning, RecordBatchStream,
     SendableRecordBatchStream, Statistics,
 };
+use datafusion::scalar::ScalarValue;
 use datafusion_ext::errors::ExtensionError;
 use datafusion_ext::functions::VirtualLister;
 use datafusion_ext::metrics::DataSourceMetricsStreamAdapter;
@@ -48,6 +51,8 @@ use tokio_postgres::tls::MakeTlsConnect;
 use tokio_postgres::types::{FromSql, Type as PostgresType};
 use tokio_postgres::{Client, Config, Connection, CopyOutStream, NoTls, Socket};
 use tracing::{debug, warn};
+
+use self::query_exec::{create_count_record_batch, PostgresQueryExec};
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub enum PostgresDbConnection {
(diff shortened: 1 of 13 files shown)


crates/sqlexec/src/remote/rewriter.rs
```diff
@@ -15,10 +19,10 @@ use protogen::metastore::types::options::{
 /// can execute the DDL actions locally.
 ///
 /// NOTE: We should only have one DDL in a plan.
-#[derive(Debug)]
+// #[derive(Debug)]
```
vrongmeal, line 22: Will remove this.

vrongmeal:
TODO:

- [ ] Implement debug source where we can specify `runtime_preference`.
- [ ] Implement inserts for debug source
- [ ] Write tests using debug source

crates/protogen/src/sqlexec/physical_plan.rs
```diff
@@ -270,8 +270,8 @@ pub struct DeleteExec {
 
 #[derive(Clone, PartialEq, Message)]
 pub struct InsertExec {
-    #[prost(message, tag = "1")]
-    pub table: Option<TableEntry>,
+    #[prost(bytes, tag = "1")]
+    pub provider_id: Vec<u8>, // UUID
```
tychoish, line 274: I see we do this in other places, but I read this and thought "surely there's a way to avoid needing to specify the type in a comment, and it looks like https://crates.io/crates/prost-uuid/0.2.0 is a thing that exists? 

Need not change it everywhere here or even do it in this PR, but does this seem unreasonable?

crates/sqlexec/src/planner/physical_plan/insert.rs
```diff
@@ -74,10 +71,14 @@ impl ExecutionPlan for InsertExec {
...
-            if this.table.meta.is_temp {
-                this.insert_temp(context).await
-            } else {
-                this.insert_native(context).await
+            match this.provider {
+                ProviderReference::RemoteReference(_) => Err(DataFusionError::Internal(
+                    "required table provider, found remote reference to insert".to_string(),
+                )),
+                ProviderReference::Provider(provider) => {
+                    // TODO: Add background job to track storage for native tables.
```
tychoish, line 79: is this in this PR (if so, maybe edit the description to include this?) otherwise, might this be it's own ticket in the epic?

crates/protogen/src/sqlexec/physical_plan.rs
vrongmeal: Yeah, I think we can do it. Will start from this PR.

crates/sqlexec/src/planner/physical_plan/insert.rs
vrongmeal: So we have background jobs, it got removed when working on RPC stuff, I'll create an issue to track these.

crates/protogen/src/sqlexec/physical_plan.rs
vrongmeal: So, the issue here is the `uuid` version used by this library. We can implement this on our own and use it. I'll create an issue to track it.

vrongmeal: 1884

crates/sqlexec/src/remote/rewriter.rs
```diff
@@ -72,8 +75,14 @@ impl TreeNodeRewriter for DDLRewriter {
...
                 ExtensionType::Insert => {
                     let insert_into = ext.node.as_any().downcast_ref::<Insert>().unwrap();
-                    if insert_into.table.meta.is_temp {
-                        self.set(Self::InsertIntoTempTable(insert_into.table.clone()))?;
+                    if insert_into.runtime_preference == RuntimePreference::Local {
+                        let provider = match &insert_into.provider {
+                            ProviderReference::Provider(provider) => provider.clone(),
+                            ProviderReference::RemoteReference(_) => unreachable!(
+                                "required local table, found remote reference to table"
+                            ),
```
vrongmeal, line 83: Should I return an internal error instead of `unreachable!` here?

crates/datasources/src/postgres/mod.rs
```diff
@@ -642,6 +647,54 @@ impl TableProvider for PostgresTableProvider {
 
         Ok(Arc::new(exec))
     }
+
+    async fn insert_into(
+        &self,
+        state: &SessionState,
+        input: Arc<dyn ExecutionPlan>,
+        _overwrite: bool,
+    ) -> DatafusionResult<Arc<dyn ExecutionPlan>> {
```
scsmithr, line 656: I think the implementation works for a first pass. I do think there's some opportunities to avoid a single insert (where we collect everything in memory).

- We may be able to do copy in (https://docs.rs/tokio-postgres/latest/tokio_postgres/struct.Transaction.html#method.copy_in) and write to a buffer.
- We could have an alternative execution plan implementation that streams the output of `input` into postgres without the intermediate batch writing.

crates/sqlexec/src/remote/rewriter.rs
scsmithr: I think this would be considered a programmer bug. If that's the case, then the panic seems reasonable to me.

scsmithr (approved):

crates/sqlexec/src/planner/physical_plan/remote_scan.rs
```diff
@@ -31,6 +32,25 @@ pub enum ProviderReference {
...
+
+impl Eq for ProviderReference {}
+
+impl Hash for ProviderReference {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        if let Self::RemoteReference(id) = self {
+            id.hash(state)
+        }
+    }
+}
```
scsmithr, line 52: Actually this seems kind of concerning. Do we need the hash.

vrongmeal: So it's due to the `LogicalPlan` serialization stuff. I had a branch where I removed all that but it got deleted. I am planning on removing all this. It's never used but exists because of trait constraints.

crates/datasources/src/postgres/mod.rs
vrongmeal: Yup, that makes sense! I will create an issue for this.

greyscaled (commented):
This is a drive by and just double-checking:

> Fixes: https://github.com/GlareDB/glaredb/pull/1183

That issue is already closed and related to window builds, I'm double-checking this is the right issue

vrongmeal:
> This is a drive by and just double-checking:
> 
> > Fixes: #1183
> 
> That issue is already closed and related to window builds, I'm double-checking this is the right issue

Fixed in PR 🫣
It was #1883

crates/datasources/src/postgres/mod.rs
vrongmeal: #1888

adhish20 (approved):

PR #1887, opened 2023-10-05
fix: builtin table functions should not need a schema to resolve

### LINKED ISSUE #1383: Table functions shouldn't need schema?
## Context

```
〉create schema test_schema;
create schema

〉set search_path to test_schema;
set local

〉show search_path;
+-------------+
| search_path |
+-------------+
| test_schema |
+-------------+

〉create view test_view as select * from generate_series(1, 3);
Error: Error during planning: Missing table function: 'generate_series'

〉create view test_view as select * from public.generate_series(1, 3);
create view
```

## Expected

Successful query.

## Actual

Error (since schema not specified).

adhish20:
Fixes #1383


universalmind303:
can we add some slt's for this?

crates/sqlexec/src/metastore/catalog.rs
```diff
@@ -255,6 +255,30 @@ impl SessionCatalog {
...
+            .expect("object name points to invalid function");
+
+        match ent {
+            CatalogEntry::Function(function) => {
+                if function.meta.builtin && function.func_type == FunctionType::TableReturning {
+                    Some(function.clone())
+                } else {
+                    None
+                }
+            }
```
vrongmeal, line 277:
```suggestion
            CatalogEntry::Function(function) if function.meta.builtin && function.func_type == FunctionType::TableReturning => {
                Some(function.clone())
            }
```

Can simplify here.

vrongmeal (commented):
Generally, it looks good, but we do need some SLTs to go with here.

PR #1894, opened 2023-10-09
feat: Remote session context per database instead of per connection

gruuya:


crates/sqlexec/src/context/local.rs
```diff
@@ -120,9 +123,7 @@ impl LocalSessionContext {
         // The main difference here is creating a new catalog mutator without a
         // metastore client to "detach" ourselves from the metastore we're
         // connected to (likely running in-memory).
-        let vars = self
-            .get_session_vars()
-            .with_remote_session_id(client.session_id(), VarType::System);
+        let vars = self.get_session_vars();
```
gruuya, line 126: I think I should also update vars here with the db id from the client.

scsmithr: Yeah I think that would be reasonable since the "local" database we were connected to is now no longer accessible.

py-glaredb/src/session.rs
```diff
@@ -22,7 +18,6 @@ use crate::{error::PyGlareDbError, logical_plan::PyLogicalPlan, runtime::wait_fo
 #[pyclass]
 pub struct LocalSession {
     pub(super) sess: PyTrackedSession,
-    pub(super) engine: Engine,
```
scsmithr, line 25: I believe we'll want to keep this around so that's not accidentally dropped. The engine contains the background jobs worker and the metastore client workers.

gruuya: Makes sense; I brought back the engine and the `close` method, though now it only does the engine shutdown.

crates/sqlexec/src/context/remote.rs
```diff
@@ -137,42 +143,45 @@ impl RemoteSessionContext {
...
+        &self,
         table_ref: ResolvedTableReference,
         args: Option<Vec<FuncParamValue>>,
         opts: Option<HashMap<String, FuncParamValue>>,
     ) -> Result<(Uuid, Arc<dyn TableProvider>)> {
-        self.catalog
+        // TODO: Remove this lock so we're not holding it for the duration of
+        // the table load. We can do that by:
+        // 1. Storing multiple catalog versions
+        // 2. Getting a reference to the version we care about (by passing it in as an arg)
```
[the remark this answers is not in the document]
gruuya: Yup, opened https://github.com/GlareDB/glaredb/issues/1896 to track that.

scsmithr (approved):

tychoish (approved):

PR #1895, opened 2023-10-09
feat: experimental prql support

### LINKED ISSUE #1772: native PRQL support
## Description
Since GlareDB is primarily analytics focused, it'd be cool to support the analytics focused query language [PRQL](https://prql-lang.org/). 

duckdb & clickhouse do it! 😉 

### PRQL
```prql
from employees
select {id, first_name, age}
sort age
take 10
```
### SQL equivalent
```sql
SELECT
  id,
  first_name,
  age
FROM
  employees
ORDER BY
  age
LIMIT
  10
  ```
  
  
This potentially could be upstreamed to datafusion as well.

### PR DESCRIPTION
**Pull request by universalmind303**
allows users to use the experimental `prql` language instead of sql

closes #1772 

See https://prql-lang.org/ for more info.

To use, simply set the dialect
```sql
set dialect = 'prql';
```
then you can use prql syntax

```elm
from my_table 
filter some_column == "hello world"
take 1
```

To switch back to sql, just set the dialct to `sql`

```sql
set dialect = 'sql';
```

## Note: 
custom table functions & expressions are not yet supported, This gives us relatively limited prql functionality & should be marked as experimental.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -105,6 +105,21 @@ dependencies = [
 
 [[package]]
 name = "anstream"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0ca84f3628370c59db74ee214b3263d58f9aadd9b4fe7e711fd87dc452b7f163"
+dependencies = [
+ "anstyle",
+ "anstyle-parse",
+ "anstyle-query",
+ "anstyle-wincon 1.0.2",
+ "colorchoice",
+ "is-terminal",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstream"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b1f58811cfac344940f1a400b6e6231ce35171f614f26439e80f8c1465c5cc0c"
@@ -112,7 +127,7 @@ dependencies = [
  "anstyle",
  "anstyle-parse",
  "anstyle-query",
- "anstyle-wincon",
+ "anstyle-wincon 2.1.0",
  "colorchoice",
  "utf8parse",
 ]
@@ -143,6 +158,16 @@ dependencies = [
 
 [[package]]
 name = "anstyle-wincon"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c677ab05e09154296dd37acecd46420c17b9713e8366facafa8fc0885167cf4c"
+dependencies = [
+ "anstyle",
+ "windows-sys",
+]
+
+[[package]]
+name = "anstyle-wincon"
 version = "2.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "58f54d10c6dfa51283a066ceab3ec1ab78d13fae00aa49243a45e4571fb79dfd"
@@ -156,6 +181,9 @@ name = "anyhow"
 version = "1.0.75"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a4668cab20f66d8d020e1fbc0ebe47217433c1b6c8f2040faf858554e394ace6"
+dependencies = [
+ "backtrace",
+]
 
 [[package]]
 name = "apache-avro"
@@ -182,6 +210,16 @@ dependencies = [
  "typed-builder 0.14.0",
  "uuid",
  "zerocopy",
+]
+
+[[package]]
+name = "ariadne"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72fe02fc62033df9ba41cba57ee19acf5e742511a140c7dbc3a873e19a19a1bd"
+dependencies = [
+ "unicode-width",
+ "yansi",
 ]
 
 [[package]]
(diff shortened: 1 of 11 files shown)


scsmithr:
> custom table functions & expressions are not yet supported, This gives us relatively limited prql functionality & should be marked as experimental.

Is this an inherit limitation with prql (in that you can't do `from my_func(...)`) or would it just require some extra configuration on our end?

universalmind303:
> Is this an inherit limitation with prql (in that you can't do from my_func(...)) or would it just require some extra configuration on our end?

It seems possible, but requires a bit of extra work. 

It requires us to register a bunch of custom modules when we initialize the prql parser. I've been looking into how to actually implement it on our end (it seems like they do it for duckdb out of the box https://prql-lang.org/book/how-do-i/read-files.html).

scsmithr:
Can we add a change here https://github.com/glaredb/glaredb/blob/514c93912959ef63169eb1592e7fa22f4b8ccb0f/crates/pgsrv/src/handler.rs#L851-L854 to check the prql var? Would let us use prql over the postgres protocol.

universalmind303:
follow up WRT:
> Is this an inherit limitation with prql (in that you can't do from my_func(...)) or would it just require some extra configuration on our end?


It looks like there is currently no way to extend prql with our own functions. They have a [module system,](https://prql-lang.org/book/reference/spec/modules.html) but is not exposed through the public api. So currently all you could do is prepend the query with a bunch of inline variables (with no namespacing)

```elm
let generate_series = start stop step:1 -> s"select * from generate_series({start:0}, {stop:0}, {step:0})"
let csv_scan = path -> s"select * from csv_scan('{path}')"
let json_scan = path -> s"select * from json_scan('{path}')"
let parquet_scan = path -> s"select * from parquet_scan('{path}')"
# ...

<your actual query>
```

The other options would be to either open up an issue in PRQL repo to add support for glaredb dialect ([as other db's have done](https://github.com/PRQL/prql/issues/3630)), or open up an issue/pr to allow custom modules.

scsmithr:
We can start with the prefixing and seeing how that plays out.

I think we can open up an issue and see what the maintainers think.

tychoish:
>  (it seems like they do it for duckdb out of the box https://prql-lang.org/book/how-do-i/read-files.html).

There seems like there's very little PRQL code for duckdb in the prql repo, so it seems like we could open it as a PR and even maintain a patchset in the short term. (vendor copy or "weak fork") 

I'd even be find merging this, as is and doing the prefix/etc. at some later/intermediate point, particularly if it's not a huge lift

tychoish:
I was also thinking that having some _very basic_ SLTs that use prql just to make sure that we don't drop support for it (or that passing sql or prql in eachother's modes works as expected,) would be nice.

scsmithr (commented):
> I was also thinking that having some very basic SLTs that use prql just to make sure that we don't drop support for it (or that passing sql or prql in eachother's modes works as expected,) would be nice.

Could we add this?

scsmithr (approved):
PLTs look good to me.

max-sixty:
Hi from the PRQL team!

Thanks for adding PRQL to GlareDB. A couple of follow-ups come to mind, we're keen to be a good platform to build on, and so happy to add more here:
- Ideally we'd have add GlareDB as a [supported dialect](https://prql-lang.org/book/project/target.html#supported). To do that we would add `target:sql.glaredb` (or `target:sql.glare`?), and add integration tests within PRQL. Is there an existing SQL dialect we could start from?
- We added ClickHouse to our website and docs as a recommended integration when they added native support, we'd be happy to have GlareDB there too. Also happy to wait until you're more confident that it's a good fit if you prefer.
- As discussed above, the module system in PRQL is quite nascent. But it's also quite close, and so to the extent it would be helpful for you, that would be a useful impetus for us to polish it and document it. What specifically would be most helpful for GlareDB there?
- What else would be helpful from our end?

universalmind303:
@max-sixty thanks for reaching out, and great work on PRQL, I've been following the project for quite some time now!


> Ideally we'd have add GlareDB as a [supported dialect](https://prql-lang.org/book/project/target.html#supported). To do that we would add target:sql.glaredb (or target:sql.glare?), and add integration tests within PRQL. Is there an existing SQL dialect we could start from?

We use postgres as our base, so that would probably be the easiest starting point.

> We added ClickHouse to our website and docs as a recommended integration when they added native support, we'd be happy to have GlareDB there too. Also happy to wait until you're more confident that it's a good fit if you prefer.

I think that's a great idea. We'd be thrilled to be listed as a supported dialect. (I don't see us dropping support for it)

> As discussed above, the module system in PRQL is quite nascent. But it's also quite close, and so to the extent it would be helpful for you, that would be a useful impetus for us to polish it and document it. What specifically would be most helpful for GlareDB there?

When I was implementing the integration, I thought that it'd be preferred if external modules could be registered. Either through the `Options`, or passed directly into `compile`. I'm happy to open up an issue in `prql` with more details as what would be preferred from our side; we have a lot of custom table & scalar functions that probably would be best maintained in our own repo than within prql's internal modules.

max-sixty:
> When I was implementing the integration, I thought that it'd be preferred if external modules could be registered. Either through the `Options`, or passed directly into `compile`. I'm happy to open up an issue in `prql` with more details as what would be preferred from our side; we have a lot of custom table & scalar functions that probably would be best maintained in our own repo than within prql's internal modules.

Yes, interesting. I quite like it! I opened https://github.com/PRQL/prql/issues/3670

---

Keep in touch on things — feel free to put issues in for anything that would be helpful. Looking forward to collaborating.

PR #1900, opened 2023-10-10
feat: support insert into for mysql

### LINKED ISSUE #1889: Insert into for MySQL

adhish20:
Fixes #1889

### DIFF
--- a/crates/datasources/src/mysql/mod.rs
+++ b/crates/datasources/src/mysql/mod.rs
@@ -8,7 +8,7 @@ use std::task::{Context, Poll};
 
 use crate::common::ssh::session::SshTunnelSession;
 use crate::common::ssh::{key::SshKey, session::SshTunnelAccess};
-use crate::common::util::{self, create_count_record_batch};
+use crate::common::util::{self, create_count_record_batch, COUNT_SCHEMA};
 use async_stream::stream;
 use async_trait::async_trait;
 use chrono::{NaiveDate, NaiveDateTime, NaiveTime, Timelike};
@@ -282,6 +282,12 @@ impl VirtualLister for MysqlAccessor {
     }
 }
 
+#[derive(Debug, Clone)]
+enum QueryType {
+    Dql,
+    Dml,
+}
+
 pub struct MysqlTableProvider {
     predicate_pushdown: bool,
     table_access: MysqlTableAccess,
@@ -374,7 +380,7 @@ impl TableProvider for MysqlTableProvider {
             query,
             arrow_schema: projected_schema,
             metrics: ExecutionPlanMetricsSet::new(),
-            is_read_only: true,
+            query_type: QueryType::Dql,
         }))
     }
 
@@ -393,12 +399,12 @@ impl TableProvider for MysqlTableProvider {
             let mut row_sep = "";
             for row_idx in 0..batch.num_rows() {
                 write!(&mut values, "{row_sep}(")?;
-                row_sep = ",\n\t";
+                row_sep = ",";
 
                 let mut col_sep = "";
                 for col in batch.columns() {
                     write!(&mut values, "{col_sep}")?;
-                    col_sep = ", ";
+                    col_sep = ",";
 
                     let val = ScalarValue::try_from_array(col.as_ref(), row_idx)?;
                     util::encode_literal_to_text(util::Datasource::MySql, &mut values, &val)
@@ -416,7 +422,7 @@ impl TableProvider for MysqlTableProvider {
         }
 
         let query = format!(
-            "INSERT INTO {}.{} VALUES\n\t{}",
+            "INSERT INTO {}.{} VALUES {}",
             self.table_access.schema, self.table_access.name, values
         );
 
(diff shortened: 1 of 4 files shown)


crates/datasources/src/mysql/mod.rs
```diff
@@ -374,6 +375,54 @@ impl TableProvider for MysqlTableProvider {
...
+        let mut values = String::new();
+
+        let mut input = execute_stream(input, state.task_ctx())?;
+        while let Some(batch) = input.next().await {
+            let batch = batch?;
+
+            let mut row_sep = "";
+            for row_idx in 0..batch.num_rows() {
+                write!(&mut values, "{row_sep}(")?;
+                row_sep = ",\n\t";
```
tychoish, line 396: do we really need to have these new-lines and indentation?

It seems like a lot of code to add new lines and what not to statements where we're just going to pass it to another service, where either no one will appreciate the pretty formatting, or it will show up in logs where multi-line logging isn't helpful

crates/datasources/src/mysql/mod.rs
```diff
@@ -372,6 +374,62 @@ impl TableProvider for MysqlTableProvider {
...
+            }
+        }
+
+        if values.is_empty() {
+            let batch = create_count_record_batch(0);
+            let schema = batch.schema();
+            return Ok(Arc::new(MemoryExec::try_new(&[vec![batch]], schema, None)?));
+        }
+
+        let query = format!(
```
tychoish, line 418: this is sort of a rust question, but when do we do string formatting and when do we do string-building?

tychoish (dismissed):

tychoish (commented):
sorry accidental approval button

crates/datasources/src/mysql/mod.rs
adhish20: formatting is just a more readable way imo

crates/datasources/src/mysql/mod.rs
vrongmeal: Yeah, I added new lines for debugging. Helps on console!

crates/datasources/src/mysql/mod.rs
vrongmeal: What do you mean by "string-building"?

crates/datasources/src/mysql/mod.rs
tychoish: Given that people don't see it, except in logs, (and again, multi-line logging is going to be confusing in production) I think we shouldn't prioritize the new lines?

crates/datasources/src/mysql/mod.rs
```diff
@@ -372,6 +374,62 @@ impl TableProvider for MysqlTableProvider {
...
+        );
+
+        debug!(%query, "inserting into mysql datasource");
+
+        Ok(Arc::new(MysqlExec {
+            predicate: "".to_string(),
+            table_access: self.table_access.clone(),
+            accessor: self.accessor.clone(),
+            query,
+            arrow_schema: self.arrow_schema.clone(),
```
vrongmeal, line 430: This seems weird. Shouldn't this be `COUNT_SCHEMA`? Or am I missing something

adhish20: ah yes! checking

crates/datasources/src/mysql/mod.rs
tychoish: for sure, though we're doing a lot of formatting here and above (sort of proceduraly building strings) so there's a lot of (probably? depends on how it gets optimized?) little string allocations, and a buffered string builder would be more efficient (and also frankly with the exception of this line probably easier to read.

crates/datasources/src/mysql/mod.rs
vrongmeal: Yeah, we can remove new lines here.

crates/datasources/src/mysql/mod.rs
vrongmeal: A buffered builder will definitely help (setting initial capacity a bit high, if that's what you're talking about). A rust `String` is made up of `Vec` so the capacity should double once it runs out.

vrongmeal: But that doesn't compete with `format!`. I think (and hope, never actually checked) it only allocates once, by calculating all the sizes and is more readable

crates/datasources/src/mysql/mod.rs
adhish20: fixed

crates/datasources/src/mysql/mod.rs
adhish20: I have removed the tabs and new lines

crates/datasources/src/mysql/mod.rs
tychoish: So true, but on line 496, we're using this `write!` macro call, and then writing that in this string, so we're presumably calling format a bunch of times.

tychoish (approved):
seems good. is there any tresting that we can/should do/add?

adhish20:
> seems good. is there any tresting that we can/should do/add?

There is another issue to add tests for the datasource inserts

tychoish:
> There is another issue to add tests for the datasource inserts

great, lets do this!

crates/datasources/src/mysql/mod.rs
vrongmeal: Oh, I see what you're asking. It's basically the same difference as pushing into a vector and allocating a new vector. `format!`, even though it would allocate once, would allocate all again and again if used in a loop. Building the string using `write!` doubles the allocated size of string during each, making it better for such cases.

vrongmeal (approved):
Looks good!

PR #1903, opened 2023-10-11
feat: implement query `\timing` command for local repl

gruuya:
Implement a \timing command in analogy with the one in [psql](https://www.postgresql.org/docs/current/app-psql.html#APP-PSQL-META-COMMAND-TIMING):

```sql
> \timing
Timing is on
> select count(*) from lineitem;
┌─────────────────┐
│ COUNT(UInt8(1)) │
│              ── │
│           Int64 │
╞═════════════════╡
│        29999795 │
└─────────────────┘
Time: 0.067s
```

### DIFF
--- a/benchmarks/tpch/duckdb_queries/q13.py
+++ b/benchmarks/tpch/duckdb_queries/q13.py
@@ -30,9 +30,6 @@ def q():
         c_count desc
 	"""
 
-    utils.get_customer_ds()
-    utils.get_orders_ds()
-
     q_final = duckdb.sql(query_str)
 
     utils.run_query(Q_NUM, q_final)
--- a/crates/glaredb/src/args/local.rs
+++ b/crates/glaredb/src/args/local.rs
@@ -38,6 +38,9 @@ pub struct LocalClientOpts {
     #[clap(flatten)]
     pub storage_config: StorageConfigArgs,
 
+    #[clap(long, default_value = "false", hide = true)]
+    pub timing: bool,
+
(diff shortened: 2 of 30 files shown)


crates/glaredb/src/args/local.rs
```diff
@@ -38,6 +38,9 @@ pub struct LocalClientOpts {
     #[clap(flatten)]
     pub storage_config: StorageConfigArgs,
 
+    #[clap(long, default_value = "false", hide = true)]
+    pub timing: bool,
```
gruuya, line 42: This could also not be hidden, but I didn't feel like it was worth extending the existing cli args just with this.

crates/glaredb/src/local.rs
```diff
@@ -226,6 +233,7 @@ impl LocalSession {
                         self.opts.mode,
                         self.opts.max_width,
                         self.opts.max_rows,
+                        now,
```
gruuya, line 236: This means only `ExecutionResult::Query` statements will be timed, so for now no DDL/DML timing info.

benchmarks/tpch/duckdb_queries/q13.py
```diff
@@ -30,9 +30,6 @@ def q():
         c_count desc
 	"""
 
-    utils.get_customer_ds()
-    utils.get_orders_ds()
-
```
vrongmeal, line 35: thank you for fixing this!

tychoish (approved):
I like it! Fixed up the commit message, otherwise seems good.

PR #1908, opened 2023-10-11
feat: Support polars lazy frames (the lazy way)

scsmithr:
Quick fix for querying polars lazy frames by collecting the result in memory.

A more robust solution for lazy frames may be forthcoming.

### DIFF
--- a/py-glaredb/src/environment.rs
+++ b/py-glaredb/src/environment.rs
@@ -3,6 +3,7 @@ use datafusion::{
     arrow::{pyarrow::PyArrowType, record_batch::RecordBatch},
     datasource::TableProvider,
 };
+use pyo3::types::IntoPyDict;
 use pyo3::types::PyTuple;
 use pyo3::{prelude::*, types::PyType};
 use sqlexec::environment::EnvironmentReader;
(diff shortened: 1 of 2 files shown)


py-glaredb/src/environment.rs
```diff
@@ -99,6 +103,38 @@ fn resolve_polars(py: Python, var: &PyAny) -> PyResult<Option<Arc<dyn TableProvi
...
+        .unwrap();
+
+    if !var.is_instance(polars_type).unwrap() {
+        return Ok(None);
+    }
+
+    let kwargs = &[("streaming", true)];
+    let df = var.call_method("collect", (), Some(kwargs.into_py_dict(py)))?;
+
+    let arrow = df.call_method0("to_arrow").unwrap();
```
universalmind303, line 126: we probably want to call `rechunk` before calling `to_arrow`.

universalmind303: scratch that. `to_arrow` already performs a `rechunk`

scsmithr:
Checking on test failures...

universalmind303 (approved):

PR #1909, opened 2023-10-12
feat: Support python contexts, add default in-memory connection

scsmithr:
Context:

```python
with glaredb.connect() as con:
    con.execute("select 1")   
```

Adds `sql` and `execute` as top-level functions using an in memory db. Meaning you can now do this:

```python
import glaredb

glaredb.sql("select 1")
```


py-glaredb/tests/test_connection.py
```diff
@@ -0,0 +1,27 @@
...
+    out = glaredb.execute("select 1 as a").to_pandas()
+    expected = pd.DataFrame({"a": [1]})
+    assert out.equals(expected)
+
+
+# def test_default_conn_uses_same_db():
+#     # Create table
+#     glaredb.execute("create table hello (a int)")
+#     # Try to query it. This would error if we weren't using the same db.
+#     glaredb.execute("select * from hello")
```
scsmithr, line 27: Commented out since global, in-memory storage for table data is per-process and not per-engine, which causes other tests to fail.

universalmind303:
@scsmithr is this good to go? looks like It needs a rebase.

tychoish:
I'll fix the rebase and take this over.

py-glaredb/tests/test_connection.py
tychoish: should we just omit this? or is there a way to scope the execution of this test (or I guess something that we could if-against)?

scsmithr: I'd like to keep this in and open an issue.

Right now when creating an in-memory connection, we use a global object store (https://github.com/glaredb/glaredb/blob/b9e4a2f76d9fcbc78eb6f631b2b593319d30ad6f/crates/object_store_util/src/conf.rs#L10) which is "correct", but causes this issue where we don't have the isolation we'd want in tests. I'd like for us to figure out what isolation might look like (e.g. have a global hashmap and let the user provide custom name for the memory store if they choose like `glaredb.connect("memory://custom/")`).

scsmithr:
> I'll fix the rebase and take this over (if that makes sense)

That sounds good to me. I think I stopped this because I wanted to the address the `sql`/`execute` thing first (which you merged).

PR #1924, opened 2023-10-15
feat: Scheduling

scsmithr:


tychoish (commented):
I have read this in a sort of "bearing witness" way, and I think we've been using a lot of words for various components and so the code and (my?) mental model is hasn't caught up to what these constructs are called in the code... so things like `Task` (query) vs `Executor` (process) vs `Pipeline` (also query) vs `Plan` (query) (writing them out made it easier but I it was less obvious than I expected it, and the naming isn't (and probably can't be) 100% clear.

scsmithr:
Will work on documenting all of this. Currently it's in the stage of "it works" so let's go ahead and clean it up so we can iterate on this.

crates/datafusion_ext/src/vars/constants.rs
```diff
@@ -196,6 +196,14 @@ pub(super) const DIALECT: ServerVar<Dialect> = ServerVar {
     description: "Dialect of the sql engine",
 };
 
+pub(super) const ENABLE_EXPERIMENTAL_SCHEDULER: ServerVar<bool> = ServerVar {
+    name: "enable_experimental_scheduler",
+    value: &true,
```
scsmithr, line 201: Change to false before merge.

scsmithr:
- [ ] Remove enqueuing executors, pretty much combine worker and scheduler
- [ ] Executors: Pass in channel directly, mpsc -> watch
- [ ] Remove `TaskExecutor` trait, since we'll just have a single impl.

vrongmeal:
I made the above listed changes. Here's what I learned and think the remote executors would look like:

1. Remote executors and scheduler would have a bi-directional gRPC communication.
1. Once an executor is created, it notifies the remote node of its availability.
   * This would happen at regular intervals as sort of a health-check for both the scheduler and executor.
1. If the starts executing a task (local or remote), it would notify the remote node of its unavailability. (This can happen async)
   * The remote node should be able to reschedule the task if this notification hasn't reached it yet.
   * Once the execution is completed, it would notify of its availability again!
1. A stub remote executor is created on the remote scheduler node which listens for tasks (similar to other executors).
1. Once the stub receives a task, it sends it over to the remote node.
   * We need to make sure (somehow) that the local executors receive tasks before the stub (if available).

The above assumes the following:

* Any query request goes through the scheduler node.
  * Currently we can simply send the plan to scheduler node (if current node isn't it).
  * To remove the constraint, we need to figure out if all the tasks can be scheduled on the current node or not. If there are pending tasks in the queue, we should send them to the scheduler node which should be able to schedule them aptly.
* Each operation is scheduled by 1 scheduler.
  * A node can listen to as many remote nodes for tasks. This makes the node a "worker" and the nodes it listens to a "scheduler".
  * It's recommended that there's a dedicated set of scheduler nodes in a cluster and each worker listens for tasks on them.

**PS:** _These are some raw thoughts. Comments are definitely welcome. I'm going to think about them more as well._

tychoish:
> Currently we can simply send the plan to scheduler node (if current node isn't it).

I presume that there might be more than one scheduler node. We should assume that, in general most nodes will be (able) to be schedulers. We may prevent this from happening at the proxy/routing/cluster level in practice, but I think in general there's no reason why two different queries from two different connections on the same deployment need to execute on the same node.

> Once an executor is created, it notifies the remote node of its availability. [...]

There's a piece of this that I'd like to think about/talk about more in detail. In general: 
- a scheduler shouldn't need to know about the executors at all: the scheduler maintains a list of tasks and provides a way for executors (of any persuasion) to pull tasks off the queue and ensures that only one executor can pull a task at a time. 
- there is this problem of "service discovery" (how do executors discover nodes that have schedulers that they can/should ask for work, that I think we haven't really solved. There are some harder problems in this:
  - how do we avoid having _too_ much communication between nodes: if each node has 32 or more executors (number tbd but it could be 32 or even 2-4 times as many) and worker clusters start to get big than that ends up being a lot of communication and checking between nodes. We could solve this by adding another layer on the executor side. 
  - how do nodes learn about other nodes? [this gets hard, but I think for start we (cloud) can tell the glaredb instances about this on startup, they can have a global in-memory table of their cluster) and cloud could over-time inform the instances about it; there are some also some even worse solutions. I think for version 0 of this we can just pass it statically to glaredb on startup and as long as service discovery is sufficiently abstract, we can improve upon it later. 
  - how does the system shutdown cleanly?  

As a result of this and this thinking work, that we can start on next week; I'd be inclined to really focus on making the local execution-flavor of this scheduler system work well, (and perform well,) and be merged, and the remote stuff will (I think,) be pretty easy to experiment with bit-by-bit.

vrongmeal:
I think we should definitely catch up over a call. There are a few things that we need to clear out regarding "pull" based.

> We may prevent this from happening at the proxy/routing/cluster level in practice, but I think in general there's no reason why two different queries from two different connections on the same deployment need to execute on the same node.

That's not what I meant. They do not need to execute on the same node, but it depends on the size of the cluster. We can definitely have a set of dedicated scheduler nodes. But a query must go through one of those nodes. If all the nodes are schedulers, then all executors need to listen on those nodes which would definitely not be recommended.

tychoish:
> I think we should definitely catch up over a call. There are a few things that we need to clear out regarding "pull" based.

Definitely can do early next week.

Just to be clear, we don't need to decide on the specifics of the implementation of all executors (or remote executors?) in order to merge this. (e.g. the interface of the scheduler doesn't change,). 

> They do not need to execute on the same node, but it depends on the size of the cluster. 

I think maintaining the property that an instance of GlareDB cannot assume that it has exclusive access to a deployment/catalog/etc is a good one. 

What depends on the size of the cluster?

> We can definitely have a set of dedicated scheduler nodes. 

If left to my own devices I would err towards _not_ having a dedicated set of scheduler nodes. We're going for an architecture that is very distributed and stateless and also very horizontally scalable, and having more schedulers means that the system as a whole is more robust (because each scheduler becomes responsible for less work).  

> But a query must go through one of those nodes. If all the nodes are schedulers, then all executors need to listen on those nodes which would definitely not be recommended.

I think we're concerned about the same case/problem where there ends up being a sort of explosion of the number of connections between nodes (particularly because the nodes themselves are divided into many independent threads.) I think this is a very reasonable thing to be concerned about, but I'm less worried about it (now) for a couple of reasons:

- given the ability to do bi-directional streams, and the fact that we only need a connection between the nodes' scheduler and all the executor threads in the working group we're talking about (`numNodes*numNodes*numThreads`) and not `numThreads!`. This makes a big difference.
- in the worst case, schedulers don't need to have one connection for every _executor_ in the cluster: there can be a component in the system between the executors and the schedulers on every node, so that the number of actual connections is "just" `numNodes^2`. This keeps the numbers pretty small for a while. e
- the number of nodes in the cluster/group is not going to be globally *that* big, at least for a while, eventually we'll split into smaller working groups of nodes for each cluster. the size of clusters is something that we can and should play with over time, based on demand. 
- I don't think that it's necessary for all nodes with executors to necessary connect to all other nodes. It would not be bad if we said "every worker node should attempt to connect to `num` nodes (with schedulers). There's some existing p2p networking research that we can leverage (if needed) to figure out how to have a well connected network without having fully connected network of nodes.

scsmithr (commented):
Approved (can't approve own PR)

universalmind303:
One small suggestion. The `distexec` code seems pretty isolated from the rest of the code in `sqlexec`. I'm thinking it might make sense to instead move this into a separate `scheduler` crate.

tychoish:
sidebar: the thing that we're calling the scheduler here is probably better thought of as a `ExecutionQueue`, the scheduler name references a design where the scheduler is more active in deciding where the work needs to execute; and the model we have moving forward has the "scheudler" just responsible for storing the partitioned work and distributing it when workers (local or remote) ask for it. 

I agree that we should eventually factor it out, but maybe best to do that on top of this, and just keep things easier to review?

tychoish (approved):

universalmind303:
> I agree that we should eventually factor it out, but maybe best to do that on top of this, and just keep things easier to review?

seems reasonable.

Issue #1885, opened 2023-10-05
Add the storage background jobs back again


vrongmeal:
## Description

Add the bg jobs back again that were removed when working on hybrid exec (RPC) stuff.

For example: https://github.com/GlareDB/glaredb/blob/7b4937a1c067e68863d2a2787ed05ccdf4c28f30/crates/sqlexec/src/planner/physical_plan/insert.rs#L118

greyscaled:
I'm not entirely sure we need to do this (at all). We should discuss further offline.

tychoish:
I think this is "gone away" for various reasons. The commit did the appropriate cleanup, and we're addressing this accounting using other means.

PR #1925, opened 2023-10-16
chore: Add storage tracking background job back

vrongmeal:
Fixes part of #1885

### DIFF
--- a/crates/sqlexec/src/planner/physical_plan/insert.rs
+++ b/crates/sqlexec/src/planner/physical_plan/insert.rs
@@ -12,11 +12,17 @@ use datafusion::physical_plan::{
     SendableRecordBatchStream, Statistics,
 };
 use datafusion::scalar::ScalarValue;
+use datasources::native::access::NativeTable;
+use datasources::native::access::NativeTableStorage;
 use futures::{stream, StreamExt};
 
 use std::any::Any;
 use std::fmt;
 use std::sync::Arc;
+
+use crate::background_jobs::storage::BackgroundJobStorageTracker;
+use crate::background_jobs::JobRunner;
+use crate::metastore::catalog::CatalogMutator;
 
 use super::remote_scan::ProviderReference;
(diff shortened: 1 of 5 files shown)


scsmithr:
Seems we're hitting a panic in one of the SLTs

crates/sqlexec/src/planner/physical_plan/insert.rs
```diff
@@ -76,8 +82,42 @@ impl ExecutionPlan for InsertExec {
                     "required table provider, found remote reference to insert".to_string(),
                 )),
                 ProviderReference::Provider(provider) => {
-                    // TODO: Add background job to track storage for native tables.
-                    Self::do_insert(provider, this.source, context).await
+                    let is_native_table = provider.as_any().downcast_ref::<NativeTable>().is_some();
```
tychoish, line 85: I guess I have a couple of questions:
- what do we do for writing storage for other non-native tables?
- is this supposed to be "bytes written since (this merges and/or the beginning of a table?)"
- it looks like "we try and enqueue a job every time there's a write, and it noops if there's something in the queue waiting to run; when the job runs (every 5m) we record the total current size; if the service shuts down while the job is in the queue, it never runs."
- don't we want something like: "we send a timestamp and a storage size (at the given moment), maybe a query ID or sequence number?, to a <service> internally, which adds it to a batch, which is flushed periodically to a delta file (parquet compresses this well, adding a bunch of rows in one shot is more effective than adding one row at a time; the data is more useful in this form, and it's possible to have "graceful shutdown" include this one last write...)

tychoish (approved):

tychoish (commented):
I think maybe it makes sense to either:
- close this without merging (as we are tracking this data using different tools,) 
- remove background jobs entirely as they don't make sense in the context of distributed execution. (I'd be willing to do this in a different PR if it doesn't make sense to do it here)

vrongmeal:
Closing in favour of #2148

PR #1926, opened 2023-10-16
feat(python): make sql methdod eager for ddl/dml operations

tychoish:
It's confusing for DDL/DML operations that the SQL method is lazy, so this PR changes the semantics to conform with that expectation. 

The one catch in this is that the object returned for eagerly executed operations is the plan that was executed, which, if run again might lead to confusing errors. I think it should be possible to construct an "empty" or "noop" plan in this case (which might error if called again, but that's probably correct.) So I want to give that a try.

### DIFF
--- a/py-glaredb/src/session.rs
+++ b/py-glaredb/src/session.rs
@@ -1,5 +1,7 @@
 use crate::execution_result::PyExecutionResult;
+use datafusion::logical_expr::LogicalPlan;
 use futures::lock::Mutex;
+use futures::TryFutureExt;
 use pyo3::prelude::*;
 use sqlexec::engine::{Engine, TrackedSession};
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


crates/sqlexec/src/planner/logical_plan.rs
```diff
@@ -101,6 +101,7 @@ pub enum LogicalPlan {
     Datafusion(DfLogicalPlan),
     /// Plans related to transaction management.
     Transaction(TransactionPlan),
+    Noop(),
```
scsmithr, line 104:
```suggestion
    Noop,
```

scsmithr (commented):
Looks reasonable to me. Could we add a quick python test to assert this behavior? E.g. create a table then try selecting from it.

PR #1927, opened 2023-10-17
feat: Implementation of sub-directory support for external native storage

### LINKED ISSUE #1873: Enable using different folders in same bucket in external object storage
## Description

Right now we only support a "root" location so that a given bucket cannot hold multiple catalogs/dbs.

gruuya:
This enables one to use different paths/folders within the same bucket for GS and S3 object store types. 

For instance in the case of MinIO:

```shell
glaredb -l http://localhost:9000/this-is-the-bucket/this/is/folder/root -o access_key_id=** -o secret_access_key=****
```

For the sake of integration testing, a fake GCS server is also added, along the lines of what is used in arrow-rs: https://github.com/apache/arrow-rs/blob/master/object_store/CONTRIBUTING.md#gcp

Closes #1873.


tychoish (commented):
this seems pretty elegant! I definitely don't have any obvious things or big things that jump out at me, though it seems like there should be a prefix cached within the storage config structs, just for (maybe?) logging or introspection in cases where the prefix/directory gets captured by the implementation of the client, but I think it's _fine_.

I don't want to be the only one to review this, and while I think at some point (before too long) we'll need to break apart the slt tests, for now it's ok.

crates/datasources/src/native/access.rs
```diff
@@ -36,35 +35,43 @@ pub use deltalake::protocol::SaveMode;
...
     /// Arcs all the way down...
     store: SharedObjectStore,
 }
 
 impl NativeTableStorage {
-    /// Create a native table storage provider from the given config.
-    pub fn from_config(db_id: Uuid, conf: StorageConfig) -> Result<NativeTableStorage> {
-        let store = conf.new_object_store()?;
-        Ok(NativeTableStorage {
+    /// Create a native table storage provider with the .
```
scsmithr, line 57: "with the ..."

tychoish (approved):

PR #1928, opened 2023-10-17
fix: Immediately update events in telemetry

vrongmeal:
NOTE: This does remove `session_query_metrics` table.


scsmithr (approved):
Fortunately (unfortunately?) `session_query_metrics` wasn't documented, so little chance that it was actually being used for anything.

crates/sqlexec/src/metrics.rs
```diff
@@ -144,6 +101,8 @@ pub struct QueryMetrics {
     pub elapsed_compute_ns: Option<u64>,
     /// Number of output rows. Currently only set for SELECT queries.
     pub output_rows: Option<u64>,
+    /// Number of bytes processed during the execution of query.
+    pub bytes_processed: Option<u64>,
```
greyscaled, line 105: (this is a total aside and just me being around): one thing I want to ensure we get the naming agreement on is "query" and "operation". An operation can have more than one query (for example: sub-query), and based on plan partitioning, can be distributed across multiple workers.

That isn't to say "execution of query" is wrong, I just want to like make sure the names and identifiers map well and so when we say bytes processed in a query, is it always a "query" or is it some piece of a plan?

scsmithr:
Seems to be a couple SLTs we'll need to remove

crates/sqlexec/src/metrics.rs
vrongmeal: What do you mean by "sub-query"? Is it simply multiple queries like:

```sql
SELECT * FROM a;
SELECT * FROM b;
```

vrongmeal:
> Seems to be a couple SLTs we'll need to remove

Yup!

crates/sqlexec/src/metrics.rs
greyscaled: What about:

```sql
SELECT *
FROM my_table
WHERE id IN (
  SELECT id
  FROM my_other_table
  WHERE organization_id = '123';
);
```

vrongmeal: I think the final result would be the total bytes processed by the complete query (including all the sub-queries).

Issue #947, opened 2023-05-04
Implement `ToSql`, `FromSql` (and other required traits) for `Decimal128`


vrongmeal:
## Context

This is required to support to and from Postgres conversions.

There are other traits related to MySQL which would be good to have as well (but currently `rust_decimal` works well for that). We might just use `rust_decimal` for `FromSql` for now (since it fully supports it), but we need both in future for binary conversions.

scsmithr:
Is this still required?

PR #1935, opened 2023-10-20
feat: Enable binary encoding when requested by the client

### LINKED ISSUE #1921: Ensure float size is correct when encoding to pg values
## Description

See https://github.com/PRQL/prql/pull/3669#issuecomment-1762545507

It looks like we're passing Float8 in the postgres row descriptor, but the column values are actually f32.

I don't have a reproducible sql example yet...

vrongmeal:
Missing:
- [ ] Binary encoding for decimals (numerics) #947

Fixes #1921

### DIFF
--- a/crates/testing/tests/sqllogictests/tests.rs
+++ b/crates/testing/tests/sqllogictests/tests.rs
@@ -110,3 +110,39 @@ SELECT public_key
         Ok(())
     }
 }
+
+pub struct PgBinaryEncoding;
+
+#[async_trait]
+impl FnTest for PgBinaryEncoding {
+    async fn run(
+        &self,
+        _config: &Config,
+        client: TestClient,
+        _vars: &mut HashMap<String, String>,
+    ) -> Result<()> {
+        let client = match client {
+            TestClient::Pg(client) => client,
+            TestClient::Rpc(_) => {
+                return Err(anyhow!("cannot run pg binary encoding test on rpc"));
+            }
+        };
+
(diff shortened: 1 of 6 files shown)


crates/testing/tests/sqllogictests/tests.rs
```diff
@@ -110,3 +110,39 @@ SELECT public_key
...
+        let rows = client.query("select 1, 2.2", &[]).await?;
+        test_assert!(
+            rows.len() == 1,
+            anyhow!("number of rows returned ({}) != 1", rows.len())
+        );
+
+        let int: i64 = rows[0].try_get(0)?;
+        test_assert!(int == 1, anyhow!("int value from column 0 ({}) != 1", int));
+
+        let float: f64 = rows[0].try_get(1)?;
```
tychoish, line 140: do we need to try special handling for f32s? is this a risk?

tychoish (approved):

crates/testing/tests/sqllogictests/tests.rs
vrongmeal: I can add a quick test here but pretty sure that works! We never sent incorrect data :)

scsmithr (approved):
This looks good to me.

Could we change the decimal impl to return an error instead of panicking? https://github.com/glaredb/glaredb/blob/f5bdd8c2704cd202aa16d70c7fea16f899ca5c37/crates/pgrepr/src/writer.rs#L177-L179

Issue #1838, opened 2023-09-27
Azure support for data


scsmithr:
## Description

Follow on issue for https://github.com/GlareDB/glaredb/issues/1806.

Add azure support for glaredb data.

This may also coincide with https://github.com/GlareDB/glaredb/issues/1100

PR #1937, opened 2023-10-20
feat: Azure blob storage

scsmithr:
Adds support for reading from azure storage containers.

Closes https://github.com/GlareDB/glaredb/issues/1838

# SQL

Table function support for `azure://...` paths:

```sql
SELECT * FROM csv_scan(
	'azure://glaredb-test/copy_to/with_opts.csv',
	account_name => '${AZURE_ACCOUNT}',
	access_key => '${AZURE_ACCESS_KEY}'
);
```

Create credentials:

```sql
CREATE CREDENTIALS azure_creds
	PROVIDER azure
	OPTIONS (
		account_name = '${AZURE_ACCOUNT}',
		access_key = '${AZURE_ACCESS_KEY}',
	);
```

Table function with azure creds:

```sql
select * iceberg_scan('azure://glaredb-test/iceberg/tables/lineitem_simple', azure_creds);
```

# CI

Adds the required secrets for hitting out azure account. The added SLTs are based off the existing tests we have for S3 and GCS.

# Next

- Azure data dir support (`con = glaredb.connect('azure://container/db')`)
- Panic workaround on invalid access keys


crates/sqlexec/src/planner/physical_plan/copy_to.rs
```diff
@@ -124,6 +129,35 @@ impl CopyToExec {
...
+                // By converting to a data source url, we can get the path we
+                // need.
+                //
+                // @vaibhav I'd like for us to look into switchin all object
+                // store "locations" to use the full url (with scheme) so that
+                // we can be consistent with this. It would also help with user
+                // experience since they wouldn't need to know which part of the
+                // location is the "bucket" and which is the "location" (path).
+                let source_url = DatasourceUrl::try_new(&azure_options.location)
+                    .map_err(|e| DataFusionError::External(Box::new(e)))?;
```
scsmithr, line 157: @vrongmeal I want to point this out that azure is being handled slightly differently than the other object stores (since we're going through the generic access). Instead of having separate "bucket" and "location" fields, we just have a single "location" field that has the full url to the object.

While this isn't a change a I wanted to make in this pr, I do think moving in that direction for the other stores would provide a better ux. Instead of the user needing to give us separate locations and buckets, they just throw the full url our way and we handle that directly.

scsmithr:
Add azure to `init_session_registry`

scsmithr:
> Add azure to `init_session_registry`

Never mind, I actually already did this.

universalmind303:
Kinda unrelated to the PR, but we should look at trying to upsource some of the object store utils to object store. It's kinda silly that every library I know of that uses object store is creating a lot of nearly identical utility functions.

testdata/sqllogictests_object_store/azure/validation.slt
```diff
@@ -0,0 +1,31 @@
...
+#     OPTIONS (
+#         account_name = '${AZURE_ACCOUNT}',
+#         access_key = '${AZURE_ACCESS_KEY}',
+#         location = 'azure://glaredb-test/missing_table.parquet'
+#     );
+
+# # Validation test with the wrong bucket name
+# statement error
+# CREATE EXTERNAL TABLE wrong_bucket_name
+#     FROM gcs
```
universalmind303, line 26: should this be `azure`?

tychoish (approved):
I like the tests, things seem good:
- I'm surprised that the URI scheme is `azure` but seems good.
- I feel like there could probably be more tests of failure cases (but this is a general critique of our testing and not specific to this') 
- there are a lot of notes for future work/fixes, and it feels like it might be useful to collect them somewhere other than comments, (or put some kind of `TODO(azure)` tag) though I don't feel strongly at all about this.

testdata/sqllogictests_object_store/azure/validation.slt
scsmithr: Yeah, these are all commented out until we can get proper non-panicking validation.

scsmithr:
> I like the tests, things seem good:
> 
>     * I'm surprised that the URI scheme is `azure` but seems good.

What would you expect?

>     * I feel like there could probably be more tests of failure cases (but this is a general critique of our testing and not specific to this')

Like with invalid credentials? What classes of failures would you want to see?

>     * there are a lot of notes for future work/fixes, and it feels like it might be useful to collect them somewhere other than comments, (or put some kind of `TODO(azure)` tag) though I don't feel strongly at all about this.

I agree, I'll working on opening up issues after sql server.

universalmind303:
Do we plan on eventually supporting any/all of the additional azure config options?

https://github.com/apache/arrow-rs/blob/11b2f5fecc257d97005f2393ee17777ed5d38e7c/object_store/src/azure/builder.rs#L193

while not required, It would also be nice to allow the key aliases as well

```sql
CREATE CREDENTIALS azure_creds
	PROVIDER azure
	OPTIONS (
		account_name = '${AZURE_ACCOUNT}', -- azure_storage_account_name | account_name
		account_key = '${AZURE_ACCESS_KEY}',
		-- azure_storage_account_key | azure_storage_access_key | azure_storage_master_key | master_key | account_key | access_key
		azure_container_name = '${AZURE_CONTAINER_NAME}',

	);
```

tychoish:
>What would you expect?

`abs` (for azure blob storage) or `azbs` but I guess it kind of doesn't matter, and this seems fine; it's more about following the established custom/pattern

> Like with invalid credentials? What classes of failures would you want to see?

Invalid credentials, or tables/files that don't exist, buckets that do exist but that we don't have permissions too? I guess some of these are functionally the same.

scsmithr:
> abs (for azure blob storage) or azbs but I guess it kind of doesn't matter, and this seems fine; it's more about following the established custom/pattern

I was basing it off https://docs.rs/object_store/latest/object_store/azure/struct.MicrosoftAzureBuilder.html#method.with_url.

scsmithr:
Followup items tracked in https://github.com/GlareDB/glaredb/issues/2096

scsmithr:
> Do we plan on eventually supporting any/all of the additional azure config options?
> 
> https://github.com/apache/arrow-rs/blob/11b2f5fecc257d97005f2393ee17777ed5d38e7c/object_store/src/azure/builder.rs#L193
> 
> while not required, It would also be nice to allow the key aliases as well
> 
> ```sql
> CREATE CREDENTIALS azure_creds
> 	PROVIDER azure
> 	OPTIONS (
> 		account_name = '${AZURE_ACCOUNT}', -- azure_storage_account_name | account_name
> 		account_key = '${AZURE_ACCESS_KEY}',
> 		-- azure_storage_account_key | azure_storage_access_key | azure_storage_master_key | master_key | account_key | access_key
> 		azure_container_name = '${AZURE_CONTAINER_NAME}',
> 
> 	);
> ```

I'd like to. Right now we're storing key/value pairs in the protobufs (for azure at least) so it would be easy enough for us to extend.

PR #1944, opened 2023-10-23
chore: Send `bytes_read` metrics to telemetry.

vrongmeal:


crates/datafusion_ext/src/metrics.rs
```diff
@@ -212,7 +212,7 @@ pub struct AggregatedMetrics {
     /// Total time taken for a plan to execute.
     pub elapsed_compute_ns: u64,
     /// Total bytes processed.
-    pub bytes_processed: u64,
+    pub bytes_read: u64,
```
scsmithr, line 215: While we're here, lets go ahead an rename the above gauge name too.

crates/datafusion_ext/src/metrics.rs
```diff
@@ -223,22 +223,20 @@ impl AggregatedMetrics {
...
-        };
-        self.elapsed_compute_ns += metrics.elapsed_compute().unwrap_or_default() as u64;
-        self.bytes_processed += metrics
-            .sum_by_name(BYTES_PROCESSED_GAUGE_NAME)
-            .map(|m| m.as_usize() as u64)
-            .unwrap_or_default();
+        if let Some(metrics) = plan.metrics() {
+            self.elapsed_compute_ns += metrics.elapsed_compute().unwrap_or_default() as u64;
+            self.bytes_read += metrics
+                .sum_by_name(BYTES_PROCESSED_GAUGE_NAME)
```
greyscaled, line 236: As Sean was saying, let's ujpdate the guage name

crates/datafusion_ext/src/metrics.rs
vrongmeal: Updated.

crates/datafusion_ext/src/metrics.rs
vrongmeal: Updated!

scsmithr (approved):

greyscaled (approved):

crates/datafusion_ext/src/metrics.rs
```diff
@@ -223,22 +222,20 @@ impl AggregatedMetrics {
...
-            None => return,
-        };
-        self.elapsed_compute_ns += metrics.elapsed_compute().unwrap_or_default() as u64;
-        self.bytes_processed += metrics
-            .sum_by_name(BYTES_PROCESSED_GAUGE_NAME)
-            .map(|m| m.as_usize() as u64)
-            .unwrap_or_default();
+        if let Some(metrics) = plan.metrics() {
+            self.elapsed_compute_ns += metrics.elapsed_compute().unwrap_or_default() as u64;
+            self.bytes_read += metrics
```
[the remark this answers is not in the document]
scsmithr: This is taking the recorded bytes from the gauge which is always in bytes. The wrapper around the stream internally calls into this during execution:

```rust
    /// Track metrics based on the poll result from an async stream.
    pub fn record_poll(
        &self,
        poll: Poll<Option<Result<RecordBatch>>>,
    ) -> Poll<Option<Result<RecordBatch>>> {
        if let Poll::Ready(maybe_batch) = &poll {
            match maybe_batch {
                Some(Ok(batch)) => {
                    self.bytes_processed.add(batch.get_array_memory_size());
                    self.baseline.record_output(batch.num_rows());
                }
                Some(Err(_)) => self.baseline.done(),
                None => self.baseline.done(),
            }
        }
        poll
    }
```

The `batch.get_array_memory_size()` returns the byte size of the batch which increments the gauge by that amount.

Issue #1936, opened 2023-10-20
Confusing error when argument to `csv_scan` not parsed correctly


scsmithr:
## Description

(I'm unsure how to best describe the issue I'm seeing)

We have `./testdata/csv/userdata1.csv` that we can query normally:

```
% cargo run --bin glaredb -- -q "select * from csv_scan('./testdata/csv/userdata1.csv') limit 1"
    Finished dev [unoptimized + debuginfo] target(s) in 0.52s
     Running `target/debug/glaredb -q 'select * from csv_scan('\''./testdata/csv/userdata1.csv'\'') limit 1'`
┌─────────────────────┬───────┬────────────┬───────────┬──────────────────┬────────┬─────────────┬──────────────────┬───────────┬───────────┬──────────┬────────────────┬──────────┐
│ registration_dttm   │    id │ first_name │ last_name │ email            │ gender │ ip_address  │ cc               │ country   │ birthdate │   salary │ title          │ comments │
│ ──                  │    ── │ ──         │ ──        │ ──               │ ──     │ ──          │ ──               │ ──        │ ──        │       ── │ ──             │ ──       │
│ Timestamp<ns, UTC>  │ Int64 │ Utf8       │ Utf8      │ Utf8             │ Utf8   │ Utf8        │ Utf8             │ Utf8      │ Utf8      │  Float64 │ Utf8           │ Utf8     │
╞═════════════════════╪═══════╪════════════╪═══════════╪══════════════════╪════════╪═════════════╪══════════════════╪═══════════╪═══════════╪══════════╪════════════════╪══════════╡
│ 2016-02-03T07:55:29 │     1 │ Amanda     │ Jordan    │ ajordan0@com.com │ Female │ 1.197.201.2 │ 6759521864920116 │ Indonesia │ 3/8/1971  │ 49756.53 │ Internal Audi… │ 1E+02    │
└─────────────────────┴───────┴────────────┴───────────┴──────────────────┴────────┴─────────────┴──────────────────┴───────────┴───────────┴──────────┴────────────────┴──────────┘
```

When I prefix with a `file://` scheme (still using the relative path), I get a confusing error:

```
% cargo run --bin glaredb -- -q "select * from csv_scan('file://./testdata/csv/userdata1.csv') limit 1"
    Finished dev [unoptimized + debuginfo] target(s) in 0.52s
[...]

PR #1945, opened 2023-10-23
fix: add better error messages for parsing datasource urls

### LINKED ISSUE #1897: Bad error message when selecting from non-existent file
## Description

```
> select * from './testdata/userdata1.parquet';
Error: External error: Unable to convert path "./testdata/userdata1.parquet" to URL
```

The error is confusing, and doesn't indicate that the file was not found. Tested on 0.5.1 and latest main.

The correct path is `./testdata/parquet/userdata1.parquet`, which works as expected.

universalmind303:
closes #1897
closes https://github.com/GlareDB/glaredb/issues/1936

The error messages should properly bubble up now. 

Additionally, we now canonicalize the relative paths during `DatasourceUrl::try_new`. This ensures that the path is absolute, and that it is a valid file/directory.

### DIFF
--- a/crates/datasources/src/object_store/local.rs
+++ b/crates/datasources/src/object_store/local.rs
@@ -31,7 +31,8 @@ impl ObjStoreAccess for LocalStoreAccess {
     }
 
     fn path(&self, location: &str) -> Result<ObjectStorePath> {
-        Ok(ObjectStorePath::from_filesystem_path(location)?)
+        ObjectStorePath::from_filesystem_path(location)
+            .map_err(|e| super::errors::ObjectStoreSourceError::ObjectStorePath(e))
     }
 
     /// Given relative paths and all other stuff, it's much simpler to use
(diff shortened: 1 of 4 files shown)


crates/datasources/src/object_store/local.rs
```diff
@@ -49,13 +50,13 @@ impl ObjStoreAccess for LocalStoreAccess {
...
         )?;
-
         let mut objects = Vec::new();
         for path in paths {
             let path = path?;
             let meta = path.metadata()?;
+            let loc = self.path(path.to_string_lossy().as_ref());
             let meta = ObjectMeta {
-                location: self.path(path.to_string_lossy().as_ref())?,
+                location: loc?,
```
universalmind303, line 59: revert.

tychoish, line 59:
```suggestion
                location: self.path(path.to_string_lossy().as_ref())?,
```
push the button ;)

tychoish (approved):
looks good, modulo the test failure

PR #1950, opened 2023-10-24
fix: add better error messages for http globbing

### LINKED ISSUE #1747: Return better error when using a glob with http/https
## Description

For this case:
```
select * from 'http://host.com/path/*'
```

We also do not support
```
select * from 'http://host.com/path/*.parquet'
```

There is not a good way for us to support globs with http sources, so we should instead return a better error.

universalmind303:
closes #1747

Note: this does explicitly forbid a literal `*` in a url. I don't think this should ever happen as any actual `*` in a url should be urlencoded.


scsmithr (approved):

crates/datasources/src/object_store/http.rs
```diff
@@ -65,6 +65,11 @@ impl ObjStoreAccess for HttpStoreAccess {
         store: &Arc<dyn ObjectStore>,
         pattern: &str,
     ) -> Result<Vec<ObjectMeta>> {
+        if pattern.contains('*') {
+            return Err(ObjectStoreSourceError::Static(
+                "Globbing not supported for HTTP store",
+            ));
+        }
```
vrongmeal, line 72: I don't know if we want to return an error here. What if there's a URL with `*` in it? Do we expect each URL to have %-encoded chars?

Also, there are more chars that can say it's a glob, for eg., `?`.

universalmind303: > I don't know if we want to return an error here. What if there's a URL with * in it? Do we expect each URL to have %-encoded chars?

see my initial comment. I'd assume any actual `*` would be url encoded.

universalmind303: @vrongmeal I updated the logic to better handle files with `*` in them.

tychoish (approved):

PR #1951, opened 2023-10-24
ci: update ci for slts

universalmind303:
The wall of text when running the SLT's is overkill. Makes it super hard to grok what is going on in the tests. 

This should only give the debug log if you [enable debug logging](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging).

image

--- 

It'll also use the compact logging mode if not ran in debug mode. 

Additionally, quite a few of our "info" logs seemed like they were actually "debug" logs as they didn't seem relevant for general informational purposes, but for debugging purposes only. 

--- 
### Screenshot of new logging output when running locally:
image


.github/workflows/ci.yaml
```diff
@@ -140,7 +140,12 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Public SQL Logic Tests
-        run: just slt -v 'sqllogictests/*'
+        run: |
+          if [[ "$ACTIONS_STEP_DEBUG" == "true" ]]; then
+            just slt -v 'sqllogictests/*'
+          else
+            just slt 'sqllogictests/*'
+          fi
```
tychoish, line 148: I spent a hot second trying to figure out if you could get the `slt` arg parser to check this (or check it in the `justfile` and we _could_ but it seems like a lot of worlk.

.github/workflows/ci.yaml
```diff
@@ -140,7 +140,12 @@ jobs:
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
       - name: Public SQL Logic Tests
-        run: just slt -v 'sqllogictests/*'
+        run: |
+          if [[ "$ACTIONS_STEP_DEBUG" == "true" ]]; then
```
tychoish, line 144: I think the reason I am inclined to doing this in the slt/justfile is that, this is a bashism, and I _think_ on ubuntu I would not be surprised if this ends up being dash.

.github/workflows/ci.yaml
universalmind303: I don't think we should couple the justfile to the CI & GH specific variables. 

IMO, this is a concern of the ci, not the script runner.

universalmind303: @tychoish I updated the logic to use github actions conditionals instead of bash. LMK what you think.

tychoish: Looks good!

PR #1962, opened 2023-10-26
feat: add js bindings

universalmind303:
I originally set out to automate the generation of the [builtin functions](https://docs.glaredb.com/glaredb/sq-functions/index/) in our docs, thinking the docs were written using a JS framework. **It turns out they are not.**

But anyways, this seems like a pretty cool addition nonetheless. 

Todo:

- [x] finish up JS interop 
- [x] clean up docstrings
- [ ] get ci working  -- https://github.com/GlareDB/glaredb/issues/1999
- [x] try to see if we can remove `unsafe`


universalmind303:
I'll try to work on getting the CI working in a separate PR. 

So any reviewers can just ignore the `js-polars/CI.yml` file for the time being.

Cargo.toml
```diff
@@ -9,6 +9,9 @@ edition = "2021"
 
 [profile.release]
 lto = "thin"
+opt-level = 3
+codegen-units = 1
+strip = true
```
universalmind303, line 14: reduces the binary size from `~120mb` to `~70mb`. Compile times are a bit longer though

scsmithr (approved):

js-glaredb/rustfmt.toml
```diff
@@ -0,0 +1,2 @@
+tab_spaces = 2
```
scsmithr, line 1: Could we remove this?

universalmind303: oh yeah, that was auto generated from napi.

PR #1996, opened 2023-10-27
chore: Use ld_classic when linking on mac

scsmithr:
I was having some issues with panics not showing backtraces, and instead showing the following:

```
libunwind: stepWithCompactEncoding - invalid compact unwind encoding
```

This seems to be from the latest mac update using a new linker. See https://github.com/rust-lang/rust/issues/113783

As an aside, I think might have an errant panic somewhere that's not getting bubbled up properly.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,9 @@
 [alias]
 xtask = "run --bin xtask --"
+
+[target."aarch64-apple-darwin"]
+# Get proper backtraces in mac Sonoma. Currently there's an issue with the new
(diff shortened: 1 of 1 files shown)


universalmind303 (approved):

tychoish (approved):
seems quite reasonable though I don't have a mac to test it and we also don't do a build to check this in CI (the first isn't particularly tractable, the second could be at least for some kind of basic pass)

scsmithr:
Going to convert this to draft. It might just make sense to open up an issue for this with the workaround that we can point to for mac devs.

PR #1997, opened 2023-10-27
feat: add excel reader

universalmind303:
Currently this only adds the table function `read_excel`. Will follow up in a separate PR to enable `create table` syntax. 

A lot of the credit here goes to the roapi implementation.


scsmithr (approved):

scsmithr:
We could possibly also add inferring the function from the `xlsx` extension as well.

tychoish (commented):
I edited the issue to include the work in the second PR noted in the description here, and also added a line item for adding some kind of integration/smoke test just to make sure that we don't accidentally break it.

tychoish (approved):

PR #1948, opened 2023-10-23
chore: Update delta and datafusion

scsmithr:

PR #2000, opened 2023-10-27
chore(deps): bump primary dependency group

tychoish:
Closes #1988
Closes #1984
Closes #1983
Closes #1980
Closes #1977
Closes #1975
Closes #1969
Closes #1967
Closes #1966


tychoish:
(This is everything except deltalake)

Cargo.toml
```diff
@@ -24,7 +24,8 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0.104"
 chrono = "0.4.31"
 anyhow = "1.0.72"
-
+prost = { workspace = true }
```
scsmithr, line 27:
```suggestion
```

Needs to specify the version if in the root toml.

scsmithr:
Seems like we need to bump `tonic-build = "0.10"` in the metastore and protogen tomls.

(We might be able to take that out of the metastore toml since all our grpc stuff is in protogen)

tychoish:
I've been doing the fixes on top of this in the #2001 pr, so that we can see "what's our changes to make things work," on top of "the core dependabot updates". 

My plan has been to merge #2001 first (which will merge into this, #2000) and then, merge #2000 into main once it works.

tychoish:
> Seems like we need to bump tonic-build = "0.10" in the metastore and protogen tomls.

This makes sense: should we make this a workspace dependency? 

> And that's going to start to trickle in some conflicting TryFrom impls. I've made those changes in https://github.com/GlareDB/glaredb/pull/1948/files which wil should be able to copy into this branch (from_i32 -> try_from)

oh interesting, I ran into this problem this morning, when I tried to do one single omnibus upgrade, and backed out of it.

scsmithr:
> This makes sense: should we make this a workspace dependency?

I think that'd make sense. I imagine it would work the same way as normal dependencies.

> oh interesting, I ran into this problem this morning, when I tried to do one single omnibus upgrade, and backed out of it.

Yup there's definitely some cascading changes here. I wouldn't say this is normal, but it's hitting us this go around given we have ~2 months of backed up dependency updates.

scsmithr (approved):

PR #2003, opened 2023-10-28
fix: mongodb pushdown filters

tychoish:
still in progress. 
- doesn't compile because the cursor isn't Sync
- there might be an unnecessary layer of indirection around the cursor at the moment
- retryable semantics may not be correct.
- want to gut check some of the type conversions.


crates/datasources/src/mongodb/mod.rs
```diff
@@ -271,18 +279,144 @@ impl TableProvider for MongoTableProvider {
...
+                .as_ref(),
+        );
+
+        // let cursor = match .await {
+        //     Ok(cursor) => cursor,
+        //     Err(e) => {
+        //         yield Err(DataFusionError::External(Box::new(e)));
+        //         return;
+        //     }
+        // };
```
tychoish, line 320:
```suggestion
```

crates/datasources/src/mongodb/exec.rs
```diff
@@ -14,35 +14,25 @@ use datafusion::physical_plan::{
...
 
-/// Field name in mongo for uniquely identifying a record. Some special handling
-/// needs to be done with the field when projecting.
-const ID_FIELD_NAME: &str = "_id";
-
 #[derive(Debug, Clone)]
 pub struct MongoBsonExec {
-    schema: Arc<ArrowSchema>,
-    collection: Collection<RawDocumentBuf>,
+    cursor: Arc<Cursor<RawDocumentBuf>>,
```
scsmithr, line 27: This is what's causing stuff to not be Sync.

What we can do here is wrap it in a mutex like so:

```
cursor: Mutex<Option<Cursor<RawDocumentBuf>>>
```

(This requires adding parking_lot to the cargo.toml)

Then in `execute`, we take the cursor out of the mutex:

```rust
    fn execute(
        &self,
        partition: usize,
        _context: Arc<TaskContext>,
    ) -> DatafusionResult<SendableRecordBatchStream> {
        if partition != 0 {
            return Err(DataFusionError::Execution(format!(
                "only single partition supported"
            )));
        }

        let cursor = {
            let mut cursor = self.cursor.lock();
            cursor.take().ok_or_else(|| {
                DataFusionError::Execution(format!(
                    "execution called on partition {partition} more than once"
                ))
            })?
        };

        Ok(Box::pin(DataSourceMetricsStreamAdapter::new(
            BsonStream::new(cursor, self.schema.clone(), self.limit),
            partition,
            &self.metrics,
        )))
    }
```

`execute` is only called once per partition at the very beginning of execution, and so the mutex is only getting locked once with no contention. The actual execution stream (what we return) remains lock free.

Also the `Arc` isn't needed here.

scsmithr, line 27: We'll also need to keep the schema here for the `schema` method.

crates/datasources/src/mongodb/mod.rs
```diff
@@ -271,18 +279,136 @@ impl TableProvider for MongoTableProvider {
...
-            self.collection.clone(),
-            limit,
-        )))
+        let filter = exprs_to_mdb_query(_filters).expect("could not build query");
+        let cursor = Arc::new(
+            self.collection
+                .find(Some(filter), Some(find_opts))?
+                .into()
+                .as_ref(),
+        );
```
scsmithr, line 312: We can change this to:

```
        let cursor = self
            .collection
            .find(Some(filter), Some(find_opts))
            .await
            .map_err(|e| DataFusionError::External(Box::new(e)))?;
```

crates/datasources/src/mongodb/mod.rs
```diff
@@ -271,18 +279,136 @@ impl TableProvider for MongoTableProvider {
...
+        let mut find_opts = FindOptions::default();
+        find_opts.limit = limit.map(|v| v as i64);
+        find_opts.projection = Some(proj_doc);
 
-        Ok(Arc::new(MongoBsonExec::new(
-            projected_schema,
-            self.collection.clone(),
-            limit,
-        )))
+        let filter = exprs_to_mdb_query(_filters).expect("could not build query");
```
scsmithr, line 306: We'll want to return an error here (e.g. with `map_err(|e| ...)`) to avoid panicking.

Longer term I think we'll want to indicate to the MongoBsonExec whether or not we were successful in creating the filter. If we're not, then we have to the filtering in memory.

tychoish: > Longer term I think we'll want to indicate to the MongoBsonExec whether or not we were successful in creating the filter. If we're not, then we have to the filtering in memory.
I think we'll always want to do the filtering in memory, unless we're successful at creating a filter that _perfectly_ transforms the users' query: the goal of this, I think, is to limit the number of records that data-fusion has to consider, not to obviate data-fusion entirely (for that we'd need/want to write a complete sql-to-mdb query translator, and that's a much much larger undertaking)

crates/datasources/src/mongodb/mod.rs
```diff
@@ -271,18 +279,136 @@ impl TableProvider for MongoTableProvider {
...
+        let cursor = Arc::new(
+            self.collection
+                .find(Some(filter), Some(find_opts))?
+                .into()
+                .as_ref(),
+        );
+
+        Ok(Arc::new(MongoBsonExec::new(cursor, limit)))
```
universalmind303, line 314: I don't think we should be creating a cursor in the `scan`. Instead we should push the filter down to the physical exec. That should also keep the Exec threadsafe.

crates/datasources/src/mongodb/exec.rs
tychoish: It looks like the Arc is needed, because the cursor doesn't implement clone 😞

scsmithr: We can remove the clone, it shouldn't be needed.

tychoish: I thought clone was passing when passing "by value"

crates/datasources/src/mongodb/mod.rs
tychoish: This doesn't seem to be consistent with how the code works, consistently: mysql and postgres operate the way you expect; bigquery and snowflake operate (roughly, to my reading) as here. 

It seems from what Sean said above not only does Exec not need to be thread safe, but that those methods aren't ever called more than once.

crates/datasources/src/mongodb/exec.rs
scsmithr: Clone just lets you make a copy of it via `clone()`. Everything is pass by value (objects and references).

universalmind303:
@tychoish what happens if a filter can't be pushed down? does it fallback to in memory, or does it error out?

crates/datasources/src/mongodb/mod.rs
scsmithr: Actually for this let's do this:

```rust
let filter = match exprs_to_mdb_query(filters).ok()
```

Then down below use `filter` instead of `Some(filter)`. Since we're using `TableProviderFilterPushDown::Inexact`, there's going to be a filter node above this anyways.

tychoish:
> @tychoish what happens if a filter can't be pushed down? does it fallback to in memory, or does it error out?

It's my belief that this ends up being pretty course, but I'd love for everyone to check my logic/understanding:
- we go through the expressions provided and attempt to convert them into their equivalent mongodb query fragments. There's no magic, no aggregation, no synthetic expressions, nothing, if we support the operator, we add the clause to the query, and if not we continue.
- we're always going to run the datafusion query over the contents of the stream and I think I did nothing that would remove that current behavior: the goal of this PR is to reduce the total size of the payload coming off of MongoDB, which should drive a significant performance improvement, and the rest of the query (which could include transforms/etc.) is probably best run locally anyway (though we can easily extend and push down other operations.

The questions/concerns are basically:
- is DF ever going to give this data source a a query that includes fields/columns with names that don't map directly to what MongoDB knows about? if we got an expression for a column name that didn't map exactly, we could add an equality match for something that doesn't exist, and that would filter all results out?
- I could imagine a query operator that operates on a synthetic/aggregated field that this code would add, but would find no results from, but I'm unsure how decomposed the expression we get is going to be. 
- eventually, perhaps, we can/should write a thing that generates MDBQL from DFs AST: that feels like a huge undertaking right now. (that'd also be kind of a killer product on its own?)

tychoish:
> I'm not sure what's being asked here. The only expressions we get provided in the filters are expressions directly related to columns that are found in the schema.

Fair enough, you did a great job of answering it. My fears are allayed.

> The currently implementation attempts to convert all filters to the equivalent mongo expressions, and errors otherwise. If we go with my above suggestion, we can instead not return an error, and just None when doing the actual mongo query as there's going to be filter node above this node anyways.

Wild! That wasn't my intention at all. 

I will definitely apply your suggestion, but actually maybe we want to do something even more permissive: attempt to convert what we can, skip everything else, and then let the other filter take over.

tychoish:
I've added, what I hope are "continue on error semantics" to the query
building along with a debug trace of these errors. I think it'd be
good, eventually to have metrics for this (and in general all pushdown
failures) just to be able to see where the optimization. possibilities
are.

crates/datasources/src/mongodb/mod.rs
tychoish: I think I have something that resembles your intent here; let me know :D

scsmithr (approved):

Issue #2005, opened 2023-10-28
Panics in pretty format tests


scsmithr:
## Description

```
---- pretty::tests::simple_no_batches stdout ----
thread 'pretty::tests::simple_no_batches' panicked at crates/arrow_util/src/pretty.rs:234:31:
attempt to subtract with overflow

---- pretty::tests::multiple_small_batches stdout ----
thread 'pretty::tests::multiple_small_batches' panicked at crates/arrow_util/src/pretty.rs:234:31:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- pretty::tests::multiple_batches_with_slicing stdout ----
thread 'pretty::tests::multiple_batches_with_slicing' panicked at crates/arrow_util/src/pretty.rs:234:31:
attempt to subtract with overflow

---- pretty::tests::large_batch_with_max_rows stdout ----
thread 'pretty::tests::large_batch_with_max_rows' panicked at crates/arrow_util/src/pretty.rs:234:31:
attempt to subtract with overflow

---- pretty::tests::multiple_small_batches_with_max_rows stdout ----
thread 'pretty::tests::multiple_small_batches_with_max_rows' panicked at crates/arrow_util/src/pretty.rs:234:31:
attempt to subtract with overflow

---- pretty::tests::simple_single_batch stdout ----
thread 'pretty::tests::simple_single_batch' panicked at crates/arrow_util/src/pretty.rs:234:31:
attempt to subtract with overflow


failures:
    pretty::tests::large_batch_with_max_rows
    pretty::tests::multiple_batches_with_slicing
    pretty::tests::multiple_small_batches
    pretty::tests::multiple_small_batches_with_max_rows
    pretty::tests::simple_no_batches
    pretty::tests::simple_single_batch
```

I'm not sure if this is from a dep update. Also doesn't seem to be failing on ci

PR #2009, opened 2023-10-28
fix: Remove terminal width detection from pretty table

scsmithr:
And instead alway require that it's passed in. Removes some magic and makes things a bit more reproducible.

Fixes https://github.com/GlareDB/glaredb/issues/2005 (terminal width was getting reported as zero inside eshell)


scsmithr:
Merging since this hurts my dev workflow right now

PR #2017, opened 2023-10-30
feat(tls): make tls mode configurable

tychoish:
configuration on top of #2016

### DIFF
--- a/crates/glaredb/src/args/mod.rs
+++ b/crates/glaredb/src/args/mod.rs
@@ -4,6 +4,8 @@ use clap::{Parser, ValueEnum};
 use std::fmt::Write as _;
 use std::path::PathBuf;
 use url::Url;
+
+use crate::rpc_proxy::TLSMode;
 pub mod local;
 pub mod server;
 
@@ -51,13 +53,11 @@ pub struct RpcProxyArgs {
     #[clap(long)]
     pub cloud_auth_code: String,
 
-    /// Disable RPC TLS
+    /// Mode for TLS Validation (required, optional, disabled)
     ///
     /// (Internal)
-    ///
-    /// Note: in the future, this will be 'on' by default
-    #[clap(long, default_value="true", action = clap::ArgAction::Set, hide = true)]
-    pub disable_tls: bool,
+    #[clap(long, value_enum, hide = true)]
+    pub tls_mode: TLSMode,
 }
 
 #[derive(Parser)]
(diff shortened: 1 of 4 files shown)


crates/glaredb/src/args/mod.rs
```diff
@@ -4,6 +4,8 @@ use clap::{Parser, ValueEnum};
 use std::fmt::Write as _;
 use std::path::PathBuf;
 use url::Url;
+
+use crate::rpc_proxy::TLSMode;
```
greyscaled, line 8: (I'm a novice Rustacean): I'm wondering if this is needed since it's all in the same crate?

greyscaled: Ah, I see, it is needed when in parent dir

crates/glaredb/src/rpc_proxy.rs
```diff
@@ -16,6 +17,17 @@ pub struct RpcProxy {
     handler: RpcProxyHandler<CloudAuthenticator>,
 }
 
+#[derive(Clone, Debug, ValueEnum, Default)]
+#[clap(rename_all = "lower")]
+pub enum TLSMode {
+    Required,
+    Optional,
+
+    /// Note: in the future, this will be 'required' by default
```
greyscaled, line 26:
```suggestion
    /// Note: in the future, this will be 'Required' by default
```

just to match value above, maybe? Or not...doesn't matter

greyscaled (approved):
This looks great!

greyscaled (commented):
I think we need to touch this logic

https://github.com/GlareDB/glaredb/blob/bba05df3d155b8944d6d3074b4f97dd30a9d14b4/crates/sqlexec/src/remote/client.rs#L158

```
Error: Remote session error: failed to initialize remote session: status: Unknown, message: "h2 protocol error: http2 error: connection error detected: frame with invalid size", details: [], metadata: MetadataMap { headers: {} }
```

crates/glaredb/src/args/mod.rs
tychoish: symbols are file scoped (because translation units still matter apparently), but same.

crates/sqlexec/src/remote/client.rs
```diff
@@ -151,13 +151,9 @@ impl RemoteClient {
...
-        if !disable_tls {
-            info!("set rpc destination scheme to https");
-
-            dst.dst
-                .set_scheme("https")
-                .expect("failed to upgrade scheme from http to https");
-        }
+        dst.dst
+            .set_scheme("https")
+            .expect("failed to upgrade scheme from http to https");
```
greyscaled, line 156: Review: If this works, I think we don't need `dst` to be `mut` anymore and can do this higher up.

greyscaled (approved):

PR #2022, opened 2023-10-31
feat: DESCRIBE table

universalmind303:
still todo:

- [x] `describe` over RPC


crates/sqlexec/src/context/local.rs
```diff
@@ -84,7 +84,9 @@ impl LocalSessionContext {
         conf = conf
             .with_extension(Arc::new(catalog_mutator))
             .with_extension(Arc::new(native_tables.clone()))
-            .with_extension(Arc::new(TempCatalog::default()));
+            .with_extension(Arc::new(TempCatalog::default()))
+            .with_extension(Arc::new(catalog.clone()));
```
universalmind303, line 88: @scsmithr I know we originally removed the `catalog` from the extension as we only needed the mutator for the DDL's, but now we need the catalog again. 

I could move the logic that uses the catalog into the `session_planner`, but it seemed more appropriate to have it in the `DescribeTableExec`. 

any thoughts on this approach?

crates/sqlexec/src/remote/planner.rs
```diff
@@ -189,6 +190,21 @@ impl ExtensionPlanner for DDLExtensionPlanner {
                     or_replace: lp.or_replace,
                 })))
             }
+            ExtensionType::DescribeTable => {
+                let lp = require_downcast_lp::<DescribeTable>(node);
+                let runtime = if lp.temp || lp.builtin {
+                    RuntimePreference::Local
+                } else {
+                    RuntimePreference::Remote
+                };
```
universalmind303, line 199: builtin and temp tables can be resolved locally. Native tables need to be resolved on remote.

crates/sqlexec/src/context/local.rs
universalmind303: We could also split this up into multiple execs `[DescribeNativeTableExec, DescribeBuiltinTableExec, DescribeTempTableExec]`. 

This would make it so the execs don't need the full catalog, but it seemed simpler to just pass in the catalog through the extensions. If there's a good reason to not have the catalog on the extension, i can refactor this into multiple execs.

testdata/sqllogictests/describe.slt
```diff
@@ -0,0 +1,31 @@
+# works with builtin tables
+
+query III
+describe glare_catalog.ssh_keys;
+----
+ssh_tunnel_oid UInt32 f
+ssh_tunnel_name Utf8 f
+public_key Utf8 f
```
scsmithr, line 8:
```suggestion
query III rowsort
describe glare_catalog.ssh_keys;
----
ssh_tunnel_oid UInt32 f
ssh_tunnel_name Utf8 f
public_key Utf8 f
```

For consistent ordering.

testdata/sqllogictests/describe.slt
```diff
@@ -0,0 +1,31 @@
...
+
+# and native tables
+statement ok
+create table t1 (a int, b int);
+
+query III
+describe t1;
+----
+a Int32 t
+b Int32 t
```
scsmithr, line 19:
```suggestion
query III rowsort
describe t1;
----
a Int32 t
b Int32 t
```

testdata/sqllogictests/describe.slt
```diff
@@ -0,0 +1,31 @@
...
+# and temp tables 
+
+statement ok
+create temp table temp1 (a int, b int);
+
+query III
+describe temp1;
+----
+a Int32 t
+b Int32 t
```
scsmithr, line 31:
```suggestion
query III rowsort
describe temp1;
----
a Int32 t
b Int32 t
```

crates/sqlexec/src/planner/physical_plan/describe_table.rs
```diff
@@ -0,0 +1,211 @@
...
+    let resolver = EntryResolver {
+        catalog: &catalog,
+        temp_objects: temp_catalog,
+        schema_search_path,
+    };
+
+    let ent = resolver
+        .resolve_entry_from_reference(tbl_ref.into())
+        .and_then(ResolvedEntry::try_into_table_entry)
+        .expect("entry should be checked during planning");
```
scsmithr, line 172: I'd return an error here since we don't have a guarantee that the table hasn't been dropped between planning and execution.

crates/sqlexec/src/planner/physical_plan/describe_table.rs
```diff
@@ -0,0 +1,211 @@
...
+) -> Result<RecordBatch> {
+    let metastore_client = mutator.get_metastore_client();
+    let mut catalog = catalog.as_ref().clone();
+    let schema_search_path = vars.implicit_search_path();
+    let temp_catalog = temp_catalog.unwrap_or_default();
+
+    catalog
+        .maybe_refresh_state(metastore_client, false)
+        .await
+        .unwrap();
```
scsmithr, line 161: Return error -- there might be a network flake in the metastore connection, ideally that doesn't cause a panic.

crates/sqlexec/src/context/local.rs
scsmithr: What's stopping us from putting a `TableEntry` on the Describe exec instead of the `OwnedFullObjectReference`?

We're already getting table entry here: https://github.com/glaredb/glaredb/blob/b0e3334fb23b86e90a14908ae945a02a2f6015bc/crates/sqlexec/src/planner/session_planner.rs#L668-L670

And so we can just stick that on the describe plans and reference that directly in the exec without needing to reference the catalog.

We have a couple other execs that already use `TableEntry`, e.g. updates:
- https://github.com/glaredb/glaredb/blob/b0e3334fb23b86e90a14908ae945a02a2f6015bc/crates/protogen/src/sqlexec/physical_plan.rs#L261-L269
- https://github.com/glaredb/glaredb/blob/b0e3334fb23b86e90a14908ae945a02a2f6015bc/crates/sqlexec/src/planner/physical_plan/update.rs#L20-L25

universalmind303: oh good idea. I didn't notice `TableEntry` was serializable.

scsmithr (approved):

PR #2026, opened 2023-11-01
feat: Add "read-only" mode

### LINKED ISSUE #2010: Create read-only mode
(add catalog "allow-writes" flag and enforce)

vrongmeal:
```
> create external table local_table from local ( location 'testdata/csv/empty_col.csv' );
Table created
> select * from local_table;
┌───────┬──────┬───────┐
│       │ col1 │ col2  │
│    ── │ ──   │ ──    │
│ Int64 │ Utf8 │ Utf8  │
╞═══════╪══════╪═══════╡
│     0 │ a    │ hello │
│     1 │ b    │ world │
└───────┴──────┴───────┘
> insert into local_table values (2, 'c', 'glaredb');
Error: Cannot write to read-only table: local_table
```

### TODO

- [ ] #2027

Fixes #2010


crates/datafusion_ext/src/runtime/table_provider.rs
```diff
@@ -21,13 +21,15 @@ use super::runtime_group::RuntimeGroupExec;
 pub struct RuntimeAwareTableProvider {
     pub preference: RuntimePreference,
     pub provider: Arc<dyn TableProvider>,
+    pub allow_writes: Option<bool>,
```
tychoish, line 24: this should definitely be an enum for "allowed_operations" (read, write_ddl, write_dml, write_all) or something like that. 

(there are potentially a few states for this, and the difference between false and not-specified-optional is confusing)

tychoish (commented):
This looks good, but I think I'd also:
- [x] enum
- [ ] adding a couple of tests seems important

crates/datafusion_ext/src/runtime/table_provider.rs
vrongmeal: > this should definitely be an enum for "allowed_operations" (read, write_ddl, write_dml, write_all) or something like that.

Agreed. Will update this and add tests.

testdata/sqllogictests/allowed_operations.slt
```diff
@@ -0,0 +1,14 @@
+statement ok
+CREATE EXTERNAL DATABASE debug_db FROM debug;
```
tychoish, line 2: this is cool, and I definitely 100% got to this point and don't fully understand how this works or would catch a bug if we accidentally regressed?

tychoish (commented):
this looks really cool and more or less like what I'd expect, but I definitely would like other folks to look at this too

testdata/sqllogictests/allowed_operations.slt
vrongmeal: We'll catch a bug here. Trying to insert should always check for permissions first. Moreover, we have an error message in the check that should force us to add different tests in the future if something regressed.

tychoish (approved):

Issue #2033, opened 2023-11-02
tilde path doesn't work when selecting from path


universalmind303:
## Description

```sql
select * from '~/Downloads/userdata1.parquet';
Error: External error: External error: No such file or directory (os error 2)
```

the full path however does work

```sql
select * from '/Users/me/Downloads/userdata1.parquet';
```

PR #2036, opened 2023-11-02
fix: allow tilde paths

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2033

### DIFF
--- a/crates/ioutil/src/lib.rs
+++ b/crates/ioutil/src/lib.rs
@@ -1,3 +1,15 @@
 //! IO and formatting related utilies.
+
+use std::path::{Path, PathBuf};
 pub mod fmt;
 pub mod write;
+
+pub fn resolve_path(path: &Path) -> std::io::Result<PathBuf> {
+    if path.starts_with("~") {
+        if let Some(homedir) = home::home_dir() {
+            return Ok(homedir.join(path.strip_prefix("~").unwrap()));
+        }
+    }
+
+    path.canonicalize().map(|p| p.to_path_buf())
+}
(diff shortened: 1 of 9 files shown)


crates/ioutil/src/lib.rs
```diff
@@ -1,3 +1,15 @@
 //! IO and formatting related utilies.
+
+use std::path::{Path, PathBuf};
 pub mod fmt;
 pub mod write;
+
+pub fn resolve_path(path: &Path) -> std::io::Result<PathBuf> {
+    if path.starts_with("~") {
```
scsmithr, line 8:
```suggestion
    if path.starts_with("~/") {
```

Maybe? Windows temp files are prefixed with tilde...

scsmithr (approved):

crates/ioutil/src/lib.rs
tychoish: I would hope that path.join on line 10 would avoid `//` on join (otherwise it would just be concat?) so I think we'd be safe to not do this.

tychoish (approved):
lgtm modulo adding a test for `resolve_path()`

vrongmeal (approved):

vrongmeal:
@universalmind303 can we add SLTs? We can use the `HOME` environment variable or set the test variable using a `Hook`.

```
statement ok
COPY (SELECT 101) TO '${HOME}/data.csv';

query I
SELECT * FROM '~/data.csv';
----
101
```

This reminds me, this doesn't work with `COPY .. TO ..`:

```
> copy (select 1) to '~/Desktop/data.csv';
Error: External error: IO error: No such file or directory (os error 2)
> copy (select 1) to '/Users/vrongmeal/Desktop/data.csv';
Copy success
```

PR #2057, opened 2023-11-06
feat: `ALTER [DATABASE/TABLE] .. SET ACCESS_MODE TO ..`

vrongmeal:
```
> create external database pg from postgres (
::: host = 'localhost',
::: port = 5433,
::: user = 'glaredb',
::: password = 'password',
::: database = 'glaredb_test' );
Database created
> select * from pg.public.test_insert where a = 111;
┌───────┬──────┐
│     a │ b    │
│    ── │ ──   │
│ Int32 │ Utf8 │
╞═══════╪══════╡
└───────┴──────┘
> insert into pg.public.test_insert values (111, 'glaredb');
Error: Not allowed to write into the object: pg.public.test_insert
> select * from pg.public.test_insert where a = 111;
┌───────┬──────┐
│     a │ b    │
│    ── │ ──   │
│ Int32 │ Utf8 │
╞═══════╪══════╡
└───────┴──────┘
> alter database pg set access_mode to read_write;
Database altered
> insert into pg.public.test_insert values (111, 'glaredb');
Inserted 1 row
> select * from pg.public.test_insert where a = 111;
┌───────┬─────────┐
│     a │ b       │
│    ── │ ──      │
│ Int32 │ Utf8    │
╞═══════╪═════════╡
│   111 │ glaredb │
└───────┴─────────┘
> alter database pg set access_mode to read_only;
Database altered
> insert into pg.public.test_insert values (1234, 'glaredb');
Error: Not allowed to write into the object: pg.public.test_insert
```


crates/sqlexec/src/planner/physical_plan/alter_database_rename.rs
universalmind303, line 1: rename file to `alter_database`

crates/sqlexec/src/parser.rs
```diff
@@ -127,17 +128,37 @@ impl fmt::Display for DropDatabaseStmt {
...
+            }
+            Self::GrantAccess { permission } => {
+                write!(f, "GRANT ACCESS TO {permission}")
+            }
+            Self::RevokeAccess { permission } => {
+                write!(f, "REVOKE ACCESS TO {permission}")
+            }
+        }
+    }
+}
```
scsmithr, line 151: Was there an existing system that you based the `REVOKE`/`GRANT` format on? Typically when I see these, I associated permissions for a user and not the database object itself. The way I see this, we're setting a property on an object that applies to everybody accessing that object.

When setting the property of an object, I'd expect to see something more along the lines of:

```
ALTER DATABASE <database> SET WRITE_MODE READONLY;
ALTER DATABASE <database> SET WRITE_MODE WRITE_DML;
```

(or similar)

(also open to changing my mind)

tychoish: I like `SET` / `UNSET` as `GRANT` / `REVOKE`.

What is the default? Is _that_ settable somewhere on the deployment?

crates/sqlexec/src/planner/physical_plan/alter_database_rename.rs
vrongmeal: Done :D

crates/sqlexec/src/parser.rs
vrongmeal: Yeah, I guess `SET`/`UNSET` makes more sense in this context.

> I like SET / UNSET as GRANT / REVOKE.

@tychoish A bit unclear, do you prefer `SET`/`UNSET` or `GRANT`/`REVOKE`?

> What is the default? Is that settable somewhere on the deployment?

The default for native and temporary tables is both read and write permissions. The default for external sources is read only.

crates/protogen/proto/metastore/service.proto
```diff
@@ -94,15 +94,55 @@ message CreateExternalDatabase {
   // next: 5
 }
 
-message AlterTableRename {
+message AlterTableOperationRename {
+  string new_name = 1;
+}
+
+message AlterTableOperationGrantAccess {
+  uint64 permission = 1;
```
vrongmeal, line 102: Can move the permission enum inside proto to make these typesafe as well.

crates/sqlexec/src/parser.rs
tychoish: Prefer `SET` / `UNSET`

We should make sure to document these defaults.

scsmithr (approved):

tychoish (approved):

scsmithr:
Going to go ahead and merge this to avoid stacking up potential conflicts.

PR #2060, opened 2023-11-07
test: Add s3 tests for parquet and glob tests for diff schema

vrongmeal:


scsmithr (approved):
This looks good to me. Can we open up an issue to maybe explore allowing different schemas? A use case here might be schema evolution, e.g. a new column gets added to parquet files that are generated at a later point in time.

And the result could be:

```
---------------
|  a   |  b   |
---------------
|  1   | NULL |
| NULL |  2   |
---------------
```

tychoish (approved):

scsmithr:
Opened issue for the different schema thing: https://github.com/GlareDB/glaredb/issues/2061

PR #2088, opened 2023-11-13
fix(bindings-node): panic on empty result set

### LINKED ISSUE #2055: node bindings: How do I get columns values when converting to record batches?
## Description

I'm running the following:

```javascript
import glaredb from "@glaredb/glaredb"

let con = await glaredb.connect();
let res = await con.sql(`
  SELECT *
  FROM 'https://github.com/GlareDB/glaredb/raw/main/testdata/parquet/userdata1.parquet'
  LIMIT 10
`);

const s = await res.recordBatches()

console.log(s)

console.log(s[0].columns[2])
```

The first console log prints out:

```
[
  {
    schema: { fields: [Array], metadata: {} },
    columns: [
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object], [Object],
      [Object]
    ],
    rowCount: 10
  }
]
```

And the second prints:

```
{
  datatype: 'Utf8',
  offsets: [
     0, 0,  0, 0,  6, 0,  0, 0, 12, 0,
     0, 0, 18, 0,  0, 0, 24, 0,  0, 0,
    30, 0,  0, 0, 37, 0,  0, 0, 43, 0,
     0, 0, 48, 0,  0, 0, 52, 0,  0, 0,
    57, 0,  0, 0
  ],
  values: [
     65, 109,  97, 110, 100,  97,  65, 108,  98, 101,
    114, 116,  69, 118, 101, 108, 121, 110,  68, 101,
    110, 105, 115, 101,  67,  97, 114, 108, 111, 115,
     75,  97, 116, 104, 114, 121, 110,  83,  97, 109,
    117, 101, 108,  72,  97, 114, 114, 121,  74, 111,
    115, 101,  69, 109, 105, 108, 121
  ]
}
```

What am I supposed to do to get the actual values that can be displayed?

### LINKED ISSUE #2056: node bindings: Panic when converting empty query results to arrow
## Description

```javascript
import glaredb from "@glaredb/glaredb"

let con = await glaredb.connect();
let res = await con.sql(`
  SELECT *
  FROM 'https://github.com/GlareDB/glaredb/raw/main/testdata/parquet/userdata1.parquet'
  WHERE false
`);

const s = await res.toArrow()
```

Results in:

```
thread 'tokio-runtime-worker' panicked at bindings/nodejs/src/execution_result.rs:30:49:
called `Option::unwrap()` on a `None` value
```

universalmind303:
closes #2056 
closes #2055

re #2055:
we just temporarily disable this function until we can properly support it.


bindings/nodejs/src/execution_result.rs
```diff
@@ -27,7 +26,11 @@ impl JsExecutionResult {
     pub(crate) async fn to_arrow_inner(&mut self) -> napi::Result<Vec<u8>> {
         let res = match &mut self.0 {
             ExecutionResult::Query { stream, .. } => {
-                let batch = stream.next().await.unwrap().map_err(JsGlareDbError::from)?;
+                let batch = stream.next().await;
+                if batch.is_none() {
+                    return Ok(vec![]);
+                }
+                let batch = batch.unwrap().map_err(JsGlareDbError::from)?;
```
scsmithr, line 33: There's no need to check that we have at least one batch, we should just be iterating over the stream and writing all batches that we get to the writer.

The schema can be gotten by calling `stream.schema()`, we don't need to the first batch.

universalmind303: For this operation, we always expect it to be a single batch. Since we are writing it to arrow file format, it always needs to be within a single batch anyways.

universalmind303: Now that I think about it.. Until we have native handling of recordbatches, we should probably just coalesce all plans to a single record batch.

universalmind303: Just realized what you were suggesting. Ignore previous comment.

scsmithr: > For this operation, we always expect it to be a single batch.

This isn't the case. The results of a query very much could result in a stream of multiple batches. All our output operations (psql, python) should be handling the case where there's multiple record batches on the stream.

> Since we are writing it to arrow file format, it always needs to be within a single batch anyways.

The IPC format can handle multiple batches (by just calling `write` more than once). And the arrow js side should be able to load a table backed by multiple batches (if it doesn't that's a pretty significant bug on their end since IPC is very much expected to be storing multiple batches).

universalmind303: yeah sorry, I originally misunderstood your comment. Just pushed up some changes that should properly handle the conversion.

scsmithr (approved):

bindings/nodejs/src/logical_plan.rs
```diff
@@ -46,10 +46,11 @@ impl JsLogicalPlan {
...
 
-    #[napi(catch_unwind)]
-    pub async fn record_batches(&self) -> napi::Result<Vec<crate::record_batch::JsRecordBatch>> {
-        self.execute_inner().await?.record_batches().await
-    }
+    // TEMPORARILY DISABLED -- we need to figure out how to convert the raw record batches to JS objects.
+    // #[napi(catch_unwind)]
+    // pub async fn record_batches(&self) -> napi::Result<Vec<crate::record_batch::JsRecordBatch>> {
+    //     self.execute_inner().await?.record_batches().await
+    // }
```
tychoish, line 53: open ticket?

is it better to make this private and panic rather than carry around commented out code?

tychoish (approved):
I don't feel great about commenting out code rather than putting in a panic (and/or) making things private (just so that they bitrot slower?

bindings/nodejs/src/logical_plan.rs
universalmind303: I don't think it really warrants an issue yet. There hasn't been any requests for this specific feature, and knowing it is going to be quite challenging, maybe it's best to just remove all of this for now?

tychoish (approved):

PR #2103, opened 2023-11-15
chore: Send metrics for hybrid exec queries (rpc)

vrongmeal:
We're getting results but there are a couple of issues:

1. Missing data:

   ```jsonc
   {
     "database_id": "00000000-0000-0000-0000-000000000000", // This is OK
     "connection_id": "00000000-0000-0000-0000-000000000000", // Missing
     "query_text": "<empty>", // Missing
     "telemetry_tag": "unknown", // Missing
     "execution_status": "success", // OK
     "error_message": null, // OK
     "elapsed_compute_ns": 9758096, // OK
     "output_rows": 5, // OK
     "bytes_read": 15466360, // OK
     "bytes_written": 0 // OK
   }
   ```

2. We're aggregating metrics on each `Poll::Ready(_)`, which might be
   inefficient (though we need definitive benchmarks to prove this).

Apart from the above-mentioned points, this should be good to merge in.


tychoish:
nice!

universalmind303:
So if i understand correctly, If telemetry is enabled _(which should only be the case for cloud)_, It will send these metrics to our telemetry server (segment)?

crates/datafusion_ext/src/metrics.rs
```diff
@@ -287,6 +301,79 @@ impl RecordBatchStream for BoxedStreamAdapater {
...
+                .bytes_written
+                .as_ref()
+                .expect("AggregateMetricsStreamAdapter should have bytes_written gauge")
+                .set(agg_metrics.bytes_written.unwrap_or_default() as usize);
+
+            match &maybe_batch {
+                Some(Ok(batch)) => self.metrics.baseline.record_output(batch.num_rows()),
+                _ => self.metrics.baseline.done(),
+            };
+        }
```
scsmithr, line 365: From what I'm seeing, we seem to be setting `bytes_written` and `bytes_read` on every batch, and _overwriting_ the existing values.

Shouldn't we only be doing this at the end? As in match on `Poll::Ready(None)` and nothing else? This would mean we're at the end of the stream, and thus only walk the execution plan once.

vrongmeal: > Shouldn't we only be doing this at the end? As in match on Poll::Ready(None) and nothing else? This would mean we're at the end of the stream, and thus only walk the execution plan once.

This won't work in cases where, for example, we have a `LimitExec` which stops execution before the stream completes.

vrongmeal:
> So if i understand correctly, If telemetry is enabled (which should only be the case for cloud), It will send these metrics to our telemetry server (segment?) on every query?

Yeah!

crates/datafusion_ext/src/metrics.rs
scsmithr: I'm assuming in that case we drop the stream? Since computing the aggregates and emitting telemetry don't block, could we implement `Drop` for this?

vrongmeal: > I'm assuming in that case we drop the stream? Since computing the aggregates and emitting telemetry don't block, could we implement Drop for this?

I had my doubts, and yeah, this doesn't work! We can't control when the stream is dropped. It's highly dependent on DF code and relying on dropping the stream *before* getting metrics is quite iffy!

tychoish: > From what I'm seeing, we seem to be setting `bytes_written` and `bytes_read` on every batch, and _overwriting_ the existing values.

Is overwriting a problem, in and of itself? he operation may not be efficient, but it's not incorrect? (we're not loosing data, the read/write counters are still correct?)

> Shouldn't we only be doing this at the end? As in match on `Poll::Ready(None)` and nothing else? This would mean we're at the end of the stream, and thus only walk the execution plan once.

This would certainly be more efficient, but relative to the other things going on here (e.g. sending the batch back to the client) even if it's wickedly slow (1ms? 2?) it's still a drop in the bucket compared to the transit time. (And this is isolated exclusively to hybrid exec, because the Adapter only wraps these executions.)

tychoish:
> So if i understand correctly, If telemetry is enabled _(which should only be the case for cloud)_, It will send these metrics to our telemetry server (segment?) on every query?

Yes. To be clear, we already send these metrics for complete operations (e.g. pg protocol), this is just adding these metrics in for hybrid execution. (these metrics give us what limited observability we have __and__ support the billing pipeline. Eventually we should probably find a way to be able to ingest this data directly into a glaredb source, but need a more reliable way to do streaming ingestion.

crates/datafusion_ext/src/metrics.rs
tychoish: My assumption, is that the two things that take measurable time here is traversing a tree that easily fits in memory, and mutex contention:
- I'd think that there's not a lot of chance for mutex contention. 
- the tree traversal seems like it could take some time particularly in pathological cases, but most of the time it's not going to be a problem, and I think even in the trivial case, the percentage of time spent doing this (relative to query execution is low.)
- if we think this is potentially a problem for user-observable latency, we could: 
  - record things lazily. 
  - store a single future to collect this information, overwriting the future on each batch, and only resolving it (at the end? on destruction?)
  - make sure it runs _after_ we start writing the batch to the socket. (maybe this is how this works?)

tychoish (commented):
@scsmithr could you take a look at this?

crates/datafusion_ext/src/metrics.rs
scsmithr: > Is overwriting a problem, in and of itself? he operation may not be efficient, but it's not incorrect? (we're not loosing data, the read/write counters are still correct?)

It's not, I just misread the flow thinking we'll be getting the metrics just from the last batch read, and not the everything up to that point. This logic should be fine as is.

> This would certainly be more efficient, but relative to the other things going on here (e.g. sending the batch back to the client) even if it's wickedly slow (1ms? 2?) it's still a drop in the bucket compared to the transit time. (And this is isolated exclusively to hybrid exec, because the Adapter only wraps these executions.)

I agree, I'm not worried about this being a performance concern for us. If anything, we'll want to measure (as well as add tests for the expected metrics values so we can make changes here as necessary).

> I'd think that there's not a lot of chance for mutex contention.

At this point, mutexes/atomics will only accessed on a single thread.

> the tree traversal seems like it could take some time particularly in pathological cases, but most of the time it's not going to be a problem, and I think even in the trivial case, the percentage of time spent doing this (relative to query execution is low.)

Agreed.

> if we think this is potentially a problem for user-observable latency, we could:

I'd want to think about this a bit more. As the code is right now, I don't see this having an observable impact, and has the lowest chance of accidentally not recording a metric. I think we can get this in now, add tests, and then optimize later if we see a need for it.

scsmithr (approved):

crates/datafusion_ext/src/metrics.rs
tychoish: > the lowest chance of accidentally not recording a metric

I'm inclined to not worry about this error case: I think it's better (and preferable) to have a system that has a failure mode of undercounting, which would happen in situations where we have an error.

In the future, I imagine that we'll have to to do something clever/extra here if/when we have retriable errors in distributed execution.

tychoish (approved):

PR #2120, opened 2023-11-16
feat: add lance as external table

universalmind303:


crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -634,6 +635,21 @@ impl<'a> SessionPlanner<'a> {
...
                 })
             }
+            TableOptions::LANCE => {
+                let location: String = m.remove_required("location")?;
+                let mut storage_options = StorageOptions::try_from(m)?;
+                if let Some(creds) = creds_options {
+                    storage_options_with_credentials(&mut storage_options, creds);
+                }
+                // Validate that the table exists.
+                let _table = scan_lance_table(&location, storage_options.clone()).await?;
```
scsmithr, line 645: As an aside, I think `open_lance_table` might be a better name for this function since it's not scanning anything upfront, and and instead is just returning a `TableProvider` implementation (Dataset).

scsmithr (approved):

PR #2138, opened 2023-11-20
chore(deps): Update datafusion, delta, pyo3

scsmithr:


crates/datafusion_ext/src/transform.rs
scsmithr, line 1: `transform_up_mut` with `FnMut` added to datafusion

crates/sqlbuiltins/src/functions/lance.rs
```diff
@@ -38,10 +38,11 @@ impl TableFunc for LanceScan {
...
     ) -> Result<Arc<dyn TableProvider>> {
         let (source_url, storage_options) = table_location_and_opts(ctx, args, &mut opts)?;
-        let dataset = scan_lance_table(&source_url.to_string(), storage_options)
+        let _dataset = scan_lance_table(&source_url.to_string(), storage_options)
             .await
             .map_err(|e| ExtensionError::Access(Box::new(e)))?;
 
-        Ok(Arc::new(dataset))
+        unimplemented!()
+        // Ok(Arc::new(dataset))
```
scsmithr, line 46: lance needs updated

crates/sqlexec/src/dispatch/external.rs
```diff
@@ -488,8 +488,9 @@ impl<'a> ExternalDispatcher<'a> {
                 storage_options,
                 ..
             }) => {
-                let dataset = scan_lance_table(location, storage_options.clone()).await?;
-                Ok(Arc::new(dataset))
+                let _dataset = scan_lance_table(location, storage_options.clone()).await?;
+                unimplemented!()
+                // Ok(Arc::new(dataset))
```
scsmithr, line 493: lance needs updated

testdata/sqllogictests/functions/current_schemas.slt
```diff
@@ -16,15 +16,17 @@ create schema current_schemas_test;
 statement ok
 set search_path = 'public,current_schemas_test';
 
-query T
+query T rowsort
 select current_schemas()::text;
 ----
-[public, current_schemas_test]
+current_schemas_test
+public
```
scsmithr, line 23: There was some updating to datafusion that changes how lists are handled. I don't want this, but there's now no way for us to reference a session var that's a list and _not_ have it return a column.

crates/sqlexec/src/planner/physical_plan/copy_to.rs
```diff
@@ -161,7 +157,7 @@ impl CopyToExec {
         };
 
         let stream = execute_stream(self.source, context.clone())?;
-        let count = sink.write_all(vec![stream], &context).await?;
+        let count = sink.write_all(stream, &context).await?;
```
tychoish, line 160: this is maybe actually a big deal? wouldn't this mean that we don't have to collect the entire stream in memory to write it?

tychoish (commented):
I just fixed the lint error, there's also a bigquery slt failure (in addition to the missing lance things which are already called out; I think the cause is we're rendering something as a table that used to be rendered as a struct/list, which feels separate. Did some poking, my bet is that a print/format/display helper needs to be implemented or imported that wasn't before?

scsmithr:
> there's also a bigquery slt failure

There's been some updates to datafusion's scalar values which I think are causing the list related issues. What we're seeing is very similar to the `current_schemas` thing above where it seems like lists are now treated as columns, and not a single value in a row.

crates/sqlexec/src/planner/physical_plan/copy_to.rs
scsmithr: Nah this is effectively the same as before. The interface changed from accepting a vector of multiple streams with the implementation being expected to read from each, to just accepting a single stream. I'm not sure why the change was made (and before it was changed to accepting a vector, it previously just accepted a single stream).

tychoish:
does it make sense to split the pyo3 upgrade out out of this PR?

scsmithr:
> does it make sense to split the pyo3 upgrade out out of this PR?

It got bumped in delta or datafusion (I forget which) and so needed to be bumped for us too.

PR #2140, opened 2023-11-21
chore: Send `user_id` in RPC execute request to collect metrics

vrongmeal:


vrongmeal:
Need a way to test this out with a proxy set-up. Would that be possible?

tychoish (approved):
testing aside, this seems pretty safe.

vrongmeal:
Should be good to merge this in.

PR #2147, opened 2023-11-23
feat(kdl): kdl_select function

tychoish:
builds upon and extends #2145 with a corresponding projection. 

it also ends up including some refactors to the matcher for common use. 

If you have not read and approved #2145 do that before approving this.

### DIFF
--- a/crates/sqlexec/src/functions.rs
+++ b/crates/sqlexec/src/functions.rs
@@ -4,6 +4,7 @@ use datafusion::common::ScalarValue;
 use datafusion::error::Result;
 use datafusion::logical_expr::{ColumnarValue, ScalarUDF, Signature, TypeSignature, Volatility};
 use datafusion::prelude::Expr;
+use kdl::{KdlDocument, KdlNode, KdlQuery};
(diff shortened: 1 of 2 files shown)


crates/sqlexec/src/functions.rs
```diff
@@ -256,44 +266,112 @@ fn kdl_matches() -> ScalarUDF {
...
+        }
+        Some(val) => {
+            return Err(datafusion::common::DataFusionError::Execution(format!(
+                "invalid type for KQL expression {}",
+                val.data_type(),
+            )))
+        }
+        None => {
+            return Err(datafusion::common::DataFusionError::Execution(
+                "unknown KQL query".to_string(),
```
scsmithr, line 348: Missing KQL query?

scsmithr (approved):

crates/sqlexec/src/functions.rs
tychoish:
```suggestion
                "missing KQL query".to_string(),
```

PR #2148, opened 2023-11-24
chore: Remove background jobs

vrongmeal:


scsmithr (commented):
What's the rationale for removing background jobs? Will we be adding in a replacement for things that require long-running execution?

tychoish:
> What's the rationale for removing background jobs? Will we be adding in a replacement for things that require long-running execution?

This might be me ([see discussions here](https://github.com/GlareDB/glaredb/pull/1925)) for more discussion. 

The leading question in my mind is: where would background jobs run in the context of distributed execution? Certainly the "run this job on an interval" is not something that (I think) would be reasonable to maintain in GlareDB itself. Even without distributed execution, if you ended up with the same catalog loaded into two instances at once you could end up with multiple executions of the same job, which causes problems if any data is written (internally or externally) as a result of a background job. So the utility is pretty limited. 

Having the ability to run a _single_ job in the background that could continue to execute until completion/cancelation (or instance termination,) is something that I think we could have (and this current implementation could have/use)? This would be good for things like "start table optimize" or a longer running `COPY TO` query. [you could also imagine caching cursors to shared storage, and building a resumable-cursor]. We could use an external service to coordinate interval jobs (e.g. cloud; though the distributed questions there remain, there's just slightly less complexity.) 

But back to this code: 
- giving users a way to "unhook" the current job from the session/its lifecycle is great, but this code doesn't expose that, I think. 
- as glaredb engineers, background jobs are _very_ difficult to use correctly, because they basically cannot have side-effects (and/or we have to accept more-than-once execution) and the use case for that seems very narrow anyway.
- we could keep this table and most of this code but without the interval, as a placeholder for the eventual development of the "single-shot/unhooked background jobs" but I think implementing that might require more design, and might not be compatible with this code anyway.

scsmithr:
> Certainly the "run this job on an interval" is not something that (I think) would be reasonable to maintain in GlareDB itself.

I think this is the thing I agree with the most. Having systems external to glaredb responsible for running jobs (queries) on an interval seems to be a bit cleaner. I do think there may be something at some point that will make us revisit this (like long running ALTER TABLE statements that we might want to just kick to the background). But I think the "simple" cases can be better served outside glaredb.

> Even without distributed execution, if you ended up with the same catalog loaded into two instances at once you could end up with multiple executions of the same job, which causes problems if any data is written (internally or externally) as a result of a background job.

In my mind (if we stuck with background jobs), their state/status wouldn't be stored in the catalog, but in a persistent table on disk (object store). Then each instance would periodically read that table and grab jobs as necessary.

scsmithr (approved):

tychoish:
> In my mind (if we stuck with background jobs), their state/status wouldn't be stored in the catalog, but in a persistent table on disk (object store). Then each instance would periodically read that table and grab jobs as necessary.

100% agree with this. We might want to have some way of doing this that isn't parquet files or columnar at all, _but_ that's extremely a later problem. 

> Having systems external to glaredb responsible for running jobs (queries) on an interval seems to be a bit cleaner.

For a lot of cases (triggers, materialized views, etc.) this is def cleaner and quicker

> I do think there may be something at some point that will make us revisit this (like long running ALTER TABLE statements that we might want to just kick to the background). But I think the "simple" cases can be better served outside glaredb.

- simple case, definitely better for starters
- I think the "persistent cursor," feature (e.g. start an operation, let it run, have a UUID that can be used to check the status of the operation and then access the result set) would cover things like this. 
- glaredb could be made to offload some of those kinds of operations (altertables, optomize, etc.) to be run by the system user (e.g. cloud).

tychoish (approved):

PR #2166, opened 2023-11-27
feat(bson): add COPY TO support for writing bson files

tychoish:
- [x] wire up copy to bson
- [x] test copy to (file) functionality


crates/datasources/Cargo.toml
```diff
@@ -59,6 +59,9 @@ dashmap = "5.5.0"
 calamine = { version = "0.22.1", features = ["dates"] }
 tiberius = { version = "0.12.2", default-features = false, features = ["tds73", "rustls", "chrono"] }
 lance = { git = "https://github.com/universalmind303/lance", rev = "81158eb540ff88ab5b4fce3a1170447760137412" }
+bson = "2.7.0"
+arrow-schema = "47.0.0"
+arrow-array = "47.0.0"
```
scsmithr, line 64: We shouldn't be adding in arrow dependencies directly and instead should use the arrow crate that's re-exported by datafusion (e.g. `use datafusion::arrow::datatypes::SchemaRef;`)

tychoish:
I've done another pass at refactoring this with a couple of notable changes:
- I pulled in some `FileType` handling code from datafusion, so that we could add an option to the enum, which fixed all the test failures. I'm inclined to think that we just sort of have to do this sometimes (datafusion isn't going to want to support all the random things we do), and this is pretty inoffensive. 
- pulled out the `RecordBatch` -> `bson document` converter/iterator. 
- wired up `COPY TO` and left the read/write from file stuff on its own. 

My inclination is to do some course testing of the copy to (which in this mode functions like an export) as a start. 

I _could_ (relatively easily) add "read bson file into memory and have a memory-table" `read_bson` function, but I think it's not that much harder (and more useful generally) to write the functions for the scanning, which is probably copying and tweaking some JSON handling code and doing `s/j/b/` and a few other tweaks. 

I'm going to retitle this to "add COPY TO bson" support, and I think we should merge it, and I'll do the object_store (scanning) thing, in another PR.

crates/datasources/src/common/sink/bson.rs
```diff
@@ -0,0 +1,201 @@
...
+    pub fn from_obj_store(store: Arc<dyn ObjectStore>, loc: impl Into<ObjectPath>) -> BsonSink {
+        BsonSink {
+            store,
+            loc: loc.into(),
+        }
+    }
+
+    async fn pstream_into_inner(&self, stream: SendableRecordBatchStream) -> Result<usize> {
+        self.formatted_stream(stream).await
+    }
```
scsmithr, line 53:
```suggestion
    async fn stream_into_inner(&self, stream: SendableRecordBatchStream) -> Result<usize> {
        self.formatted_stream(stream).await
    }
```

scsmithr:
I think COPY TO makes sense first. That's typically the hardest/most tedious part, so once that's done, the rest should fall into place pretty easily.

tychoish:
> I think COPY TO makes sense first. That's typically the hardest/most tedious part, so once that's done, the rest should fall into place pretty easily.

Definitely the easiest in terms of "this code is isolated from other code" and sufficiently abstract. The integration with object_store is the hard part, I think, as we discussed in the out-of-band conversation.

scsmithr:
👀 
```
FAILED tests/tests/test_bson.py::test_copy_to - psycopg2.errors.InternalError_: internal error: unsupported output format: bson
```

crates/datasources/src/bson/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        Self {
+            batch: batch.clone(),
+            schema: field_names,
+            row: 0,
+            started: false,
+            columns: Vec::<Vec<bson::Bson>>::with_capacity(batch.num_columns()),
+        }
+    }
```
scsmithr, line 30:
```suggestion
    pub fn new(batch: StructArray, fields: Fields) -> Self {
        let mut field_names = Vec::with_capacity(fields.len());
        for field in &fields {
            field_names.push(field.name().to_owned())
        }

        Self {
            batch: batch.clone(),
            schema: field_names,
            row: 0,
            started: false,
            columns: Vec::with_capacity(batch.num_columns()),
        }
    }
```

Types can be inferred here.

crates/datasources/src/bson/mod.rs
```diff
@@ -0,0 +1,281 @@
...
+}
+
+impl Iterator for BsonBatchConverter {
+    type Item = bson::Document;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if !self.started {
+            self.setup().ok()?;
+        }
```
scsmithr, line 48: We lose the error here and the iterator just stops. If we can't convert something to bson, there's no indication to the user that we failed to do so, and instead write out partial data.

Would it make sense to have the iterator return `Result<bson::Document, ArrowError>`?

crates/protogen/src/sqlexec/copy_to.rs
```diff
@@ -102,6 +102,9 @@ impl TryFrom<crate::metastore::types::options::CopyToFormatOptions> for CopyToFo
         value: crate::metastore::types::options::CopyToFormatOptions,
     ) -> Result<Self, Self::Error> {
         match value {
+            crate::metastore::types::options::CopyToFormatOptions::Bson => {
+                Ok(CopyToFormatOptions::default())
```
scsmithr, line 106: Should we be specifying bson specific options here (even if empty)?

crates/datasources/src/bson/mod.rs
tychoish: I don't like it 😂

crates/protogen/src/sqlexec/copy_to.rs
tychoish: there aren't any options right now, (I think?) and it feels like we can add this later when we need to, right?

crates/datasources/src/bson/mod.rs
scsmithr: I'm fine with whatever here and just wanted to point out it's an option.

crates/protogen/src/sqlexec/copy_to.rs
scsmithr: Seems reasonable then, I _think_ using default here is fine. If everything passes, then we're handling that fine.

tychoish: eventually it might be nice to have a "write to n partitions" to facilitate parallel table scans, or to be able to write with hive-partitioning, but for now I think having this be "one file output" is good enough.

crates/datasources/src/bson/mod.rs
```diff
@@ -40,11 +38,14 @@ impl BsonBatchConverter {
...
-    type Item = bson::Document;
+    type Item = Result<bson::Document, ArrowError>;
 
     fn next(&mut self) -> Option<Self::Item> {
         if !self.started {
-            self.setup().ok()?;
+            match self.setup() {
+                Ok(_) => {}
+                Err(e) => return Some(Err(e)),
+            };
```
scsmithr, line 48:
```
if let Err(e) = self.setup() {
    return Some(Err(e));
}
```

scsmithr (approved):

PR #2167, opened 2023-11-27
chore: Split session catalog out into its own crate

scsmithr:
Adds a `catalog` crate which contains `SessionCatalog`, `TempCatalog` and the metastore client. This is mostly needed to allow table functions to have access to the catalog directly, and not have to write one-off methods on the table function context to get entries.

This may also allow us to move the physical/logical plans into their own crate if we feel that helps with code organization.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1385,6 +1385,24 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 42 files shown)


tychoish (approved):
seems fine!

PR #2172, opened 2023-11-28
feat: Expose catalog through table func context

scsmithr:
Exposes the session catalog through the table func context provider.

Also a bit of refactoring here and there.


crates/sqlexec/src/planner/physical_plan/create_table.rs
```diff
@@ -148,9 +151,19 @@ impl CreateTableExec {
...
+        //
+        // TODO: We should be returning _what_ was updated from metastore
+        // instead of needing to do this. Sean has a stash working on this.
+        let new_catalog = SessionCatalog::new(
+            state,
+            ResolveConfig {
+                default_schema_oid: 0,
+                session_schema_oid: 0,
+            },
+        );
```
universalmind303, line 166: Been looking at this & i don't think i fully understand the `ResolveConfig` and what's happening here. 

AFAIK, the `ResolveConfig` is just a config object containing the global schema and session ids? 

Why are we setting them to `0` here? Is this just a temporary workaround?

scsmithr: Yep that's pretty much it. The resolve config here is set to zeros since we don't actually need them, we're just creating a new session catalog to pull out the new table entry that metastore just created.

And `sqlbuiltins` is now importing quite a few crates which make it tricky to import into the `catalog` crate (cycle) and so I had to resort to the resolve config. I want to look at possibly splitting up `sqlbuiltins` to avoid that cycle, but figured it'd be a bit of a refactor.

universalmind303 (approved):

vrongmeal (approved):

PR #2178, opened 2023-11-29
feat: builtin descriptions & examples

### LINKED ISSUE #2177: add `example` and `description` columns to `glare_catalog.functions`
## Description

A lot of the information for the docs is manually curated & only accessible on the docs page. This makes it much more susceptible to documentation drift. It'd be nice to move some of the more simple examples & descriptions into the code. This would make them more accessible, as well as make the likelihood of the documentation going stale lower. 

Proposed solution. 

add  `example` and `description` columns to `glare_catalog.functions`
```sql
> select function_name, example from glare_catalog.functions where function_type = 'table' and example is not null;
┌───────────────┬──────────────────────────────────────────────────────────────────────────────────────────────────┐
│ function_name │ example                                                                                          │
│ ──            │ ──                                                                                               │
│ Utf8          │ Utf8                                                                                             │
╞═══════════════╪══════════════════════════════════════════════════════════════════════════════════════════════════╡
│ parquet_scan  │ ↵                                                                                                │
│               │ -- Read a relative path.↵                                                                        │
│               │ SELECT * FROM parquet_scan('./my_data.parquet');↵                                                │
│               │ -- Read all parquet files in a directory.↵                                                       │
│               │ SELECT * FROM parquet_scan('./directory_of_data/*.parquet');↵                                    │
│               │ -- Read parquet files from multiple directories.↵                                                │
│               │ SELECT * FROM parquet_scan('./**/*.parquet');↵                                                   │
│               │ -- Read multiple explicitly provided files.↵                                                     │
│               │ SELECT * FROM parquet_scan(['./directory_of_data/1.parquet', './directory_of_data/2.parquet']);↵ │
│               │                                                                                                  │
│ csv_scan      │ ↵                                                                                                │
│               │ -- Read a relative path.↵                                                                        │
│               │ SELECT * FROM csv_scan('./my_data.csv');↵                                                        │
│               │ -- Read all csv files in a directory.↵                                                           │
│               │ SELECT * FROM csv_scan('./directory_of_data/*.csv');↵                                            │
│               │ -- Read csv files from multiple directories.↵                                                    │
│               │ SELECT * FROM csv_scan('./**/*.csv');↵                                                           │
│               │ -- Read multiple explicitly provided files.↵                                                     │
│               │ SELECT * FROM csv_scan(['./directory_of_data/1.csv', './directory_of_data/2.csv']);↵             │
│               │                                                                                                  │
│ ndjson_scan   │ ↵                                                                                                │
│               │ -- Read a relative path.↵                                                                        │
│               │ SELECT * FROM json_scan('./my_data.json');↵                                                      │
│               │ -- Read all json files in a directory.↵                                                          │
│               │ SELECT * FROM json_scan('./directory_of_data/*.json');↵                                          │
│               │ -- Read json files from multiple directories.↵                                                   │
│               │ SELECT * FROM json_scan('./**/*.json');↵                                                         │
│               │ -- Read multiple explicitly provided files.↵                                                     │
│               │ SELECT * FROM json_scan(['./directory_of_data/1.json', './directory_of_data/2.json']);↵          │
│               │                                                                                                  │
└───────────────┴──────────────────────────────────────────────────────────────────────────────────────────────────┘
```

universalmind303:
closes #2177.

Still need to finish adding examples and descriptions to all builtin functions, but this gives us a simple example & brief description for each builtin function. 

example:

```sql
> select function_name, example, description from glare_catalog.functions where example is not null ORDER BY function_name asc;
┌────────────────────────────────────┬───────────────────────────────────────┬─────────────────────────────────────────────────────────────────────┐
│ function_name                      │ example                               │ description                                                         │
│ ──                                 │ ──                                    │ ──                                                                  │
│ Utf8                               │ Utf8                                  │ Utf8                                                                │
╞════════════════════════════════════╪═══════════════════════════════════════╪═════════════════════════════════════════════════════════════════════╡
│ approx_distinct                    │ approx_distinct(a)                    │ Gives the approximate count of distinct elements using HyperLogLog. │
│ approx_median                      │ approx_median(a)                      │ Gives the approximate median of a column                            │
│ approx_percentile_cont             │ approx_percentile_cont(a)             │ Gives the approximate percentile of a column                        │
│ approx_percentile_cont_with_weight │ approx_percentile_cont_with_weight(a) │ Gives the approximate percentile of a column with a weight column   │
│ array_agg                          │ array_agg(a)                          │ Returns a list containing all the values of a column                │
│ avg                                │ avg(a)                                │ Returns the average of a column                                     │
│ bit_and                            │ bit_and(a)                            │ Returns the bitwise AND of a column                                 │
│ bit_or                             │ bit_or(a)                             │ Returns the bitwise OR of a column                                  │
│ bit_xor                            │ bit_xor(a)                            │ Returns the bitwise XOR of a column                                 │
│ bool_and                           │ bool_and(a)                           │ Returns the boolean AND of a column                                 │
│ …                                  │ …                                     │ …                                                                   │
│ regr_r2                            │ regr_r2(x, y)                         │ Returns the coefficient of determination (R-squared) for non-null … │
│ regr_slope                         │ regr_slope(y, x)                      │ Returns the slope of the linear regression line for non-null pairs… │
│ regr_sxx                           │ regr_sxx(y, x)                        │ Returns the sum of squares of the independent variable for non-nul… │
│ regr_sxy                           │ regr_sxy(y, x)                        │ Returns the sum of products of independent times dependent variabl… │
│ regr_syy                           │ regr_syy(y, x)                        │ Returns the sum of squares of the dependent variable for non-null … │
│ stddev                             │ stddev(a)                             │ Returns the sample standard deviation of a column                   │
│ stddev_pop                         │ stddev_pop(a)                         │ Returns the population standard deviation of a column               │
│ sum                                │ sum(a)                                │ Returns the sum of a column                                         │
│ variance                           │ variance(a)                           │ Returns the sample variance of a column                             │
│ variance_pop                       │ variance_pop(a)                       │ Returns the population variance of a column                         │
└────────────────────────────────────┴───────────────────────────────────────┴─────────────────────────────────────────────────────────────────────┘
```


crates/protogen/proto/metastore/catalog.proto
```diff
@@ -125,7 +125,11 @@ message EntryMeta {
   // Temp objects should never be persisted.
   bool is_temp = 7;
 
-  // next: 8
+  // Optional example string for
```
scsmithr, line 128: for ...

scsmithr (approved):

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    ASINH,
+    name => "asinh",
+    example => "asinh(1)",
+    description => "Compute the inverse hyperbolic sine of a number."
+);
+make_const!(
+    ATAN,
+    name => "atan",
+    example => "atan(1)",
+    description => "Computes the arctangent of a number."
```
tychoish, line 52: I think given the style set on line 40, this should be `arc tangent` and maybe say (inverse tangent?) and maybe the word "arc" should appear in the description on 46.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    ) => {
+        const $var_name: (&str, &str, &str) = ($name, $example, $description);
+    };
+}
+
+make_const!(
+    ABS,
+    name => "abs",
+    example => "abs(-1)",
+    description => "Compute the absolute value of a number."
```
tychoish, line 21: In all of these excamples we say "compute" rather than return. I think return is useful because it indicates that there's output, and I might use "compute" in situations where the function was just used for side effects.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    COSH,
+    name => "cosh",
+    example => "cosh(0)",
+    description => "Compute the hyperbolic cosine of a number."
+);
+make_const!(
+    DEGREES,
+    name => "degrees",
+    example => "degrees(1)",
+    description => "Converts an angle measured in radians to an approximately equivalent angle measured in degrees."
```
tychoish, line 100: what is the method of approximation? is it floating point rounding?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    FACTORIAL,
+    name => "factorial",
+    example => "factorial(5)",
+    description => "Compute the factorial of a number."
+);
+make_const!(
+    FLOOR,
+    name => "floor",
+    example => "floor(1.1)",
+    description => "Compute the largest integer less than or equal to a number."
```
tychoish, line 118: I would not assume (though perhaps I should,) that "`number`" meant float.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    NANVL,
+    name => "nanvl",
+    example => "nanvl(1, 2)",
+    description => "Returns the first argument if it is not NaN, otherwise returns the second argument."
+);
+make_const!(
+    PI,
+    name => "pi",
+    example => "pi()",
+    description => "Compute the constant π."
```
tychoish, line 178: does it actually compute pi for every row or just return a constant approximation?" To what precision?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+);
+
+make_const!(
+    POWER,
+    name => "pow",
+    example => "pow(2, 3)",
+    description => "Compute the power of a number."
```
tychoish, line 185: should we clarify that the second number is the exponent?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    RADIANS,
+    name => "radians",
+    example => "radians(1)",
+    description => "Converts an angle measured in degrees to an approximately equivalent angle measured in radians."
+);
+make_const!(
+    RANDOM,
+    name => "random",
+    example => "random()",
+    description => "Compute a random number between 0 and 1."
```
tychoish, line 197: I think I'd use "generates" or "returns" a random number. Also maybe designate if/how it's psudorandom?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+);
+make_const!(
+    ROUND,
+    name => "round",
+    example => "round(1.1)",
+    description => "Round a number to the nearest integer."
```
tychoish, line 203: does it round numbers that are equidistant toward or away from zero?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+make_const!(
+    TRUNC,
+    name => "trunc",
+    example => "trunc(1.1111, 2)",
+    description => "Truncate a number to the nearest integer. If a second argument is provided, truncate to the specified number of decimal places."
+);
+make_const!(
+    COALESCE,
+    name => "coalesce",
+    example => "coalesce(null, 1)",
```
tychoish, line 250: is the first argument a list, since this is a scalar function and not a window function?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    CONCAT_WITH_SEPARATOR,
+    name => "concat_ws",
+    example => "concat_ws(',', 'hello', 'world')",
+    description => "Concatenate two strings with a separator."
+);
+make_const!(
+    CHR,
+    name => "chr",
+    example => "chr(97)",
+    description => "Compute the character with the given ASCII code."
```
tychoish, line 299: what happens for numbers outside of the ascii range?

tychoish:
> The approach I've taken is consistent with similar tools in our field. In fact, many of these are copied from those tools. 

I mean I don't think this makes them good, or helpful, and I think we (maybe?) have the opportunity to help and make using our interface easier? If having help text for our methods is important, then having it be clear and useful, should also be important. 

> I'm open to making necessary changes. However, given the volume of comments, it would be helpful to prioritize the most critical ones. If there are specific changes you feel strongly about, could you make PR suggestions for those specific ones? The suggestions are much easier to apply for small changes like this.

I feel moderately strongly about it: while I did have some that were stylistic, many of my comments point out unclear or confusing semantics or situations where the string is actively confusing.

Improving things is going to take one of us a few hours of editing. Maybe the best path forward is:
- read through things and spend some amount of (time boxed, 90 minutes? 2 hours) making changes.
- copy anything you don't want to address to TODOs or similar?
- I'll make a PR after you merge that (and this) to further clean things up?

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+        description => $description:expr
+    ) => {
+        const $var_name: (&str, &str, &str) = ($name, $example, $description);
+    };
+}
+
+make_const!(
+    ABS,
+    name => "abs",
+    example => "abs(-1)",
```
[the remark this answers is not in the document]
universalmind303: The type of arguments is specified in the `parameters` field in the table.

crates/sqlbuiltins/src/functions/scalars.rs
universalmind303: it's a number because we do implicit casting to floats. `floor(1)` is valid.

crates/sqlbuiltins/src/functions/scalars.rs
universalmind303: it's as the description says. Returns the first non null **argument**. 
- `coalesce(1,2,3,4,5,6,7, null, 8) -> 1`
- `coalesce(null, 1) -> 1`
- `coalesce(null, null, null, 1) -> 1`

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,870 @@
...
+    FROM_UNIXTIME,
+    name => "from_unixtime",
+    example => "from_unixtime(1600000000)",
+    description => "Converts a unix timestamp (seconds since 1970-01-01 00:00:00 UTC) to a timestamp (Timestamp<s, UTC>)."
+);
+make_const!(
+    DIGEST,
+    name => "digest",
+    example => "digest('hello', 'sha256')",
+    description => " Compute the digest of a string using the specified algorithm. Valid algorithms are: md5, sha224, sha256, sha384, sha512, blake2s, blake2b, blake3"
```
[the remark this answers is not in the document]
universalmind303: yes, the parameters field specifies what dtypes are allowed.

universalmind303:
> I feel moderately strongly about it: while I did have some that were stylistic, many of my comments point out unclear or confusing semantics or situations where the string is actively confusing. 

What i was suggesting is using the pr `suggestions` instead of comments for the comments that are suggesting a direct change such as "use Return instead of Compute".  This makes it much easier to apply the changes (just click `accept suggestion`) instead of going through every comment & having to reason about what the change should be. 

I also don't think we want overly verbose examples & descriptions in the catalogs. I do think the "extended" examples are helpful, but i don't think the `functions` table is the best spot for that level of information. 

IMO, this is very readable from the command line, and gives you at least a starting point on what the function is, and what it does.

```
> select function_name, example, description from glare_catalog.functions where function_type = 'scalar' order by function_name asc;
┌────────────────┬──────────────────────────────────┬───────────────────────────────────────────────────────────────────────────┐
│ function_name  │ example                          │ description                                                               │
│ ──             │ ──                               │ ──                                                                        │
│ Utf8           │ Utf8                             │ Utf8                                                                      │
╞════════════════╪══════════════════════════════════╪═══════════════════════════════════════════════════════════════════════════╡
│ abs            │ abs(-1)                          │ Compute the absolute value of a number.                                   │
│ acos           │ acos(0.5)                        │ Compute the inverse cosine (arc cosine) of a number.                      │
│ acosh          │ acosh(1)                         │ Compute the inverse hyperbolic cosine of a number.                        │
│ array_append   │ array_append([1, 2], 3)          │ Append an element to the end of an array.                                 │
│ array_concat   │ array_concat([1, 2], [3, 4])     │ Concatenate two arrays.                                                   │
│ array_contains │ array_contains([1, 2], 1)        │ Returns true if the array contains the specified element                  │
│ array_dims     │ array_dims([1, 2])               │ Returns the dimensions of an array.                                       │
│ array_element  │ array_element([1, 2], 1)         │ Returns the element of an array at the specified index.                   │
│ array_has_all  │ array_has_all([1, 2], [1, 2, 3]) │ Returns true if the first array contains all elements of the second array │
│ array_has_any  │ array_has_any([1, 2], [1, 2, 3]) │ Returns true if the first array contains any elements of the second array │
│ …              │ …                                │ …                                                                         │
│ to_hex         │ TODO                             │ TODO                                                                      │
│ to_timestamp   │ to_timestamp('2020-09-08T12:00:… │ Converts a string to a timestamp (Timestamp<ns, UTC>). Alias for `TIMEST… │
│ to_timestamp_… │ to_timestamp_micros('2020-09-08… │ Converts a string to a timestamp with microsecond precision (Timestamp<…  │
│ to_timestamp_… │ to_timestamp_millis('2020-09-08… │ Converts a string to a timestamp with millisecond precision (Timestamp<m… │
│ to_timestamp_… │ to_timestamp_seconds('2020-09-0… │ Converts a string to a timestamp with second precision (Timestamp<s, UTC… │
│ translate      │ translate('hello', 'el', '12')   │ Replace all occurrences of a set of characters in a string with a new se… │
│ trim           │ trim(' hello ')                  │ Remove all spaces from the start and end of a string.                     │
│ trunc          │ trunc(1.1111, 2)                 │ Truncate a number to the nearest integer. If a second argument is provid… │
│ upper          │ upper('hello')                   │ Convert a string to uppercase.                                            │
│ uuid           │ uuid()                           │ Generate a random UUID.                                                   │
└────────────────┴──────────────────────────────────┴───────────────────────────────────────────────────────────────────────────┘
```
anything more than that, such as what I have for the table functions _(but want to find a way to slim down)_ isn't as easy to grok.

```sql
> select function_name, example, description from glare_catalog.functions where function_type = 'table' order by function_name asc;
┌────────────────────┬─────────────────────────────────────────────────────────────────────────────────────────┬────────────────────────────────────────────────────────────┐
│ function_name      │ example                                                                                 │ description                                                │
│ ──                 │ ──                                                                                      │ ──                                                         │
│ Utf8               │ Utf8                                                                                    │ Utf8                                                       │
╞════════════════════╪═════════════════════════════════════════════════════════════════════════════════════════╪════════════════════════════════════════════════════════════╡
│ csv_scan           │ ↵                                                                                       │ ↵                                                          │
│                    │ -- Read a relative path.↵                                                               │ Syntax:↵                                                   │
│                    │ SELECT * FROM csv_scan('./my_data.csv');↵                                               │ -- Single url or path.↵                                    │
│                    │ -- Read all csv files in a directory.↵                                                  │ csv_scan(<url>)↵                                           │
│                    │ SELECT * FROM csv_scan('./directory_of_data/*.csv');↵                                   │ -- Multiple urls or paths.↵                                │
│                    │ -- Read csv files from multiple directories.↵                                           │ csv_scan([<url>])↵                                         │
│                    │ SELECT * FROM csv_scan('./**/*.csv');↵                                                  │ -- Using a cloud credentials object.↵                      │
│                    │ -- Read multiple explicitly provided files.↵                                            │ csv_scan(<url>, <credential_object>)↵                      │
│                    │ SELECT * FROM csv_scan(['./directory_of_data/1.csv', './directory_of_data/2.csv']);↵    │ -- Required named argument for S3 buckets.↵                │
│                    │                                                                                         │ csv_scan(<url>, <credentials_object>, region =>↵           │
│                    │                                                                                         │ '<aws_region>')↵                                           │
│                    │                                                                                         │ -- Pass S3 credentials using named arguments.↵             │
│                    │                                                                                         │ csv_scan(<url>, access_key_id => '<aws_access_key_id>',↵   │
│                    │                                                                                         │ secret_access_key => '<aws_secret_access_key>', region =>↵ │
│                    │                                                                                         │ '<aws_region>')↵                                           │
│                    │                                                                                         │ -- Pass GCS credentials using named arguments.↵            │
│                    │                                                                                         │ csv_scan(<url>, service_account_key =>↵                    │
│                    │                                                                                         │ '<gcp_service_account_key>'                                │
│ delta_scan         │ NULL                                                                                    │ NULL                                                       │
│ generate_series    │ NULL                                                                                    │ NULL                                                       │
│ iceberg_data_files │ NULL                                                                                    │ NULL                                                       │
│ iceberg_scan       │ NULL                                                                                    │ NULL                                                       │
│ iceberg_snapshots  │ NULL                                                                                    │ NULL                                                       │
│ lance_scan         │ NULL                                                                                    │ NULL                                                       │
│ list_columns       │ NULL                                                                                    │ NULL                                                       │
│ list_schemas       │ NULL                                                                                    │ NULL                                                       │
│ list_tables        │ NULL                                                                                    │ NULL                                                       │
│ ndjson_scan        │ ↵                                                                                       │ ↵                                                          │
│                    │ -- Read a relative path.↵                                                               │ Syntax:↵                                                   │
│                    │ SELECT * FROM json_scan('./my_data.json');↵                                             │ -- Single url or path.↵                                    │
│                    │ -- Read all json files in a directory.↵                                                 │ json_scan(<url>)↵                                          │
│                    │ SELECT * FROM json_scan('./directory_of_data/*.json');↵                                 │ -- Multiple urls or paths.↵                                │
│                    │ -- Read json files from multiple directories.↵                                          │ json_scan([<url>])↵                                        │
│                    │ SELECT * FROM json_scan('./**/*.json');↵                                                │ -- Using a cloud credentials object.↵                      │
│                    │ -- Read multiple explicitly provided files.↵                                            │ json_scan(<url>, <credential_object>)↵                     │
│                    │ SELECT * FROM json_scan(['./directory_of_data/1.json', './directory_of_data/2.json']);↵ │ -- Required named argument for S3 buckets.↵                │
│                    │                                                                                         │ json_scan(<url>, <credentials_object>, region =>↵          │
│                    │                                                                                         │ '<aws_region>')↵                                           │
│                    │                                                                                         │ -- Pass S3 credentials using named arguments.↵             │
│                    │                                                                                         │ json_scan(<url>, access_key_id => '<aws_access_key_id>',↵  │
│                    │                                                                                         │ secret_access_key => '<aws_secret_access_key>', region =>↵ │
│                    │                                                                                         │ '<aws_region>')↵                                           │
│                    │                                                                                         │ -- Pass GCS credentials using named arguments.↵            │
│                    │                                                                                         │ json_scan(<url>, service_account_key =>↵                   │
│                    │                                                                                         │ '<gcp_service_account_key>'                                │
│ parquet_scan       │ ↵                                                                                       │ ↵                                                          │
│                    │ -- Read a relative path.↵                                                               │ Syntax:↵                                                   │
│                    │ SELECT * FROM parquet_scan('./my_data.parquet');↵                                       │ -- Single url or path.↵                                    │
│                    │ -- Read all parquet files in a directory.↵                                              │ parquet_scan(<url>)↵                                       │
│                    │ SELECT * FROM parquet_scan('./directory_of_data/*.parquet');↵                           │ -- Multiple urls or paths.↵                                │
│                    │ -- Read parquet files from multiple directories.↵                                       │ parquet_scan([<url>])↵                                     │
│                    │ SELECT * FROM parquet_scan('./**/*.parquet');↵                                          │ -- Using a cloud credentials object.↵                      │
│                    │ -- Read multiple explicitly provided files.↵                                            │ parquet_scan(<url>, <credential_object>)↵                  │
│                    │ SELECT * FROM parquet_scan(['./directory_of_data/1.parquet',↵                           │ -- Required named argument for S3 buckets.↵                │
│                    │ './directory_of_data/2.parquet']);↵                                                     │ parquet_scan(<url>, <credentials_object>, region =>↵       │
│                    │                                                                                         │ '<aws_region>')↵                                           │
│                    │                                                                                         │ -- Pass S3 credentials using named arguments.↵             │
│                    │                                                                                         │ parquet_scan(<url>, access_key_id =>↵                      │
│                    │                                                                                         │ '<aws_access_key_id>', secret_access_key =>↵               │
│                    │                                                                                         │ '<aws_secret_access_key>', region => '<aws_region>')↵      │
│                    │                                                                                         │ -- Pass GCS credentials using named arguments.↵            │
│                    │                                                                                         │ parquet_scan(<url>, service_account_key =>↵                │
│                    │                                                                                         │ '<gcp_service_account_key>'                                │
│ read_bigquery      │ NULL                                                                                    │ NULL                                                       │
│ read_excel         │ NULL                                                                                    │ NULL                                                       │
│ read_mongodb       │ NULL                                                                                    │ NULL                                                       │
│ read_mysql         │ NULL                                                                                    │ NULL                                                       │
│ read_postgres      │ NULL                                                                                    │ NULL                                                       │
│ read_snowflake     │ NULL                                                                                    │ NULL                                                       │
└────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────┴────────────────────────────────────────────────────────────┘
> 
```



I like the idea of the suggested `help` function. Maybe we could use that to encapsulate more detailed information

maybe something like

```
\help digest
```md
## Syntax
`digest(<expr>, <method>)`

## Arguments
<expr>     A string or binary expression, the message to be digested. 
<method>  the algorithm to use for digest. Valid options are 'md5' | 'sha224' | 'sha256' | 'sha384' | 'sha512' | 'blake2s' | 'blake2b' | 'blake3'

## Returns
`Binary`

## Examples
> select digest('hello', 'md5');
┌───────────────────────────────────┐
│ digest(Utf8("hello"),Utf8("md5")) │
│ ──                                │
│ Binary                            │
╞═══════════════════════════════════╡
│ 5d41402abc4b2a76b9719d911017c592  │
└───────────────────────────────────┘
```

crates/sqlbuiltins/src/functions/aggregates.rs
```diff
@@ -0,0 +1,363 @@
...
+make_const!(
+    MIN,
+    name => "min",
+    example => "min(a)",
+    description => "Returns the minimum value in a column"
+);
+make_const!(
+    REGR_AVGX,
+    name => "regr_avgx",
+    example => "regr_avgx(y, x)",
```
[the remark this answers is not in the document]
universalmind303: Yes, typically things are ordered alphabetically like x, y, but in statistical functions like `regr_avgx`, the order `y, x` is used because it follows the convention of modeling `y` (dependent variable) as a function of `x` (independent variable). This is standard in statistical and regression analysis contexts, despite the potential for initial mental friction. 

It's also consistent with other doc examples.

https://docs.snowflake.com/en/sql-reference/functions/regr_avgx
https://duckdb.org/docs/sql/aggregates.html

crates/sqlbuiltins/src/functions/aggregates.rs
```diff
@@ -0,0 +1,363 @@
...
+    REGR_AVGY,
+    name => "regr_avgy",
+    example => "regr_avgy(y, x)",
+    description => "Returns the average of the dependent variable for non-null pairs in a group, where x is the independent variable and y is the dependent variable."
+);
+make_const!(
+    REGR_COUNT,
+    name => "regr_count",
+    example => "regr_count(y, x)",
+    description => "Returns the number of non-null number pairs in a group."
```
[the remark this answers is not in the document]
universalmind303: All of the `aggregate` functions are designed to be used in combination with some grouping. 
```sql
select regr_count(y, x) from aggr; -- a single aggregate select
select k, regr_count(y, x) from aggr group by k; -- a groupby
select regr_count(y, x) over (partition by k) from aggr; -- window function
```

universalmind303:
@tychoish i don't mean to be argumentative or combative. I do agree that we should really strive to have good documentation, i think we are just a bit misaligned on the manifestation of those. I'm of the mind that the table fields should be concise & serve as an "entry point". (The ones addressed in this PR).  Anything beyond that (such as ones you'd find in the function reference) should not be part of the `glare_catalog.functions` table. IMO It's just information overload. 

I do really like how [snowflake](https://docs.snowflake.com/en/sql-reference-functions.) has their functions documented. They provide examples, outputs, syntax & more. Doing something like that via `\help` or `select help(<function>)` seems like a better solution to me. That way users only get the "detailed" information if it's explicitly requested, not by default.

crates/sqlbuiltins/src/functions/scalars.rs
tychoish: Having the precision would be useful, I would think.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,994 @@
...
+    strpos
+}
+document! {
+    "Extract a substring from a string with the given start position and length.",
+    "substr('hello', 2, 2)",
+    substr
+}
+document! {
+    "TODO",
+    "TODO",
```
tychoish, line 345: I think this should be populated before merging.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,994 @@
...
+    "to_timestamp('2020-09-08T12:00:00+00:00')",
+    to_timestamp
+}
+document! {
+    "Converts a string to a timestamp with millisecond precision (Timestamp<ms, UTC>)",
+    "to_timestamp_millis('2020-09-08T12:00:00+00:00')",
+    to_timestamp_millis
+}
+document! {
+    "Converts a string to a timestamp with microsecond precision (Timestamp<µs, UTC>)",
```
tychoish, line 419: if these are RFC3339 timestamp strings, then we should say that.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,994 @@
...
+    "date_part('year', '2020-01-01')",
+    date_part
+}
+document! {
+    "Converts a string to a timestamp (Timestamp<ns, UTC>). Alias for `TIMESTAMP <string>`.",
+    "to_timestamp('2020-09-08T12:00:00+00:00')",
+    to_timestamp
+}
+document! {
+    "Converts a string to a timestamp with millisecond precision (Timestamp<ms, UTC>)",
```
tychoish, line 414: the strings are inconsistent about weather or not they are a sentence or not. Some end in  a period, and some don't, I think being consistent is good styple.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,994 @@
...
+    "array_concat([1, 2], [3, 4])",
+    array_concat
+}
+document! {
+    "Returns the dimensions of an array.",
+    "array_dims([[[1]]])",
+    array_dims
+}
+document! {
+    "Check if an array is empty. Returns true if empty",
```
tychoish, line 494: Most of the verbs in these string are indicative (e.g. "Returns <>" "Converts <>") but in this and the previous few the verbs are suddenly imperative (e.g. "Check <>" "Concatenate <>" "Append <>"). This specific one is odd because it is two sentences (but I think doesn't need to be we can delete the first one?)

I think indicative verbs are best for this kind of thing, but being inconsistent stands out to me and makes me think harder about if the difference means something.

crates/sqlbuiltins/src/functions/scalars.rs
```diff
@@ -0,0 +1,994 @@
...
+    "empty([])",
+    empty
+}
+document! {
+    "Returns the element of an array at the specified index (using one-based indexing)",
```
tychoish, line 499: <3

crates/sqlbuiltins/src/builtins.rs
```diff
@@ -601,6 +612,108 @@ impl BuiltinView {
...
+    async fn create_provider(
+        &self,
+        ctx: &dyn TableFuncContextProvider,
+        args: Vec<FuncParamValue>,
+        opts: HashMap<String, FuncParamValue>,
+    ) -> datafusion_ext::errors::Result<Arc<dyn TableProvider>>;
+}
+
+/// The same as `BuiltinFunction` , but with const values.
+pub trait ConstBuiltinFunction: Sync + Send {
```
tychoish, line 639: I quite like this change to the wait the traits for functions are defined.

crates/sqlbuiltins/src/functions/table/object_store.rs
```diff
@@ -42,10 +44,40 @@ pub const READ_JSON: ObjScanTableFunc = ObjScanTableFunc(FileType::JSON, "read_n
...
+    fn sql_example(&self) -> Option<String> {
+        fn build_example(extension: &str) -> String {
+            format!("{ext}_scan('./my_data.{ext}')", ext = extension)
+        }
+        Some(build_example(self.0.to_string().as_str()))
+    }
+    fn description(&self) -> Option<String> {
+        fn build_description(extension: &str) -> String {
+            format!(
+                r#"
```
tychoish, line 63: <question>: in future this gets replaced by the online help you're building from our docs content?

PR #2179, opened 2023-11-29
chore: Send query text when executing query over RPC

vrongmeal:
```json
{
  "database_id": "00000000-0000-0000-0000-000000000000",
  "connection_id": "00000000-0000-0000-0000-000000000000",
  "query_text": "SELECT * FROM my_pg.public.lineitem WHERE l_shipdate <= DATE '1998-12-01' - INTERVAL '90' LIMIT 5",
  "telemetry_tag": "unknown",
  "execution_status": "success",
  "error_message": null,
  "elapsed_compute_ns": 11627950,
  "output_rows": 5,
  "bytes_read": 15466360,
  "bytes_written": 0
}
```

### DIFF
--- a/crates/sqlexec/src/session.rs
+++ b/crates/sqlexec/src/session.rs
@@ -402,12 +402,14 @@ impl Session {
     pub async fn create_physical_plan(
         &self,
         plan: DfLogicalPlan,
+        op: &OperationInfo,
     ) -> Result<Arc<dyn ExecutionPlan>> {
         let state = self.ctx.df_ctx().state();
         let plan = state.optimize(&plan)?;
         if let Some(client) = self.ctx.exec_client() {
             let planner = RemotePhysicalPlanner {
                 database_id: self.ctx.get_database_id(),
+                query_text: op.query_text(),
                 remote_client: client,
                 catalog: self.ctx.get_session_catalog(),
             };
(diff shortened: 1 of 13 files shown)


vrongmeal:
I didn't make many changes to incorporate an `operation_id` right now since there are going to be many changes as we're working on distributed exec but getting the query text in was quick and easy enough.

tychoish:
this is _fine_ in it's current form, getting the operation ID is definitely needed for distributed execution, and I expect that we'll want to pass around an operation context object that will have all of this request-related metadata rather than passing each one in.

crates/sqlexec/src/session.rs
```diff
@@ -493,6 +495,7 @@ impl Session {
     pub async fn execute_inner(
         &mut self,
         plan: LogicalPlan,
+        query_text: &str,
```
tychoish, line 498: this logicalplan is already a wrapper around the datafusion type, so it could become a tuple of the query-text (or empty string) and the LogicalPlan and that way the function signatures don't need to change.

vrongmeal: I tried doing that initially. I can't remember what made me change my mind and do this instead. Anyways, will update!

vrongmeal: While trying to do the same, I realised these two combine and disassociate quickly. They are required together here only. Combining them makes the code elsewhere hard to work with. Yes, this function signature feels tacky, but we can fix that.

tychoish (approved):

Issue #2104, opened 2023-11-15
`SELECT 1` runs on server


vrongmeal:
## Description

**Local:**

```
> select 1;
┌──────────┐
│ Int64(1) │
│       ── │
│    Int64 │
╞══════════╡
│        1 │
└──────────┘
```

**Server logs:**

```
  2023-11-15T13:47:11.392469Z  INFO rpcsrv::handler: executing physical plan, database_id: 00000000-0000-0000-0000-000000000000
    at crates/rpcsrv/src/handler.rs:156 on server-thread-3 ThreadId(5)
    in glaredb::server::rpc_service_request
```

PR #2183, opened 2023-11-30
fix: Hybrid planning when the root plan has local preference

vrongmeal:
Push down "remote" exec as far down as possible and replace any "local"
execs that are found (children to the remote). Finally, pack the
remote exec in a [`SendRecvJoinExec`] so the plan transformation looks
something like:
    
### Original plan:
    
```txt
                 +-----------+
                 | A (Local) |
                 +-----+-----+
                       |
                       v
              +-----------------+
      +-------+ B (Unspecified) +-------+
      |       +--------+--------+       |
      v                |                v
+------------+         |          +-----------+
| C (Remote) |         |          | F (Local) |
+-----+------+         v          +-----------+
      |          +------------+
      v          | E (Remote) |
+-----------+    +------------+
| D (Local) |
+-----------+
```
    
 ### Transformed plan:

```txt
                          +-----------+
                          | A (Local) |
                          +-----+-----+
                                |
                                v
                       +-----------------+
         +-------------+ B (Unspecified) +-------+
         |             +--------+--------+       |
         |                      |                v
         v                      |          +-----------+
+------------------+            |          | F (Local) |
| SendRecvJoinExec |            v          +-----------+
|                  |   +------------------+
|        C         |   | SendRecvJoinExec |
|        |         |   |                  |
|        v         |   |        E         |
|   D (Send-Recv)  |   +------------------+
+------------------+
```
    
> **Note:** All nodes shown in the graphs above (except `B`) are
> [`RuntimeGroupExec`]s. Other nodes are omitted for simplicity.
    
We don't "pull-up" the remote exec as much as possible because we want
to stop pulling once a "local" node is met but there's no certain way of
knowing that the node is to be run locally (since that information lies
with its n'th parent, i.e., [`RuntimeGroupExec`]).
    
For example, take the following plan in consideration which tries to
copy contents from a remote table in a local file:
    
```txt
RuntimeExec (local)
        |
    CopyToExec
        |
      Limit
        |
RuntimeExec (remote)
        |
    TableScan
```
    
If we were to pull "remote" up until the "local", we would end up with
something like:
    
```txt
RuntimeExec (local)
        |
RuntimeExec (remote)
        |
    CopyToExec
        |
      Limit
        |
    TableScan
```
    
This ends up running the `CopyToExec` on remote which is completely
incorrect.

Fixes: #2104

### DIFF
--- a/crates/sqlexec/src/remote/planner.rs
+++ b/crates/sqlexec/src/remote/planner.rs
@@ -69,6 +69,7 @@ impl DDLExtensionPlanner {
         Self { catalog }
     }
 }
+
 #[async_trait]
 impl ExtensionPlanner for DDLExtensionPlanner {
     async fn plan_extension(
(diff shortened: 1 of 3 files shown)


crates/datafusion_ext/src/vars.rs
```diff
@@ -95,6 +95,7 @@ impl SessionVars {
      is_cloud_instance: bool,
      dialect: Dialect,
      enable_experimental_scheduler: bool,
+     prefer_local_execution: bool,
```
tychoish, line 98: - is this really "prefer" and not "force"
- are there meaningfully cases in local/embedded nodes where there's more-than-one session? My instinct was that local sessions (which are the things where hybrid execution could have happened anyway,) were always single user, and had one copy of the glaredb instance loaded/running. 
- seems like we should (maybe) have this be an option on the operation itself rather than on the session (which means that the state of "should this operation be remote or local" is stateful on the session, and could be weird in multi-threaded contexts.
-

scsmithr: `prefer_local_execution` might not be the best name for this, but it's really more about setting the default location of where to run parts of the query. Currently we default to "remote" at the root and selectively pull up "local" parts as much as possible. This variable really just changes it to default "local" at the root.

I suggested a session variable since changing the default _could_ break something that we don't know about, and so having a way to go back to the other behavior seemed reasonable and lets us play with it.

tychoish: can we just have this be "execution_mode" and have it be an enum?

`should_hybrid_execution_default_local` 

------

if it defaults to local at the root will any parts be remote?

crates/sqlexec/src/remote/planner.rs
```diff
@@ -504,28 +507,40 @@ impl<'a> PhysicalPlanner for RemotePhysicalPlanner<'a> {
...
+            let exec = physical.as_any().downcast_ref::<RuntimeGroupExec>();
+            let exec_preference = exec
+                .clone()
+                .map(|e| e.preference)
+                .unwrap_or(RuntimePreference::Unspecified);
+            // if let Some(exec) = exec {
+            //     println!("exec = {}", displayable(exec.child.as_ref()).indent(true));
+            // } else {
+            //     println!("physical = {}", displayable(physical.as_ref()).indent(true));
+            // }
```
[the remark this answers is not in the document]
vrongmeal: Yes.

crates/datafusion_ext/src/vars.rs
vrongmeal: I am working on the same, as Sean mentioned. Right now, the "remote" isn't pulled up; it must also be configurable. I will mark it ready for review once it is done.

tychoish: [sounds good to me! sorry if I complicated things, but this feels clearer!]

scsmithr: > I am working on the same, as Sean mentioned. Right now, the "remote" isn't pulled up; it must also be configurable. I will mark it ready for review once it is done.

Oh, I assumed it was already doing that. If we can have it pull up matching preferences, that would be ideal (and would be useful once extended for dist exec too).

crates/sqlexec/src/remote/planner.rs
```diff
@@ -403,85 +399,198 @@ pub struct RemotePhysicalPlanner<'a> {
...
+    ///       Limit
+    ///         |
+    ///     TableScan
+    /// ```
+    ///
+    /// This ends up running the `CopyToExec` on remote which is completely
+    /// incorrect.
+    // TODO: Figure out how to "pull-up" remote more accurately. One solution
+    // might be to specialize cases where we know we can pull up, i.e., know
+    // that the node has to execute on local.
```
scsmithr, line 494: I really appreciate this comment.

scsmithr (approved):

PR #2186, opened 2023-11-30
wip(for discussion): Undo log

scsmithr:
This is the contents of the stash I referenced here: https://github.com/GlareDB/glaredb/blob/main/crates/sqlexec/src/planner/physical_plan/create_table.rs#L154-L166

This is the very beginning of an undo log for database catalogs. I don't expect this PR to be merged, nor do I think this is the final design, it's being pushed more for the discussion aspect around catalog diffs and/or transaction logging for catalogs.

### DIFF
--- a/crates/metastore/src/database.rs
+++ b/crates/metastore/src/database.rs
@@ -1,4 +1,5 @@
 //! Module for handling the catalog for a single database.
+use crate::entries::{DatabaseEntries, UndoAction, UndoLog};
 use crate::errors::{MetastoreError, Result};
 use crate::storage::persist::Storage;
 use once_cell::sync::Lazy;
@@ -30,10 +31,6 @@ use uuid::Uuid;
 
 /// Special id indicating that databases have no parents.
 const DATABASE_PARENT_ID: u32 = 0;
-
-/// Absolute max number of database objects that can exist in the catalog. This
-/// is not configurable when creating a session.
-const MAX_DATABASE_OBJECTS: usize = 2048;
 
 /// A global builtin catalog. This is meant to be cloned for every database
 /// catalog.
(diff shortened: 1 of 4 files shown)


scsmithr:
> why not store catalogs in delta and use mvcc time travel?

I think that's something we should explore at some point. 

For historical context on the catalog...

- Initially the catalog was going to be backed by persistent tables (https://github.com/GlareDB/glaredb/pull/270).
- Then with the focus on external data sources, the catalog was moved to just be a json blob to avoid worrying about storage (https://github.com/GlareDB/glaredb/pull/405).
- The current iteration is really just an extension of the json catalog (just with protobuf)

However, I still see a benefit of a log of sorts _in addition_ to time travel (and technically we already have time travel with our catalog versions, just not exposed). Having the log lets us ship "diffs" of the catalog to other nodes, instead of the entire catalog itself. The contents of this pr came about because I wanted to see if metastore could just return a "diff" of what's changed so that execs like create table don't need to recreate a session catalog just to get the new entry.

tychoish:
That seems reasonable! Good answers. 

Do we have any concerns about concurrent writes?

I think this change is pretty mild and I don't have a great problem with it. I might like to think (and have us think on our own) about if this interface/semantic exposure gets harder to support if we change the backing method. 

Random thoughts:
- i think it'd be pretty cool to be able to have delta-like mvcc but with other kinds of encoding (and some of my stuff with the datafusion fileformat work is in service of enabling that in the future. (because you wouldn't (probably) want columnar catalogs) 
- [non actionable] Protobufs are great wire-message formats because they're compact, and the semantics of the IDL and generated code are good (even if the code that's generated is less good); _but_ my gut/conventional wisdom is that they're bad for on-disk formats because you can't do incremental/lazy decoding (I think?), and they get sort of unweildly when large. Having said that, our storage layer (S3) means that we interact with storage _as messages_.

scsmithr:
> Do we have any concerns about concurrent writes?

Not with the existing implementation. If we store the catalog in delta tables though, we'd need to figure out multi table transactions for that. 

> i think it'd be pretty cool to be able to have delta-like mvcc but with other kinds of encoding (and some of my stuff with the datafusion fileformat work is in service of enabling that in the future. (because you wouldn't (probably) want columnar catalogs)

With our current system, we already have an ability to time travel (but using versions, not timestamps). Something I've wanted to explore is loading multiple versions of the catalog during execution based on which version the plan was created against. Currently we assume the latest always, but that can lead to consistency issues.

> [non actionable] Protobufs are great wire-message formats because they're compact, and the semantics of the IDL and generated code are good (even if the code that's generated is less good); but my gut/conventional wisdom is that they're bad for on-disk formats because you can't do incremental/lazy decoding (I think?), and they get sort of unweildly when large. Having said that, our storage layer (S3) means that we interact with storage as messages.

I agree with protobuf not being an amazing storage protocol, but we needed something that had a schema, quick to iterate on, and was reasonably performant. Catalogs are unlikely to get very large, and so using protobuf seemed like a reasonable tradeoff.

tychoish:
> If we store the catalog in delta tables though, we'd need to figure out multi table transactions for that.

I imagined we could keep it all in one table and get it to do something like [this](https://www.mongodb.com/docs/atlas/data-federation/supported-unsupported/pipeline/out/#parquet-file-format) with DF union types, but maybe that's a bit extra.

> we already have an ability to time travel (but using versions, not timestamps).

TBH logical clocks are better than system clocks anyway for transactions. 

> Catalogs are unlikely to get very large, and so using protobuf seemed like a reasonable tradeoff.

Yup, exactly, hence the non-actionable flag. It also seems easy enough to change/migrate at some point if we felt differently about it.

tychoish:
I don't really see a way to put a good use-case or interface on top of time-travel/undo for the catalog in the in memory case. I think if operations are idempotent then we can just save the operations and retry applying them after refreshing? But I think the case where that's applicable is pretty narrow.

Issue #1357, opened 2023-07-21
Add `--logs` flag to `local` command


vrongmeal:
## Description

We can specify a file (or even stdout/stderr) for logs when running local mode.

```
glaredb local --logs debug.log
```

## Rationale

Helpful while debugging and figuring out errors (for user).

Lilit0x:
Hi @universalmind303, can I work on this?

PR #2187, opened 2023-12-01
feat: add --logs flag to local command

Lilit0x:
Added `--logs` subcommand to local command. Defaults to `debug.log` when no file is provided.
Closes: #1357


crates/glaredb/tests/log_file_test.rs
```diff
@@ -0,0 +1,33 @@
...
+    let mut cmd = make_cli();
+    cmd.timeout(DEFAULT_TIMEOUT)
+        .arg("-q")
+        .arg("select 1;")
+        .assert()
+        .success();
+    let mut log_file = std::env::current_dir().expect("failed to retrieve current dir");
+    log_file.push("debug.log");
+    assert!(log_file.exists());
+}
```
universalmind303, line 16: i'm not sure we want the default behavior to be write to a log file. IMO it should only write to log file if specified. @vrongmeal Thoughts on this? it looks like you opened up the original issue.

I'd personally prefer something like `glaredb --log-path debug.log` or `glaredb --log-file debug.log`, and if it's not specified, no logs are written to disk.

crates/logutil/src/lib.rs
```diff
@@ -84,8 +86,17 @@ pub fn init(verbosity: impl Into<Verbosity>, mode: LoggingMode) {
             subscriber::set_global_default(subscriber)
         }
         LoggingMode::Full => {
-            let subscriber = full_fmt(level).with_env_filter(env_filter).finish();
-            subscriber::set_global_default(subscriber)
+            let subscriber = full_fmt(level).with_env_filter(env_filter);
+
+            if let Some(file) = log_file {
+                let debug_log = {
+                    let file = File::create(file).expect("Failed to create log file");
```
vrongmeal, line 93: Can we instead just `eprintln` a warning here, instead of a panic!

crates/glaredb/tests/log_file_test.rs
```diff
@@ -0,0 +1,20 @@
...
+        .arg("-q")
+        .arg("select 1;")
+        .arg("--log-file")
+        .arg(log_file_name)
+        .assert()
+        .success();
+
+    let mut log_file = std::env::current_dir().expect("failed to retrieve current dir");
+    log_file.push(log_file_name);
+    assert!(log_file.exists());
```
vrongmeal, line 19: Can we log something and check that it was actually logged? Can just check if the test log string in the file.

crates/glaredb/src/bin/main.rs
```diff
@@ -54,8 +54,12 @@ fn main() -> Result<()> {
...
     // _unless_ the user specified a logging related option.
     match (&command, cli.log_mode, cli.verbose) {
-        (Commands::Local { .. }, None, 0) => (),
-        _ => logutil::init(cli.verbose, cli.log_mode.unwrap_or_default().into()),
+        (Commands::Local(args), None, 0) => {
+            if let Some(file) = &args.log_file {
+                logutil::init(1, LoggingMode::Full.into(), Some(file))
+            }
+        }
+        _ => logutil::init(cli.verbose, cli.log_mode.unwrap_or_default().into(), None),
```
vrongmeal, line 62: We should be log to file with whatever the log mode or verbosity is passed using the command line. This only does it for 0 verbosity and no log mode.

vrongmeal (changes requested):
Left a few comments.

Also, the log file is implemented only for the "full" log mode. I think it should be implemented for all the modes.

crates/logutil/src/lib.rs
Lilit0x: `eprintln` will typically let the execution continue, but there won't be any file to write the logs to.

tychoish (approved):

universalmind303:
@scsmithr @greyscaled can one of you manually merge this?

PR #2188, opened 2023-12-01
fix: atoi and overflow handling for function params

tychoish:


crates/datafusion_ext/src/functions.rs
```diff
@@ -393,13 +393,33 @@ impl FromFuncParamValue for i64 {
...
+                        });
+                    }
+                    Ok(v as i64)
+                }
+                ScalarValue::Utf8(Some(v)) | ScalarValue::LargeUtf8(Some(v)) => v
+                    .parse::<i64>()
+                    .map_err(|_| ExtensionError::InvalidParamValue {
+                        param: v,
+                        expected: "integer",
+                    }),
```
universalmind303, line 410: do we really want implicit casting of strings? 

AFAICT, the only usage of i64 is for `generate_series`. and i dont think we'd want to implicitly allow something like `generate_series('10', '0', '-2')`

crates/datafusion_ext/src/functions.rs
```diff
@@ -393,13 +393,33 @@ impl FromFuncParamValue for i64 {
...
                 }),
             },
-
+            FuncParamValue::Ident(v) => {
+                v.parse::<i64>()
+                    .map_err(|_| ExtensionError::InvalidParamValue {
+                        param: v,
+                        expected: "integer",
+                    })
+            }
```
universalmind303, line 422: what's the case for this one? this seems really weird to me. `f("1")` where `"1"` is a column?

crates/datafusion_ext/src/functions.rs
universalmind303: I think it'd be safer to require explicit casts. `generate_series('10'::int, ...)`. `generate_series(arrow_cast('10', 'Int64'), ...)`. If the function signature is (Int64, Int64), i think we should abide by that, and not do `(Into<Int64>, Into<Int64>)`.

tychoish: does our parser support explicit casts as you suggest or is this aspirational?

universalmind303: > this is hard for the user because in many cases we don't document what the type is 

the parameters are mostly all documented inside the `glare_catalog.functions` table. 

additionally, some of builtin datafusion functions will also display a nice error if you use the wrong type.

```sql
> select avg('11');
Error: Error during planning: No function matches the given name and argument types 'AVG(Utf8)'. You might need to add explicit type casts.
        Candidate functions:
        AVG(Int8/Int16/Int32/Int64/UInt8/UInt16/UInt32/UInt64/Float32/Float64)
```

some display not as nice errors, but still helpful enough to realize you are using the wrong type

```sql
> select abs('11');
Error: Optimizer rule 'simplify_expressions' failed
caused by
This feature is not implemented: Unsupported data type Utf8 for function abs
```

 I think it'd be strange if we did implicit casting for our custom table functions, but not for other functions.

PR #2202, opened 2023-12-04
feat(bson): add table function for bson files

tychoish:
I've been poking at various parts of this intermittently for a week, and I think I have (the hard part) done, which handling and reading the objects and parsing the objects into streams and the streams into documents. This at the very least has a design that I think is reasonably robust, has some room for optimization in the future if we need, compiles, and there are only a few outstanding things to do:

- [x] implement [PartionStream objects](https://docs.rs/datafusion/latest/datafusion/physical_plan/streaming/trait.PartitionStream.html) for:
  - [x] a vector of bson documents
  - [x] the `future::stream` of bson documents.
- [x] refactor this into a proper `TableProvider`, and also split up this giant function.
- [x] testing, so much testing.

I wanted to put this up sooner rather than later to get some eyes on the rust streaming parts which I am only just learning. The implementation of the partition streams will be fairly straightforward, as will be the refactor.

I'm going to take a couple days off of this to: attend to some other things and also work on some testing infrastructure work to make it easier to write some of these tests.


crates/sqlbuiltins/src/functions/bson.rs
```diff
@@ -0,0 +1,177 @@
...
+        // files in parallel, leading to a table that will be streamed
+        let _serial_scan = opts.get("serial_scan").is_some();
+
+        // build a vector of streams, one for each file, that handle BSON's framing.
+        let mut readers = VecDeque::<
+            futures::stream::Map<
+                FramedRead<BufReader, LengthDelimitedCodec>,
+                fn(Result<BytesMut, std::io::Error>) -> Result<bson::Document, ExtensionError>,
+            >,
+        >::with_capacity(list.len());
```
scsmithr, line 94:
```suggestion
        let mut readers = VecDeque::with_capacity(list.len());
```

Could just make it inferred to reduce verbosity.

crates/sqlbuiltins/src/functions/bson.rs
```diff
@@ -0,0 +1,177 @@
...
+                }
+            }
+            first_active += 1;
+        }
+
+        // if we had to read through one or more than of the input files in the glob, we already
+        // have their documents and should truncate the vector of readers.
+        if first_active > 0 && first_active < readers.len() {
+            readers.range(first_active..);
+        }
```
scsmithr, line 159: I don't think we need the check. Also doesn't seem we're making the use the result of the range here.

scsmithr (commented):
This seems reasonable. I might've written setting up the streams a bit differently, but it's difficult to say how without actually writing it. As it is right now, it's very readable with a good bit of comments, and so have no issue with how it's currently written.

crates/sqlbuiltins/src/functions/bson.rs
tychoish: oh interesting; were that I knew that before. 

It did actually help to put the type out like this before while I was writing.

tychoish:
> This seems reasonable. I might've written setting up the streams a bit differently, but it's difficult to say how without actually writing it. As it is right now, it's very readable with a good bit of comments, and so have no issue with how it's currently written.

I definitely struggled with this and tried a few things but this is the one that seemed like it worked. I also started from a certain framing (as it were) of the problem so maybe came at this from a weird perspective. I'm pretty happy with how this part is now, so I probably won't change it, but I'd definitely be curious on what your approach would be.

tychoish:
With great thanks to @scsmithr's rust chops, the `read_bson` function is feature complete as implemented. 

Remaining work: 
- refactoring
- testing
- (another pr/post refactor) adding a table provider (not just the function) 
- (future) operate on raw documents rather than document (we may be copying input data too much)

.github/workflows/ci.yaml
```diff
@@ -151,6 +151,9 @@ jobs:
     name: SQL Logic Tests
     runs-on: ubuntu-latest-8-cores
     needs: ["build"]
+    strategy:
+      matrix:
+        protocol: ['postgres','flightsql']
```
scsmithr, line 156: Nice, looking forward to adding the rpc tests too.

crates/datasources/src/bson/builder.rs
```diff
@@ -51,6 +53,17 @@ impl RecordStructBuilder {
...
+    pub fn len(&self) -> usize {
+        match self.builders.first() {
+            Some(elem) => elem.len(),
+            None => 0,
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.len() > 0
+    }
```
scsmithr, line 65:
```suggestion
    pub fn is_empty(&self) -> bool {
        self.len() == 0
    }
```
?

crates/datasources/src/bson/builder.rs
```diff
@@ -51,6 +53,17 @@ impl RecordStructBuilder {
         })
     }
 
+    pub fn len(&self) -> usize {
+        match self.builders.first() {
+            Some(elem) => elem.len(),
+            None => 0,
+        }
+    }
```
scsmithr, line 61: Semantically, it's difficult for me to understand "what" is being measured here. By looking at the name `RecordStructBuilder`, I would have assumed number of records, but that's not the case here.

scsmithr (approved):
This looks good to me (tried to skip the diffs from flight). Just two comments on `len`/`is_empty`

.github/workflows/ci.yaml
tychoish: sorry for the weird flight merge.

tychoish:
> This looks good to me (tried to skip the diffs from flight). Just two comments on len/is_empty

Sorry about the wacky merge; this branch clearly lingered too long. I think I managed to get it into something real.

crates/datasources/src/bson/builder.rs
scsmithr: Oh, I misread this then.

PR #2203, opened 2023-12-04
chore: improve+cleanup func param processing

tychoish:
While doing #2188 and #2202 I kepy running into `FromFuncParamValue`
and it seems like we get some clarity and simplicty by factoring this
away.


scsmithr:
This looks like a good change.

Seems like there's a test failure:

```
11:13:18.596 ERROR testing::slt::cli: Error while running test `sqllogictests_object_store/azure/copy_to_and_scan` error=test fail: query failed: External error: cannot mix different types of urls
[SQL] SELECT a, b FROM csv_scan(
	[
		'azure://glaredb-test/copy_to/with_opts.csv',
		'azure://glaredb-test/copy_to/with_url.csv',
		'azure://glaredb-test/copy_to/with_creds.csv'
	],
	account_name => '***',
	access_key => '***'
);
at /home/runner/work/glaredb/glaredb/testdata/sqllogictests_object_store/azure/copy_to_and_scan.slt:64
```

but not immediately obvious why that would be failing from this.

tychoish:
> This looks like a good change.

Whew! 😂

> Seems like there's a test failure

Unsurprising. I tend to let the CI system run the integration tests, and have been keeping to just unit tests locally for battery and latency reasons for the past few weeks.

The problem, I think, is that arguments that are lists of vectors are I think you'd expect to be handled as a passthrough, but maybe aren't always, and `DatasourceUrls` are handled weirdly in this case.

scsmithr (approved):

Issue #2168, opened 2023-11-27
Chore: SLT failures with snowflake due to invalid auth tests


scsmithr:
## Description

```
17:42:23.621 ERROR testing::slt::cli: Error while running test `sqllogictests_snowflake/external_database` error=test fail: statement failed: internal error: cannot execute simple query: db error: ERROR: External database validation failed: Snowflake Auth Error (390102): Your user account has been temporarily locked due to too many failed attempts. Try again after 15 minutes or contact your account administrator for assistance. For more information about this error, go to https://community.snowflake.com/s/error-your-user-login-has-been-locked.
[SQL] CREATE EXTERNAL DATABASE external_db
	FROM snowflake
	OPTIONS (
		account = 'hmpfscx-xo23956',
		username = '***',
		password = '***',
		database = 'glaredb_test',
		warehouse = 'compute_wh',
		role = 'accountadmin',
	);
at /home/runner/work/glaredb/glaredb/testdata/sqllogictests_snowflake/external_database.slt:3
```

tychoish:
in earlier discussions we've thought that the problem here is that we see this when we're running too many instances of the tests at the same time, and that we should in some way serialize them (maybe there's an action's helper that could help this?) if nothing else we should split it out of the main integration block so we can see when it's the failure.

tychoish:
I believe that #2278 should have resolved this, if we see it again we can attempt other things.

PR #2205, opened 2023-12-04
fix: handle integer overflows

tychoish:


scsmithr (approved):
See https://github.com/GlareDB/glaredb/issues/2168 for test failure

tychoish:
> See #2168 for test failure

yup yup!

PR #2216, opened 2023-12-06
chore: Fix error message for invalid copy-to

vrongmeal:
Simple update to error message.

### DIFF
--- a/crates/sqlexec/src/planner/errors.rs
+++ b/crates/sqlexec/src/planner/errors.rs
@@ -52,6 +52,11 @@ pub enum PlanError {
     #[error("Invalid alter statement: {msg}")]
     InvalidAlterStatement { msg: &'static str },
 
+    #[error("Invalid copy to statement: {source}")]
+    InvalidCopyToStatement {
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
+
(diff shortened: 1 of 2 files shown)


crates/sqlexec/src/planner/errors.rs
```diff
@@ -52,6 +52,11 @@ pub enum PlanError {
     #[error("Invalid alter statement: {msg}")]
     InvalidAlterStatement { msg: &'static str },
 
+    #[error("Invalid copy to statement: {source}")]
+    InvalidCopyToStatement {
+        source: Box<dyn std::error::Error + Send + Sync>,
```
tychoish, line 57: `ValidationError::CredentialsNotSupportedByDatasource`

this error is always going to a Validation error, seems like we don't need the generic here?

tychoish (approved):
lgtm modulo maybe type change.

crates/sqlexec/src/planner/errors.rs
vrongmeal: No, not really. Right now it is. But this is a flexible error to mark any invalid copy to statement. Something like `InvalidCopyToStatement(ValidationError)` implies that you can only have an invalid copy to statement due to a validation of options issue.

PR #2221, opened 2023-12-07
feat(build): add aarch64 release target

tychoish:
The `justfile` already has support. 

I'd definitely like someone with a bit more rust toolchain familiarty
to gut check my assumption that this will _just work_ on such an old
version of ubuntu.


scsmithr (commented):
If it compiles, I'm sure it'll work. Old version of ubuntu is because of glibc: https://github.com/GlareDB/glaredb/pull/1739

.github/workflows/release.yaml
```diff
@@ -18,7 +18,7 @@ jobs:
             target: x86_64-pc-windows-msvc
           - os: ubuntu-2004-8-cores
             target: x86_64-unknown-linux-gnu
-          - os: ubuntu-2004-8-cores
+          - os: ubuntu-2204-8-cores
```
scsmithr, line 21: We might need to create a machine type for this. The existing 8 core ones are custom large runners.

tychoish: how do we do that?

zhangping99 (approved):

tychoish:
Just to follow up, I think @scsmithr needs to add the updated builder to the GlareDB org to unblock this.

.github/workflows/release.yaml
scsmithr: Screenshot 2024-01-05 at 1 03 19 PM

runner added

tychoish:
Seems to be hitting a link error:

https://github.com/GlareDB/glaredb/actions/runs/7438394256/job/20237097364#step:6:1315

tychoish:
I've poked at this and my current assessment is that: there's some binary blob (or something) and/or it's falling back to using the wrong linker.

I can reproduce this locally but I haven't convinced cargo/rustc to do the right thing.

universalmind303:
marking this as draft as I don't think this is actively waiting on review

tychoish:
upstream dependencies are blocking this more long term, and our current plan is to address this by replacing them or making them optional. Going to close this and revisit later.

PR #2231, opened 2023-12-08
fix: cases sensitivity for statement options keywords

### LINKED ISSUE #2199: external table `location` keyword shouldn't be case sensitive.
## Description

the `location` property in the following is case sensitive. 

```sh
> ./target/release/glaredb local -q "CREATE EXTERNAL TABLE customer FROM LOCAL (LOCATION 'testdata/parquet/userdata1.parquet')"
Error: sql parser error: missing option: location
```

```sh
> ./target/release/glaredb local -q "CREATE EXTERNAL TABLE customer FROM LOCAL (location 'testdata/parquet/userdata1.parquet')"
Table created
```

## Expected behavior

both `location` and `LOCATION` work the same, as well as any other case insensitive variants

I want to write all of my queries in spongebob case

```sql
CrEaTe ExTeRnAl TaBlE supplier fRoM lOcAl (LoCaTiOn 'testdata/parquet/userdata1.parquet')
```

seve-martinez:
Closes #2199 .

Prevents casing issues when using the CLI by forcing statement option keys to be lowercase:
```console
smartinez in ~/src/glaredb on smartinez/location_casing_error λ cargo run --bin glaredb local -q "CrEaTe ExTeRnAl TaBlE supplier fRoM lOcAl (LoCaTiOn 'testdata/parquet/userdata1.parquet')"

Table created
```


seve-martinez:
@universalmind303 I had originally made this change in the parser logic, but that caused failures in the case-sensitive slt checks. Is #2199 only specific to CLi-based inputs?

universalmind303:
> @universalmind303 I had originally made this change in the parser logic, but that caused failures in the case-sensitive slt checks. Is #2199 only specific to CLi-based inputs?

This should definitely happen in the parser only for the `location` keyword. If we lowercase the entire query, it'll result in some undesired results

`select * from "MyTable"` should not be evaluated the same as `select * from "mytable"`. SQL is not case sensitive on  keywords, but identifiers are allowed to be case sensitive. the `sqlparser` crate already handles all of the non custom keywords, so we should just need to handle the parsing of the `location` keyword.

seve-martinez:
> > @universalmind303 I had originally made this change in the parser logic, but that caused failures in the case-sensitive slt checks. Is #2199 only specific to CLi-based inputs?
> 
> This should definitely happen in the parser only for the `location` keyword. If we lowercase the entire query, it'll result in some undesired results
> 
> `select * from "MyTable"` should not be evaluated the same as `select * from "mytable"`. SQL is not case sensitive on keywords, but identifiers are allowed to be case sensitive. the `sqlparser` crate already handles all of the non custom keywords, so we should just need to handle the parsing of the `location` keyword.

If i'm following the logic correctly, `location` is a type of option, specifically for the `CREATE EXTERNAL TABLE` protocol. It looks like that's handled here:
https://github.com/GlareDB/glaredb/blob/6e2cb0f53b58e3bacd6c5129e0a4c5d033306ae2/crates/sqlexec/src/parser.rs#L848

Is `location` the only type of option we'd like to handle? Something like
```rust
if key.to_lowercase() == "location" {
  key = "location";
}
```
Otherwise, is there somewhere that `location` is defined in a keyword Enum I haven't found yet?

seve-martinez:
> > @universalmind303 I had originally made this change in the parser logic, but that caused failures in the case-sensitive slt checks. Is #2199 only specific to CLi-based inputs?
> 
> This should definitely happen in the parser only for the `location` keyword. If we lowercase the entire query, it'll result in some undesired results
> 
> `select * from "MyTable"` should not be evaluated the same as `select * from "mytable"`. SQL is not case sensitive on keywords, but identifiers are allowed to be case sensitive. the `sqlparser` crate already handles all of the non custom keywords, so we should just need to handle the parsing of the `location` keyword.

It looks like anywhere that `remove_required` or `remove_optional` is called is specifically looking for a lowercase instance of its keys:
https://github.com/GlareDB/glaredb/blob/6e2cb0f53b58e3bacd6c5129e0a4c5d033306ae2/crates/sqlexec/src/planner/session_planner.rs#L596-L599
https://github.com/GlareDB/glaredb/blob/6e2cb0f53b58e3bacd6c5129e0a4c5d033306ae2/crates/sqlexec/src/planner/session_planner.rs#L1563
https://github.com/GlareDB/glaredb/blob/6e2cb0f53b58e3bacd6c5129e0a4c5d033306ae2/crates/sqlexec/src/planner/session_planner.rs#L2052-L2055

In that case, can the change be made in `remove_required/optional` method on `StmtOpts`?
https://github.com/GlareDB/glaredb/blob/a2d1c5f4e3588f5e37cef419116a95545f093a37/crates/sqlexec/src/parser/options.rs#L277

Or rather, is `location` itself enough to warrant an additional `OptionValue`?
https://github.com/GlareDB/glaredb/blob/a2d1c5f4e3588f5e37cef419116a95545f093a37/crates/sqlexec/src/parser/options.rs#L16

universalmind303:
I think it'd be safe to assume all keys inside `options` should be case insensitive. I'd expect it to follow SQL semantics where a key of `location` would be case insensitive, but a key of `"Location"` would preserve it's casing. Since we don't support quoted keys inside `options` anyways, I wouldn't worry too much about the latter case for now. 

I think we can just `.to_lowercase()` all of the keys before inserting into the object map.

universalmind303:
@seve-martinez PR looks good, can we add a slt _(sql logic tests)_  for various casings?

seve-martinez:
> @seve-martinez PR looks good, can we add a slt _(sql logic tests)_ for various casings?

I added some to the `external_table.slt` file. Let me know if those look okay.

tychoish (commented):
- the title of the pr seems like it should be updated as it's really all option keywords
- are there other or additional tests that might be useful (I think we're good, but I wanted to check).

universalmind303 (approved):

PR #2240, opened 2023-12-08
feat: allow for tests over flightsql

### LINKED ISSUE #2234: flightsql: add slts

universalmind303:
closes #2234 

There's still a bit of cleanup to do, but it seems to work. 

I'm planning on refactoring the test suites a bit, currently they always spin up a pg instance, which is not needed at all for the flightsql tests. 

My plan is to refactor them so the following commands are accepted

```sh
> just slt 'path/to/test' # default pg behavior
> just slt 'path/to/test' --protocol=rpc # equivalent to the current `just slt 'path/to/test' --rpc-test`
> just slt 'path/to/test' --protocol=flightsql # run the tests over the flightsql wire protocol
> just slt 'path/to/test' --protocol=postgres # run the tests over the postgres wire protocol. (same as the default)
```

---

12/12 updates. 

There are a few additional changes to the following files that I introduced in this PR worth highlighting. 

`crates/sqlexec/src/session.rs`
- rename a few methods to make the intent clearer
- add a few new helper methods for the flightsql client

`crates/glaredb/src/server.rs`
There were some bigger changes in here, namely to `ComputeServer`. To make the tests easier to run in isolation (pg/rpc/flightsql), I decided to refactor the compute server a bit to make the postgres server optional. 

Previously, we would always start up a postgres instance, now it is fully optional. You can specify what protocols you want to be exposed -- **TODO: cleanup the cli args for this** --

---

12/12 updates pt 2. 

This PR really got away from me.. A lot bigger than I had originally anticipated. _sorry reviewers_

Updated the CLI args to allow for postgres to be optional. Just to be safe, I added a bunch of additional tests to make sure nothing broke in the process. So there are some changes to the cli args to handle the new functionality. 

The only breaking change is to how you run the tests. `just slt 'path/to/test' --protocol=rpc` instead of `just slt 'path/to/test' --rpc-test`.


crates/glaredb/src/server.rs
```diff
@@ -144,9 +236,35 @@ impl ComputeServer {
...
+        if self.enable_simple_query_rpc {
+            info!("enabling simple query rpc service");
+            let handler = SimpleHandler::new(self.engine.clone());
+            server = server.add_service(SimpleServiceServer::new(handler));
+        }
+        server
+    }
     /// Serve using the provided config.
     pub async fn serve(self, conf: ServerConfig) -> Result<()> {
+        println!("Starting GlareDB");
```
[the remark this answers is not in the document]
universalmind303: this is still a draft PR, not ready for review yet.

crates/protogen/proto/metastore/catalog.proto
```diff
@@ -31,6 +31,8 @@ syntax = "proto3";
 
 package metastore.catalog;
 
+option go_package = "github.com/glaredb/cloud/pkg/protogen/metastore";
```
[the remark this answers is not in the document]
scsmithr: This pr has changes from a few other prs that were already merged.

I added these options in https://github.com/GlareDB/glaredb/pull/2241. I figured adding these in would make things simpler for cloud as we could just curl and protoc with minimal fuss. Also I have no idea out to set this option when invoking protoc.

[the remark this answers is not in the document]
universalmind303: this PR is stacked on https://github.com/GlareDB/glaredb/pull/2219, so i'll need to resolve all conflicts there before this shows a proper diff.

crates/metastore/src/local.rs
```diff
@@ -34,11 +34,13 @@ pub async fn start_inprocess(
...
     tokio::spawn(async move {
-        Server::builder()
+        if let Err(e) = Server::builder()
             .add_service(MetastoreServiceServer::new(Service::new(store)))
             .serve_with_incoming(futures::stream::iter(vec![Ok::<_, MetastoreError>(server)]))
             .await
-            .unwrap()
+        {
+            eprintln!("internal error: {}", e);
+        }
```
universalmind303, line 43: I originally added this when debugging the metastore issue mentioned on 12/11. I think it makes sense to have some logging if the handle panics. Happy to remove if we don't want it though.

crates/sqlexec/src/planner/logical_plan/create_schema.rs
```diff
@@ -24,7 +24,16 @@ impl UserDefinedLogicalNodeCore for CreateSchema {
...
+        write!(
+            f,
+            "CREATE SCHEMA{if_not_exists} {schema_reference}",
+            if_not_exists = if self.if_not_exists {
+                " IF NOT EXISTS"
+            } else {
+                ""
+            },
+            schema_reference = self.schema_reference
+        )
```
universalmind303, line 36: originally added these to help debugging. Makes sense to keep them as they are better fmt strings anyways.

crates/sqlexec/src/planner/logical_plan/create_table.rs
```diff
@@ -30,7 +30,17 @@ impl UserDefinedLogicalNodeCore for CreateTable {
     }
 
     fn fmt_for_explain(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}", Self::EXTENSION_NAME)
+        write!(
```
universalmind303, line 33: same here

crates/sqlexec/src/planner/logical_plan/create_temp_table.rs
```diff
@@ -30,7 +30,17 @@ impl UserDefinedLogicalNodeCore for CreateTempTable {
     }
 
     fn fmt_for_explain(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
-        write!(f, "{}", Self::EXTENSION_NAME)
+        write!(
```
universalmind303, line 33: and here

crates/sqlexec/src/planner/physical_plan/create_temp_table.rs
```diff
@@ -100,7 +100,7 @@ async fn create_temp_table(
         && !plan.or_replace
     {
         if plan.if_not_exists {
-            return Ok(RecordBatch::new_empty(Arc::new(Schema::empty())));
+            return Ok(new_operation_batch("create table if not exists"));
```
universalmind303, line 103: Our expected `create_table` schema didn't align with the original `new_empty` which returned an empty schema. Flightsql was erroring out due to schema mismatch.

crates/sqlexec/src/session.rs
```diff
@@ -740,25 +734,46 @@ impl Session {
         }
     }
 
-    pub fn parse_query(&mut self, query: &str) -> Result<VecDeque<StatementWithExtensions>> {
+    /// Create a logical plan from a SQL query.
+    /// if the query doesn't contain exactly one statement, an error is returned.
+    pub async fn create_logical_plan(&mut self, query: &str) -> Result<LogicalPlan> {
```
universalmind303, line 739: this was the cause of the metastore issue. I was creating the plan without first updating the state, so the tables "werent there" because the catalog state was outdated.

testdata/sqllogictests/functions/kdl.slt
```diff
@@ -2,8 +2,9 @@ statement ok
 create table cows (id int, docs string);
 
 statement ok
-
 insert into cows values (0, 'name "betsy" status="amazing" age=89; details "classic" a=1 b=100;');
+
+statement ok
 insert into cows values (1, 'name "zabby" status="electric" age=120; details "zesty" a=1 b=400;');
```
universalmind303, line 8: sql logic tests explicitly forbid multiple statements per block. FlightSQL couldn't properly process these transactions, and since it's not technically allowed in slt's, i just changed this to multiple statements.

testdata/sqllogictests/functions/postgres.slt
```diff
@@ -7,25 +7,28 @@
...
 ----
 glaredb
 
+skipif glaredb_flight
 query T
 select current_database();
 ----
 ${SLT_CURRENT_DATABASE}
 
+skipif glaredb_flight
```
universalmind303, line 31: these are "hacked" in to work with postgres, I wasnt able to figure out how to implement that over flightsql, so i just skipped these. Once we have tls/auth setup, we can revisit.

crates/glaredb/src/args/server.rs
```diff
@@ -75,12 +84,28 @@ pub struct ServerArgs {
...
+    #[arg(long, hide = true, value_parser)]
     pub enable_simple_query_rpc: bool,
 
     /// API key for segment.
     ///
     /// (Internal)
-    #[clap(long, hide = true, value_parser)]
+    #[arg(long, hide = true, value_parser)]
     pub segment_key: Option<String>,
+    /// Enable the experimental Flight SQL API.
```
scsmithr, line 95:
```suggestion

    /// Enable the experimental Flight SQL API.
```

crates/glaredb/src/args/server.rs
```diff
@@ -75,12 +84,28 @@ pub struct ServerArgs {
...
+    /// By default, the Flight SQL API will be enabled on port 6789.
+    /// Use the `--flight-bind` option to change the port.
+    #[arg(long, default_value="false", action = clap::ArgAction::SetTrue)]
+    pub enable_flight_api: bool,
+
+    /// Disable the default postgres api.
+    ///
+    /// This will fully disable the postgres server on port 6543.
+    #[arg(long, default_value="false", action = clap::ArgAction::SetTrue)]
+    pub disable_postgres_api: bool,
```
scsmithr, line 110: Why is this needed?

crates/glaredb/src/server.rs
```diff
@@ -8,40 +8,42 @@ use rpcsrv::flight_handler::{FlightServiceServer, FlightSessionHandler};
...
-pub struct ServerConfig {
+pub struct PostgresProtocolConfig {
     /// Listener to use for pg handler.
-    pub pg_listener: TcpListener,
-
-    /// Address to use for the rpc handler. If not provided, an rpc handler will
-    /// not be started.
-    pub rpc_addr: Option<SocketAddr>,
+    pub listener: TcpListener,
+    pub handler: Arc<ProtocolHandler>,
```
scsmithr, line 26: Document what this is. I don't know if `handler` is for the rpc stuff or something else.

crates/glaredb/src/server.rs
```diff
@@ -166,97 +213,130 @@ impl ComputeServerBuilder {
...
-                )
-                .await?,
-            )
+            let pg_handler = Arc::new(ProtocolHandler::new(engine.clone(), handler_conf));
+            Some(PostgresProtocolConfig {
+                listener,
+                handler: pg_handler,
+            })
+        } else {
+            println!("pg_listener not provided");
```
scsmithr, line 242:
```suggestion
            debug!("pg_listener not provided");
```

crates/glaredb/src/server.rs
```diff
@@ -352,11 +417,52 @@ impl ComputeServer {
...
+            rx.await.map_err(|_| anyhow!("shutdown error"))
         }
     }
 }
 
+fn spawn_shutdown_handler(
+    engine: Arc<Engine>,
+    is_integration_testing: bool,
+    tx: oneshot::Sender<()>,
+) {
```
scsmithr, line 433: Imo this should return a receiver instead of taking a sender.

crates/rpcsrv/src/flight_handler.rs
```diff
@@ -66,74 +68,77 @@ impl FlightSessionHandler {
...
+    ) -> Result<Arc<RwLock<Session>>, Status> {
+        // If the client specified a database id, use it, otherwise create a new one.
         let db_handle = request
             .metadata()
             .get(DATABASE_HEADER)
             .map(|s| s.to_str().unwrap().to_string())
             .unwrap_or_else(|| Uuid::new_v4().to_string());
 
         if self.sessions.contains_key(&db_handle) {
+            // println!("found session for {}", db_handle);
```
scsmithr, line 80:
```suggestion
```

crates/sqlexec/src/planner/logical_plan/create_schema.rs
scsmithr: 1. Explain formatting shouldn't be the query string equivalent. If we want more information in the output, it should be in the form of `CreateSchema [refere='my_schema', if_not_exists=true]`. See other explain outputs for examples.
2. How were you using this for debugging. A `explain create schema my_schema` doesn't work.

testdata/sqllogictests/functions/kdl.slt
```diff
@@ -36,12 +37,12 @@ select count(*) from cows where kdl_matches(cows.docs, '[status]');
 ----
 2
 
-query
+query I
```
scsmithr, line 40:
```suggestion
query T
```

testdata/sqllogictests/functions/kdl.slt
```diff
@@ -36,12 +37,12 @@ select count(*) from cows where kdl_matches(cows.docs, '[status]');
...
 select cows.docs from cows where kdl_matches(cows.docs, '[b=100]');
 ----
 name "betsy" status="amazing" age=89; details "classic" a=1 b=100;
 
-query
+query I
```
scsmithr, line 45:
```suggestion
query T
```

testdata/sqllogictests/functions/postgres.slt
scsmithr: We should open an issue for this.

crates/testing/tests/sqllogictests/main.rs
```diff
@@ -13,9 +13,9 @@ fn main() -> Result<()> {
         // Rust tests
         .test("sqllogictests/ssh_keys", Box::new(SshKeysTest))?
         .test("pgproto/binary_encoding", Box::new(PgBinaryEncoding))?
-        // Add hooks
+        // // Add hooks
```
scsmithr, line 16:
```suggestion
        // Add hooks
```

crates/testing/tests/sqllogictests/main.rs
```diff
@@ -13,9 +13,9 @@ fn main() -> Result<()> {
...
         .hook("*", Arc::new(AllTestsHook))?
-        // SSH Tunnels hook
+        // // SSH Tunnels hook
```
scsmithr, line 18:
```suggestion
        // SSH Tunnels hook
```

crates/rpcsrv/src/flight_handler.rs
```diff
@@ -51,7 +53,7 @@ pub struct FlightSessionHandler {
     // We'll want to implement a time based eviction policy, or a max size.
     // We use [`Session`] instead of [`TrackedSession`] because tracked sessions need to be
     // explicitly closed, and we don't have a way to do that yet.
-    sessions: DashMap<String, Arc<Mutex<Session>>>,
+    sessions: DashMap<String, Arc<RwLock<Session>>>,
```
scsmithr, line 56:
```suggestion
    sessions: DashMap<String, Arc<Mutex<Session>>>,
```

I believe pretty strongly in using mutexes over read/write locks when using tokio's implementations as it's assumed you'll be locking over await points. If I had my way, there would be no tokio mutexes in the code base as a believe it points to a suboptimal design, but I also know it's not practical to change it.

Session access is typically serial (in that clients call one method after another and not concurrently) and a normal mutex would have lower overhead in that case. Also tokio mutexes are already slow, and tokio rwlocks are even slower.

scsmithr:
> Previously, we would always start up a postgres instance [...]

We should make sure to be clear what we mean by this. We're starting up a server implementing the pg protocol. By saying "postgres instance", that can lead to confusion as we're not starting up a postgres instance.

crates/glaredb/src/args/local.rs
```diff
@@ -1,12 +1,15 @@
+use clap::Args;
+
 use super::*;
 
-#[derive(Parser, Debug)]
+#[derive(Args, Debug)]
+#[group(id="query_input", args = ["query", "file"], multiple = false)]
```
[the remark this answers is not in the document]
universalmind303: I can revert this as it's not really relevant to this PR, Just something i found while testing to make sure i didnt break any of the flags. 

Clap has builtin macros for mutually exclusive arguments that we were previously checking manually. They give much nicer error messages, so i decided to throw this in here.

crates/sqlexec/src/planner/logical_plan/create_schema.rs
universalmind303: that makes sense. I'll revert these for now. 

> How were you using this for debugging. A explain create schema my_schema doesn't work.
I was just running the create table.slt, and there were several `create table` and `create schema` operations, I couldnt really tell which if any were failing, so I added these to try to narrow down which ones were failing.

crates/glaredb/src/args/server.rs
universalmind303: i don't think it's needed, but it is subjectively useful. It takes up resources spinning up the postgres api, if users have no need for it, i feel like they should have a way to disable it.

testdata/sqllogictests/functions/kdl.slt
universalmind303: are there any docs on what the different `query <VALUE>` arguments are? I've always just copypasta'd other examples.

universalmind303:
> > Previously, we would always start up a postgres instance [...]
> 
> We should make sure to be clear what we mean by this. We're starting up a server implementing the pg protocol. By saying "postgres instance", that can lead to confusion as we're not starting up a postgres instance.

yeah sorry, i meant a postgres compatible api here.

testdata/sqllogictests/functions/kdl.slt
scsmithr: > The characters codes are "T" for a text result, "I" for an integer result, and "R" for a floating-point result. 

From https://www.sqlite.org/sqllogictest/doc/trunk/about.wiki

We had in the contributing doc at point, but seems like it was removed.

scsmithr (approved):

Issue #2242, opened 2023-12-09
Simple query stream doesn't EOF properly


scsmithr:
## Description

https://github.com/glaredb/glaredb/blob/bd684254b29fcff0e4bcafd8f4deab1ae1a61d59/crates/rpcsrv/src/handler.rs#L378-L416

Causes context deadline exceeded on cloud.

universalmind303:
@scsmithr I wonder if it'd be easier to just use the go flightsql driver for cloud? 


The following works on [this branch](https://github.com/GlareDB/glaredb/pull/2219)
```go
package main

import (
	"database/sql"
	_ "github.com/apache/arrow-adbc/go/adbc/sqldriver/flightsql"
)

func main() {
	db, err := sql.Open("flightsql", "uri=grpc://0.0.0.0:6789")
	if err != nil {
		panic(err)
	}
	defer db.Close()

	if err = db.Ping(); err != nil {
		panic(err)
	}

	rows, err := db.Query("SELECT registration_dttm from './testdata/csv/userdata1.csv' limit 1")
	if err != nil {
		panic(err)
	}
	columns, err := rows.Columns()
	if err != nil {
		panic(err)
	}

	println(columns[0])

}
```

scsmithr:
This fix here is actually super easy.

We can switch to flight sql once it's ready. The code being written in https://github.com/GlareDB/cloud/pull/2527 would still be needed--it would be easy to swap the client out whenever.

PR #2243, opened 2023-12-10
fix: Properly exit SimpleExecuteQueryStream on stream end

scsmithr:
Closes https://github.com/GlareDB/glaredb/issues/2242

Also moves the simple stuff to its own module to split up the code a little bit.


crates/rpcsrv/src/simple.rs
```diff
@@ -0,0 +1,162 @@
...
+
+        loop {
+            match self.inner.poll_next_unpin(cx) {
+                // Drop the result, we're not sending it back to the client.
+                // And continue to the next loop iteration.
+                Poll::Ready(Some(Ok(_))) => (),
+
+                // Stream completed without error, return success to the client.
+                Poll::Ready(None) => {
+                    self.done = true; // Make sure we properly signal stream end on next poll.
```
scsmithr, line 109: This is the fix. Adds a bool so that next poll of this stream (no the inner stream) we exit.

universalmind303 (approved):

Issue #1807, opened 2023-09-22
better float formatting


universalmind303:
## Description
tpch q3 produces some long floats. we should probably truncate them when displaying, or have some config param to specify how to display floats

```sql
    select
        l_orderkey,
        sum(l_extendedprice * (1 - l_discount)) as revenue,
        o_orderdate,
        o_shippriority
    from
        customer,
        orders,
        lineitem
    where
        c_mktsegment = 'BUILDING'
        and c_custkey = o_custkey
        and l_orderkey = o_orderkey
        and o_orderdate < '1995-03-15'
        and l_shipdate > '1995-03-15'
    group by
        l_orderkey,
        o_orderdate,
        o_shippriority
    order by
        revenue desc,
        o_orderdate
    limit 10
```

```
┌────────────┬────────────────────┬─────────────────────┬────────────────┐
│ l_orderkey │            revenue │ o_orderdate         │ o_shippriority │
│         ── │                 ── │ ──                  │             ── │
│      Int64 │            Float64 │ Timestamp[µs, UTC]  │          Int64 │
╞════════════╪════════════════════╪═════════════════════╪════════════════╡
│    2456423 │        406181.0111 │ 1995-03-05T00:00:00 │              0 │
│    3459808 │ 405838.69889999996 │ 1995-03-04T00:00:00 │              0 │
│     492164 │         390324.061 │ 1995-02-19T00:00:00 │              0 │
│    1188320 │        384537.9359 │ 1995-03-09T00:00:00 │              0 │
│    2435712 │ 378673.05580000003 │ 1995-02-26T00:00:00 │              0 │
│    4878020 │        378376.7952 │ 1995-03-12T00:00:00 │              0 │
│    5521732 │        375153.9215 │ 1995-03-13T00:00:00 │              0 │
│    2628192 │ 373133.30939999997 │ 1995-02-22T00:00:00 │              0 │
│     993600 │        371407.4595 │ 1995-03-05T00:00:00 │              0 │
│    2300070 │  367371.1452000001 │ 1995-03-13T00:00:00 │              0 │
└────────────┴────────────────────┴─────────────────────┴────────────────┘
```

vrongmeal:
We do truncate floats, see: https://github.com/GlareDB/glaredb/blob/44efb74e32ef0a3cebca5c767108fe49e7ac856d/crates/repr/src/str/encode.rs#L34-L94

We can use these encoders for displaying results instead of `to_string` for all the data types. I can take this up.

PR #2245, opened 2023-12-11
Formatting : Truncating long floats

hemanth94:
Fixes the issue #1807 

I made 4 digits after decimal point as default display config for long floats.
This code only deals with floating numbers.
Should i add testcases as well? Please suggest.

Please correct me if needed, Thanks.

### DIFF
--- a/crates/arrow_util/src/pretty.rs
+++ b/crates/arrow_util/src/pretty.rs
@@ -237,6 +237,8 @@ fn truncate_or_wrap_string(s: &mut String, width: usize) {
         return;
     }
 
+    fmt_float(s); //Truncating long floats to have 4 digits after decimal point.
+
     if display_width(s) <= width {
         return;
     }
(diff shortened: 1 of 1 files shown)


hemanth94:
@universalmind303 I tried to code without using formatter, but i think floats cannot be easily extracted  from `ArrayData`, Which is underlying data form of `col: &dyn Array` in code. 

I think truncating floats using Output string and `Arrayformatter` would be much easier.
We can check for dtype by `col.data_type()` method, and later use Output String from formatter to truncate it in a different function.

Please suggest if I'm missing something. Thanks.

universalmind303:
> @universalmind303 I tried to code without using formatter, but i think floats cannot be easily extracted from `ArrayData`, Which is underlying data form of `col: &dyn Array` in code.
> 
> I think truncating floats using Output string and `Arrayformatter` would be much easier. We can check for dtype by `col.data_type()` method, and later use Output String from formatter to truncate it in a different function.
> 
> Please suggest if I'm missing something. Thanks.

You should be able to get the dtype from `&dyn Array`, then you can safely downcast to `Float64Array`. The arrow-array lib has a section on downcasting.

https://docs.rs/arrow-array/47.0.0/arrow_array/index.html#downcasting-an-array

should end up roughly looking like this
```rust
    fn try_new_from_array(col: &dyn Array, trunc: Option<usize>) -> Result<Self, ArrowError> {
        let vals = match col.data_type() {
            DataType::Float64 => {
                // safe to unwrap as we've already checked the type. 
                let floats = col.as_any().downcast_ref::<Float64Array>().unwrap(); 
                todo!("do some formatting of floats")
            }
            _ => {
                
                // normal handling using ArrayFormatter
                
            }
        };

        Ok(ColumnValues { vals })
    }
```

hemanth94:
@universalmind303 Thanks a lot for the suggestion. I was unaware of downcasting to `Float64Array`.

PR #2249, opened 2023-12-12
fix: some broken tests

universalmind303:
Turns out we weren't actually running any of the tests in `glaredb/tests`. This fixes some issues that were identified when running those tests. 

Additionally updates the unit test command to properly run all unit tests.

### DIFF
--- a/crates/glaredb/src/bin/main.rs
+++ b/crates/glaredb/src/bin/main.rs
@@ -51,16 +51,24 @@ fn main() -> Result<()> {
         None => Commands::Local(cli.local_args),
     };
 
-    match &command {
-        Commands::Local(args) => {
-            if let Some(file) = &args.log_file {
-                logutil::init(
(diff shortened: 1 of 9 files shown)


crates/glaredb/src/bin/main.rs
```diff
@@ -51,16 +51,24 @@ fn main() -> Result<()> {
...
+                ..
+            }),
+            _,
+            _,
+        ) => logutil::init(
+            cli.verbose,
+            // Use JSON logging by default when writing to a file.
+            cli.log_mode.unwrap_or(LoggingMode::Json).into(),
+            Some(log_file),
+        ),
```
universalmind303, line 68: Since we weren't actually running the tests in this crate, there was a regression in https://github.com/GlareDB/glaredb/pull/2187 that went unnoticed.

scsmithr (approved):

PR #2253, opened 2023-12-13
fix: truncate long floats in cli formatting for display

hemanth94:
Fixes the issue #1807 
Tried to truncate long floats to have 5 digits after decimal point.
Added Unit Test for a batch of floats. 
@universalmind303 I abandoned the previous logic of having a config parameter to control number of digits after decimal point, instead hardcoded it. Please suggest any corrections.

Thanks in advance.

### DIFF
--- a/crates/arrow_util/src/pretty.rs
+++ b/crates/arrow_util/src/pretty.rs
@@ -576,7 +576,7 @@ fn fmt_floats(flt: Option<&f64>) -> String {
     match flt {
         Some(float) => {
             //Change here to configure digits after decimalpoint for long floats
-            let str = format!("{:.4}", float); 
+            let str = format!("{:.5}", float); 
             str
         }
         None => String::from(""),
--- a/crates/arrow_util/src/pretty.rs
+++ b/crates/arrow_util/src/pretty.rs
@@ -1,6 +1,7 @@
 use comfy_table::{Cell, CellAlignment, ColumnConstraint, ContentArrangement, Table};
-use datafusion::arrow::array::Array;
-use datafusion::arrow::datatypes::{DataType, Field, Schema, TimeUnit};
+use datafusion::arrow::array::{Array, Float64Array};
+use datafusion::arrow::datatypes::DataType::{self,Float64};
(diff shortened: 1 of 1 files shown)


crates/arrow_util/src/pretty.rs
```diff
@@ -505,15 +505,36 @@ struct ColumnWidthSizeStats {
...
+                let floats_array = col
+                    .as_any()
+                    .downcast_ref::<Float64Array>()
+                    .unwrap()
+                    .values();
+
+                (0..floats_array.len())
+                    .map(|idx| {
+                        let s = fmt_floats(floats_array.get(idx));
+                        Ok(s)
```
tychoish, line 520: do you need to declare `s` or can you just `Ok(fmt_floats(<...>))`

crates/arrow_util/src/pretty.rs
```diff
@@ -548,6 +569,17 @@ impl ColumnValues {
     }
 }
 
+fn fmt_floats(flt: Option<&f64>) -> String {
+    match flt {
+        Some(float) => {
+            //Change here to configure digits after decimalpoint for long floats
+            let str = format!("{:.5}", float);
+            str
```
tychoish, line 577: do you need to declare `str` here or can you just return the output of `format`?

tychoish (commented):
why did you pick 5? should we indicate somehow that we've truncated the numbers in formatting?

crates/arrow_util/src/pretty.rs
hemanth94: Yes I agree, My bad. I could have just written Ok(fmt_floats(<...>)). I will change this.

crates/arrow_util/src/pretty.rs
hemanth94: Same issue with my code here, Will fix it.

hemanth94: I saw the sample floats mentioned in issue #1807  to pick 5 digits after decimals, I don't have a solid reason tbh.
I was hoping someone from the team would guide me as to what is required.

universalmind303 (approved):

Issue #2255, opened 2023-12-14
pg proxy: client-side socket close detection


greyscaled:
```json
{
  "jsonPayload": {
    "threadId": "ThreadId(2)",
    "message": "client disconnected with error.",
    "filename": "crates/glaredb/src/pg_proxy.rs",
    "target": "glaredb::pg_proxy",
    "level": "DEBUG",
    "e": "unexpected end of file",
    "line_number": 88,
    "threadName": "pgsrv-thread-0",
  },
  "severity": "DEBUG",
}
```

caused by liveness probe pinging pgsrv

https://github.com/GlareDB/glaredb/blob/691db7c2bf4faca7bc5828ba97e8948c9e6aa5d7/crates/glaredb/src/pg_proxy.rs#L71-L94

PR #2256, opened 2023-12-14
chore: Detect client disconnect in pg proxy

scsmithr:
- Lowers log level for unexpected eof errors (client disconnects).
- Raises log level for all other errors

Closes https://github.com/GlareDB/glaredb/issues/2255

### DIFF
--- a/crates/glaredb/src/pg_proxy.rs
+++ b/crates/glaredb/src/pg_proxy.rs
@@ -1,13 +1,15 @@
 use anyhow::{anyhow, Result};
+use pgsrv::errors::PgSrvError;
 use pgsrv::proxy::ProxyHandler;
 use pgsrv::ssl::SslConfig;
 use proxyutil::cloudauth::CloudAuthenticator;
+use std::io;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::Arc;
 use tokio::net::TcpListener;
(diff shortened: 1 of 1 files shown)


greyscaled (approved):

greyscaled:
can we mark this as closing the issue #2255 ?

scsmithr:
> can we mark this as closing the issue #2255 ?

done

crates/glaredb/src/pg_proxy.rs
```diff
@@ -85,7 +87,9 @@ impl PgProxy {
                         debug!("client connected (proxy)");
                         match handler.proxy_connection(inbound).await {
                             Ok(_) => debug!("client disconnected"),
-                            Err(e) => debug!(%e, "client disconnected with error."),
+                            Err(PgSrvError::Io(e)) if e.kind() == io::ErrorKind::UnexpectedEof =>
+                                trace!("client disconnected"),
```
tychoish, line 91: should the `Ok()` case also be a trace then? should we distinguish between the two?

scsmithr: will add

scsmithr: Downgraded debugs to trace, and used different messages in both cases (in case we need to distinguish).

PR #2258, opened 2023-12-15
feat: high level integration testing framework

tychoish:
This is (obviously) a very high level pytest harness for running
process and system integration tests, and the CI and developer
experience (just) integration. There's nothing special about this
pytest development, and I expect that it can grow with us as needs
evolve.

There are a class of tests that are difficult to run in the current
testing enviornments that this harness should make easy and
trivial. Such as:

- cases that run a glaredb binary.

- cases that involve the orchestration of multiple glaredb binaries. 

- operations that require introspection or manipulation of the
  filesystem.

This shouldn't replace any of the existing tests:

- use rust unittests for as much of the internals as possible.

- anything that involves the python bindings should be written (also
  using pytest) in `bindings/python/tests`). This package should
  _never_ depend on the python bindings for build efficency reasons.
  
- anything case that can be entirely expressed in SQL should be
  written as a `sqllogictest` (SLT). 
  
- anything that requires the test of a specific service level
  dependency (as in for data-sources) should run in the SLT
  integration tests.


.github/workflows/ci.yaml
```diff
@@ -16,89 +16,85 @@ jobs:
...
+        name: cache
         with:
           path: |
             ~/.cargo/
             target/
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
 
-      - name: Build
+      - name: build
         run: just build
```
greyscaled, line 31: musing: I have a preference to not name these steps, since they're 1-liner commands we know and often run. The idea here is that `build` is no better than `just build`, while adding an alias/mapping layer.

I had reflexively gotten into a habit of always naming each step, as I thought you had to, but it turns out that if you only provide `- run: just build` the name will also be `just build`.

This is opinionated and preferential, and I'm making this comment just for sharing, and in the case that it feels helpful in this repo.

.github/workflows/ci.yaml
```diff
@@ -109,98 +105,133 @@ jobs:
...
+      - uses: extractions/setup-just@v1
+      - name: install poetry
+        uses: snok/install-poetry@v1
+      - name: install python
+        uses: actions/setup-python@v5
+        with:
+            python-version: '3.11' 
+            cache: poetry
+            cache-dependency-paths: |
+              tests/poetry.lock
```
greyscaled, line 180: musing: something I've found recently with our npm bindings tests, is that the dependency overhead is so small that there's no benefit to caching, and in fact it created a bug when run infrequently and upgrading to modern yarn. I don't think that's a concern here, and I don't know the dependency overhead of these tests. In short, I'd consider not adding this until needed, unless it already is!

greyscaled (commented):
This looks great to me

.github/workflows/ci.yaml
tychoish: recompiling the c-parts of the drivers is enough (and change rarely enough) that I wouldn't be super worried about cache bugs or lack of utility here. 

I also think it does caching regardless, but we miss it by having things scoped to the directory rather than the repo.

.github/workflows/ci.yaml
tychoish: yeah, makes sense. I did a bunch of clean up here just to lowercase things, and I feel less strongly.

tests/fixtures/glaredb.py
```diff
@@ -0,0 +1,51 @@
...
+                ":".join(addr),
+            ],
+            cwd=tmpdir,
+            close_fds="posix" in sys.builtin_module_names,
+        ) as p:
+            time.sleep(0.5)
+            assert not p.poll(), p.stdout.read().decode("utf-8")
+            conn = psycopg2.connect(host=addr[0], port=addr[1], dbname="glaredb")
+            conn.autocommit = True
+            yield conn
```
scsmithr, line 49: We should open an issue on the autocommit thing since it's unexpected that it wouldn't work without setting it to true.

scsmithr (approved):
We'll want to add `__pycache__` to gitignore too.

tests/fixtures/glaredb.py
tychoish: #2273

PR #2272, opened 2023-12-15
chore(pg proxy): update cloud cert

greyscaled:
This removes extraneous cert generation for pg proxy. The test was updated to use certs with the ED25519 algorithm. I didn't feel it was important enough to source control cert generation, since there's no need to rotate these in tests. For posterity, I used:

```shell
openssl genpkey -out key.pem -algorithm ED25519
openssl x509 -new -key key.pem -subj "/C=US/ST=Texas/L=Austin/O=GlareDB/OU=Engineering/CN=glaredb.com"
```

Resolves: #2270


Dockerfile
```diff
@@ -4,23 +4,16 @@ FROM rust:1.73-bookworm AS builder
 WORKDIR /usr/src/glaredb
 COPY . .
 
-RUN apt-get update && apt-get install -y openssl ca-certificates
```
greyscaled, line 7: (review): I'm expecting that we only needed these to run the openssl x509 gen. If a build dependency relies on `ca-certificates`, I'd want to know why (and to ensure it's documented).

Dockerfile
```diff
@@ -4,23 +4,16 @@ FROM rust:1.73-bookworm AS builder
...
 # Build release binary.
 RUN just build --release
 
-# Generate certs.
-RUN ./scripts/gen-certs.sh
-
 FROM debian:bookworm-slim
 
 # Runtime deps.
 RUN apt-get update -y && apt-get install -y openssl ca-certificates openssh-client
```
greyscaled, line 14: I believe our tunnels feature relies on these.

Dockerfile
scsmithr: If it builds without it, then probably fine. I could see trying to fetch the protoc binary potentially fail without ca certs.

Dockerfile
scsmithr: Yup, we'll want to keep these in.

Dockerfile
greyscaled: CI build step passed, and I succesfully `docker build .` locally.

crates/pgsrv/src/ssl.rs
```diff
@@ -148,54 +148,21 @@ mod tests {
...
-klokrmYS4DhlIHrSnJTfpPymPhqsSVHd97Jr20nbuBbXlRsCAwEAAaNTMFEwHQYD
-VR0OBBYEFHMbgh0q1cxHVpR0UCWq92+0zQrsMB8GA1UdIwQYMBaAFHMbgh0q1cxH
-VpR0UCWq92+0zQrsMA8GA1UdEwEB/wQFMAMBAf8wDQYJKoZIhvcNAQELBQADggEB
-AEJ554uXpkgXLGECVueikjzGOyCb7aBmKYdssuutWpEGM3pZpyCKzz/tsanULH3Q
-kNu5/eJYc0BBI2YXz1eIFrJR3fM+qbpIvP1j67CLhsjtvKSQTFEvqseCCZ1l34WK
-dNzpwf0HMB5MAt6T8xeQ5gDRNo7C3HuKT/8WouaIdn9H2UEluQBDM5Yg+cpDWOlv
-Dr/IzcLC/pyEmx7spkPfvxebtY64kdFh8qyl9iWZLO9wrpaWY1vV41sjZ1F41AgY
-KaRWGJhHwhyDcYzelsn0Ew2LdB7WfFQZfZe9EHJTj7bd6y6AbkduZ31oebXYlOzo
-1DkrU2K5aMlVF/vExWpfChk=
+MIIBkzCCAUUCFGipMcv8Oq6O89V+OkbybaF4q3XaMAUGAytlcDBsMQswCQYDVQQG
```
tychoish, line 151: what kind of certificate is this? and where did it come from (I know it's in a test)

tychoish (approved):

crates/pgsrv/src/ssl.rs
tychoish: oh I see.

PR #2276, opened 2023-12-18
chore: add missing lifetimes

tychoish:
was getting compiler warnings, wanted to silenece.


universalmind303 (changes requested):
[redundant_static_lifetimes](https://rust-lang.github.io/rust-clippy/rust-1.74.0/index.html#/redundant_static_lifetimes) has been around for a while. These changes get flagged for this linting rule when running locally, maybe you're on an older version of clippy?

PR #2278, opened 2023-12-18
chore: isolate execution of snowflake integration tests

tychoish:


.github/workflows/ci.yaml
```diff
@@ -356,3 +333,56 @@ jobs:
           # Fake GCS server with a sub-directory path; run with two different folder paths to assert no conflicts arise
           just slt -l gs://$TEST_BUCKET/path/to/folder/1 -o service_account_path=/tmp/fake-gcs-creds.json 'sqllogictests_native/*'
           just slt -l gs://$TEST_BUCKET/path/to/folder/2 -o service_account_path=/tmp/fake-gcs-creds.json 'sqllogictests_native/*'
+
+  service-integration-tests-snowflake:
```
universalmind303, line 337: I think we should just allow this one to fail. 

Reference: 
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error

tychoish: Then why even bother running the test? If you run a test that don't look at the output of and that can't fail, then it provides no value and we should just skip it entirely. 

A few things: 
- this is no longer in the required builders for tests.
- there's nothing that depends on this
- the changes to the concurrency settings (should) mean that only one of these tasks can run at once, which will (once all PRs will remediate the underlying cause of the failure (e.g. that too many tests are running at once, according to their error message, which agrees informally with what we see.) 

I think we can add this in later, but I'd like to give this a week to settle in and see if it works in practice.

scsmithr (approved):

PR #2279, opened 2023-12-19
feat: encrypted flight api via cloud proxy

### LINKED ISSUE #2235: flightsql: add cloud support

### LINKED ISSUE #2238: flightsql: add tls support

universalmind303:
closes #2238 
closes #2235


crates/rpcsrv/src/flight/handler.rs
```diff
@@ -140,11 +99,57 @@ impl FlightSessionHandler {
...
+        let db_id = request
+            .metadata()
+            .get(DATABASE_HEADER)
+            .map(|s| Uuid::try_parse_ascii(s.as_bytes()).ok())
+            .flatten()
+            .unwrap_or_else(|| Uuid::new_v4());
+
+        let session_vars = SessionVars::default()
+            .with_database_id(db_id, datafusion::variable::VarType::System)
+            .with_force_catalog_refresh(true, datafusion::variable::VarType::System);
```
tychoish, line 134: is this a change over previous behaviors? 

I guess my question is really:

- how often do we create a new session in flight?
- does this mean we're forcing a catalog refresh more often now?

crates/rpcsrv/src/flight/proxy.rs
```diff
@@ -0,0 +1,219 @@
...
+                "Auth type not implemented: {authorization}"
+            )))?;
+        }
+        let base64 = &authorization[basic.len()..];
+        let bytes = BASE64_STANDARD
+            .decode(base64)
+            .map_err(|_| Status::internal("authorization not decodable"))?;
+        let s =
+            String::from_utf8(bytes).map_err(|_| Status::internal("authorization not parsable"))?;
+        let parts: Vec<_> = s.split(':').collect();
```
tychoish, line 78: the proxy authenticates the session and then passes the authorization information (as key/value pairs, which the server's API should accept. I don't think there's much utility in implementing basic authentication as such, but maybe I'm missing something. 

The (a?) goal state, I think, for this, is:

- the proxy will authenticate with cloud, which will verify a connection with some sort of out-of-bad authentication,
- cloud will generate two certificates:
  - a user certificate, good for a short period of time (circa an hour) which will contain the user/database ID, and will, allow the proxy to (re) authenticate users without communicating with cloud as often. when the cert expires the user will need to get a new cert via either the original password bootstrapping, or we can add a certificate renewal mechanism.
  - a IPC cert, good for much longer (months), which will contain only the DB name and allow the proxy to authenticate to the database (glaredb). 
- glaredb will validate that the IPC cert has been signed by the certificate authority, but will not enforce authorization beyond that. to do this glaredb will receive a copy of CA's cert so that it can validate certificates

Presently, the proxy controls all access to the database workers, and communicates with cloud to ensure that all sessions are validated and correct. Behind the proxy, all connections are trusted, and the proxy has the ability to ensure that authorization information (headers) are correct before passing them to the database workers. Cloud authenticates sessions and hands clients a valid but non-identifying certificate that is used for encryption _only_.

As a fallback, and before before IPC certs are implemented; when the database server does not have a copy of the CA, (as in today's operation) it should fall back to looking for the authorization information in the metadata map. these headers are (I think?) possible to add in at the proxy level (using information from the session authentication (and eventually the user's cert.)

tychoish (commented):
this looks great. 

one question about basic and another session refreshing.

crates/rpcsrv/src/flight/proxy.rs
universalmind303: I read this a few times, and I don't really understand if this is a question, request for change, or a comment?

crates/rpcsrv/src/flight/handler.rs
universalmind303: should be the same number of sessions, just passing in the db id from cloud proxy now. The # of sessions is unchanged. 

> does this mean we're forcing a catalog refresh more often now?

it should be unchanged.

crates/rpcsrv/src/flight/proxy.rs
tychoish: I wanted to provide some context as to the current state and future of authentication at the glaredb proxy and the glaredb workers. 

I do not think that we should provide basic-style auth headers and instead have the proxy add/set the dbname/etc. in the headers that it sends back to the workers.

crates/rpcsrv/src/flight/handler.rs
```diff
@@ -140,11 +99,56 @@ impl FlightSessionHandler {
...
+        if self.sessions.contains_key(&conn_key) {
+            let sess = self.sessions.get(&conn_key).unwrap().clone();
+            return Ok(sess);
+        }
+
+        let db_id = request
+            .metadata()
+            .get(DATABASE_HEADER)
+            .and_then(|s| Uuid::try_parse_ascii(s.as_bytes()).ok())
+            .unwrap_or_else(Uuid::new_v4);
```
scsmithr, line 130: This shouldn't be returning a new uuid ever.

crates/rpcsrv/src/flight/handler.rs
```diff
@@ -140,11 +99,56 @@ impl FlightSessionHandler {
...
+
+        let session_vars = SessionVars::default()
+            .with_database_id(db_id, datafusion::variable::VarType::System)
+            .with_force_catalog_refresh(true, datafusion::variable::VarType::System);
+
+        let sess = self
+            .engine
+            .new_untracked_session(session_vars, SessionStorageConfig::default())
```
scsmithr, line 138: Shouldn't use default here, just creates in memory storage.

crates/rpcsrv/src/flight/handler.rs
universalmind303: so we just shouldn't set the `db_id` if its not available?

scsmithr: If it's not available, we should be erroring.

If we're worried about the local case, this should be the nil uuid and not a random one.

universalmind303: > If we're worried about the local case, this should be the nil uuid and not a random one.

got it. That's the case I was trying to handle with the `new`. Will just set it to `nil` instead.

crates/proxyutil/src/cloudauth.rs
```diff
@@ -63,12 +65,12 @@ impl ServiceProtocol {
 /// Params used for cloud authentication.
 #[derive(Debug, Clone)]
 pub struct AuthParams<'a> {
-    pub user: &'a str,
-    pub password: &'a str,
-    pub db_name: &'a str,
+    pub user: Cow<'a, str>,
+    pub password: Cow<'a, str>,
+    pub db_name: Cow<'a, str>,
```
vrongmeal, line 70: Why are we using a `Cow` here? Just curious, unable to infer this by glancing over the code.

crates/rpcsrv/src/flight/handler.rs
```diff
@@ -140,11 +100,68 @@ impl FlightSessionHandler {
             .map_err(Status::from);
         Ok(Response::new(Box::pin(stream)))
     }
+
+    pub fn new(engine: &Arc<Engine>) -> Self {
+        Self {
+            engine: engine.clone(),
```
vrongmeal, line 106:
```suggestion
    pub fn new(engine: Arc<Engine>) -> Self {
        Self {
            engine,
```

Just a minor nitpick. I prefer passing in owned values (if required) inside the function arguments. It avoids the `clone`.

crates/rpcsrv/src/flight/handler.rs
```diff
@@ -218,6 +235,7 @@ impl FlightSqlService for FlightSessionHandler {
         cmd: CommandPreparedStatementQuery,
         req: Request<FlightDescriptor>,
     ) -> Result<Response<FlightInfo>, Status> {
+        println!("get_flight_info_prepared_statement: {:?}", cmd);
```
vrongmeal, line 238: Remove this later?

vrongmeal (approved):
Just a few tiny comments but looks good to me.

crates/proxyutil/src/cloudauth.rs
universalmind303: The original struct was all `&'a str`. Now with the auth implementation for flight, we need to operate on some owned values. So instead of unnecessarily converting everything to an owned value, just used Cow so we can use owned & borrowed values.

vrongmeal: Oh yeah, that makes sense

crates/rpcsrv/src/flight/proxy.rs
universalmind303: > why are we implementing http basic auth on this endpoint? 

it's really just from an ergonomic/ usability standpoint. The connectors usually expect an `AUTHORIZATION` header, and it feels weird bypassing this to just stick them into a different header when it's the same information.

```py
conn = connect("grpc+tls://0.0.0.0:6789", db_kwargs={
    DatabaseOptions.AUTHORIZATION_HEADER.value: "Basic <b64_encoded>",
    f"{DatabaseOptions.RPC_CALL_HEADER_PREFIX.value}db_name": "my-db",
    f"{DatabaseOptions.RPC_CALL_HEADER_PREFIX.value}org": "public",
})
```
vs
```py
conn = connect("grpc+tls://0.0.0.0:6789", db_kwargs={
    f"{DatabaseOptions.RPC_CALL_HEADER_PREFIX.value}db_name": "my-db",
    f"{DatabaseOptions.RPC_CALL_HEADER_PREFIX.value}org": "public",
    f"{DatabaseOptions.RPC_CALL_HEADER_PREFIX.value}user": "admin",
    f"{DatabaseOptions.RPC_CALL_HEADER_PREFIX.value}password": "admin",
})
```

> from my understanding, flightsql should (operationally) have the same security model as the RPC service. is this not the case?

yes the security model is exactly the same, we just pass in the user/pass through the `Authorization` header instead of custom headers.

tychoish: - the connectors are going to have to deal with adding the extra headers anyway?
- the `db_name` is the UUID of the deployment (correct?)
- does glaredb look at anything (when in permissive mode?) other than the db name/org name?
- the proxy sets the database name/id?

universalmind303: > the connectors are going to have to deal with adding the extra headers anyway?

yes but it's better to follow convention and use the already provided auth header instead of custom headers. We have to use custom headers for `db_name` and `org`, but there are already provided header for user/pass. It feels wrong to use our own headers for a public facing api. RPC is internal so it doesn't really matter, but IMO, the public facing one should try to follow convention as much as possible. We shouldn't follow RPC just because that's the way it was done, we should instead follow convention, and conventional api's will use basic auth header for passing user/pass.  

 
> the db_name is the UUID of the deployment (correct?)

no it's the vanity name 

> does glaredb look at anything (when in permissive mode?) other than the db name/org name?

don't know what you mean by permissive mode. 

> the proxy sets the database name/id?

It set's it on the request headers if that's what you mean.

tychoish (approved):
this seems good, 

I think we don't have a mechanism to get the certificate details to callers, which I think they'll need, unless there's some mechanism I'm missing, ore as in: https://github.com/glaredb/glaredb/blob/flightsql-tls/crates/sqlexec/src/remote/client.rs#L195

PR #2281, opened 2023-12-19
fix: Don't parse version-hint as a number for iceberg tables

### LINKED ISSUE #2277: Iceberg error: snapshot ID too big
## Description

https://twitter.com/mim_djo/status/1735480225620955304

> trying to read an iceberg table using [@GlareDB](https://twitter.com/GlareDB)
 , sorry your snapshot id is just too big 😅😋🤓

image

vrongmeal:
Read the first line of the `version-hint.text` and take it as it is to generate the metadata file name. Omit parsing it as a number since there are cases when it doesn't fit in an `i32` type. Instead of parsing it as an `i64` or `i128` we should simply skip it since such a version number seems like a bug and can happen again but shouldn't stop us from reading the table.

Fixes #2277


universalmind303 (approved):

testdata/sqllogictests_iceberg/local.slt
```diff
@@ -94,3 +94,9 @@ REG AIR    314
 SHIP       316
 TRUCK      264
 
+# We should be able to query a table with version number longer than i32 limit.
+query T
+select count(*) = 1000
+  from iceberg_scan('../../testdata/iceberg/tables/lineitem_simple_longversion');
+----
+t
```
scsmithr, line 102: Something I've found useful when adding tests in response to issues is to link to the issue directly. Prevents us from accidentally deleting it in the future thinking some other test covers it or something.

scsmithr (approved):

testdata/sqllogictests_iceberg/local.slt
vrongmeal: Makes sense! Will link it :)

vrongmeal: Updated!

Issue #2025, opened 2023-10-31
Postgres: Enums


greyscaled:

scsmithr:
Error message:

```
Error during planning: Unable to fetch table provider for 'dev.public.suppress_vulnerability': Unknown Postgres OIDs: [23, 1043, 23, 23, 1043, 1043, 3802, 1043, 1043, 1114, 1114, 108832, 3802, 1043, 1043, 1043, 1043, 20, 20, 20, 20, 1114, 1114]
```

Immediate next step: Need to specify _which_ oid is unknown. Currently shows us all the oids we have, but not which one we failed to get the type for.

PR #2283, opened 2023-12-20
fix: filter unknown oids for Postgres types

hemanth94:
Fixes issue : #2025 
Filters out unknown type - oids and returns them in error log.
@scsmithr  Please correct me if needed.

Thanks.


tychoish (approved):
@scsmithr could you gut check this real quick too?

scsmithr (approved):
Seems reasonable. Doesn't completely close the issue since we want to be able to handle the OIDs somehow (specifically enums), but this does provide a better error message.

PR #2284, opened 2023-12-20
fix: Hacky fix for running the experimental scheduler when spawning task

vrongmeal:
Until I figure out the best solution (which hopefully is close).

Pushing just to benchmark and test stuff successfully.

To run the tokio console debugger:

1. Download the CLI:
   ```
   cargo --locked tokio-console
   ```
2. Build/Run local with new just commands (along with `--debug-tokio` flag):
   ```
   just run-debug-tokio local --debug-tokio
   ```

**NOTE:** This builds in a new `target` directory: `unstable-tokio-target` so as to not mess up with existing target directory (which can cause issues with LSP etc.).

### DIFF
--- a/crates/glaredb/src/commands.rs
+++ b/crates/glaredb/src/commands.rs
@@ -52,12 +52,20 @@ trait RunCommand {
     fn run(self) -> Result<()>;
 }
 
-impl RunCommand for LocalArgs {
-    fn run(self) -> Result<()> {
+impl LocalArgs {
+    #[cfg(not(release))]
+    fn start_tokio_debugger(&self) {
         if self.debug_tokio {
             eprintln!("> Starting tokio debugger");
             console_subscriber::init();
         }
+    }
+}
+
+impl RunCommand for LocalArgs {
+    fn run(self) -> Result<()> {
+        #[cfg(not(release))]
+        self.start_tokio_debugger();
 
         let runtime = build_runtime("local")?;
         runtime.block_on(async move {
--- a/crates/glaredb/src/commands.rs
+++ b/crates/glaredb/src/commands.rs
@@ -54,6 +54,11 @@ trait RunCommand {
 
 impl RunCommand for LocalArgs {
     fn run(self) -> Result<()> {
+        if self.debug_tokio {
+            eprintln!("> Starting tokio debugger");
+            console_subscriber::init();
+        }
+
         let runtime = build_runtime("local")?;
         runtime.block_on(async move {
             let query = match (self.file, self.query) {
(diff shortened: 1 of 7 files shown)


crates/glaredb/src/commands.rs
```diff
@@ -55,6 +55,11 @@ trait RunCommand {
 
 impl RunCommand for LocalArgs {
     fn run(self) -> Result<()> {
+        if self.debug_tokio {
+            eprintln!("> Starting tokio debugger");
+            console_subscriber::init();
+        }
```
tychoish, line 61: is there an easy way to ifdef this out in release mode?

crates/sqlexec/src/distexec/scheduler.rs
```diff
@@ -25,8 +25,28 @@ impl Scheduler {
...
+        // Due to our implementation, the runtime is blocked somehow and unable
+        // to spawn new tasks.
+        //
+        // HACK: Introducing a new task here doesn't block the runtime from
+        // spawning new tasks.
+        let (dummy_task_sender, dummy_task_receiver) = async_channel::unbounded();
+        tokio::spawn(async move {
+            while let Ok(task) = dummy_task_receiver.recv().await {
+                actual_task_sender.send(task).await.unwrap();
+            }
```
tychoish, line 39: how fast would we expect this to thrash/spawn?

justfile
```diff
@@ -29,6 +29,14 @@ run *args: protoc
...
   cargo build --bin glaredb {{args}}
 
+# Build glaredb with unstable_tokio flag.
+#
+# Need to run the local command with `--debug-tokio` argument.
+build-debug-tokio *args:
+  RUSTFLAGS="--cfg tokio_unstable" CARGO_TARGET_DIR=tokio-unstable-target just build {{args}}
+
+run-debug-tokio *args:
+  RUSTFLAGS="--cfg tokio_unstable" CARGO_TARGET_DIR=tokio-unstable-target just run {{args}}
```
tychoish, line 39: I might be _slightly_ partial to using `--target-dir` over `CARGO_TARGET_DIR` but it's probably not material

tychoish (commented):
is it possible to require/switch to the new scheduler via the command line.

also we should add a task to add a new matrix dimension to the SLT tests that would run with the new scheduler. (x pg x flight once this land and after this lands.)

crates/glaredb/src/commands.rs
vrongmeal: I'll do that!

crates/sqlexec/src/distexec/scheduler.rs
vrongmeal: Immediately!

justfile
vrongmeal: Yeah, just wanted something quick.

vrongmeal:
> is it possible to require/switch to the new scheduler via the command line.

No, unfortunately have to set it for the session.

crates/glaredb/src/commands.rs
vrongmeal: Updated!

tychoish (approved):

PR #2285, opened 2023-12-20
feat: add expirations to flight sessions

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2236


crates/rpcsrv/src/util.rs
```diff
@@ -1,7 +1,95 @@
...
+        // We don't want to run this loop forever
+        // So we check if the map is still alive, and if it isn't, we break out of the loop
+        // This will prevent the task from running forever
+        tokio::spawn(async move {
+            let ttl = expiring_map.ttl;
+            loop {
+                tokio::time::sleep(Duration::from_secs(60)).await;
+                if let Some(map) = weak_map.upgrade() {
+                    let now = Instant::now();
+                    map.retain(|_, v| now - v.1 < ttl);
```
tychoish, line 54: does anything ever resize/repack the underlying map?

crates/rpcsrv/src/util.rs
```diff
@@ -1,7 +1,95 @@
...
+    /// The entries are checked every minute
+    pub fn with_ttl(ttl: Duration) -> ExpiringDashMap<K, V> {
+        let map: Arc<DashMap<K, (V, Instant)>> = Arc::new(DashMap::new());
+        let expiring_map = ExpiringDashMap { map, ttl };
+        let weak_map = Arc::downgrade(&expiring_map.map);
+        // Spawn a background task that will check for expired entries every minute
+        // We don't want to run this loop forever
+        // So we check if the map is still alive, and if it isn't, we break out of the loop
+        // This will prevent the task from running forever
+        tokio::spawn(async move {
```
tychoish, line 48: I've always implemented TTL maps more lazily with a second heap-sorted for the structure, and then reads/writes check he heap and drop some entries if needed.

crates/rpcsrv/src/util.rs
```diff
@@ -1,7 +1,95 @@
...
+
 /// Key used for the connections map.
 // TODO: Possibly per user connections?
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub(crate) struct ConnKey {
     pub(crate) ip: String,
     pub(crate) port: String,
 }
+
+pub struct ExpiringDashMap<K, V> {
```
tychoish, line 17: https://github.com/cobbinma/r-cache <- why not use something like this which has similar functionality and tests.

crates/rpcsrv/src/flight/handler.rs
```diff
@@ -101,26 +106,14 @@ impl FlightSessionHandler {
...
+    async fn get_or_create_ctx<T>(&self, request: &Request<T>) -> Result<SessionRef, Status> {
         let remote = request.remote_addr().unwrap();
 
         let ip = remote.ip().to_string();
         let port = remote.port().to_string();
         let conn_key = ConnKey { ip, port };
-
         if self.sessions.contains_key(&conn_key) {
-            let sess = self.sessions.get(&conn_key).unwrap().clone();
+            let sess = self.sessions.get(&conn_key).unwrap();
```
tychoish, line 116: is there a way to have the connections/sessions keep track of active callers/users/etc and have them drop themselves if they've been idle for a certain period of time. (e.g. have something in the destructor)

crates/rpcsrv/src/util.rs
universalmind303: Orignally i wanted to use a stack, but there arent as many good concurrent stack structures without pulling in some more dependencies. So i just went with this simple lightweight approach.

crates/rpcsrv/src/flight/handler.rs
universalmind303: not that I'm aware of. 

If you are aware of a solution that automatically manages the lifecycles based on the connection timeout, I think that would be much better and would happily do that over this cache based approach.

crates/rpcsrv/src/util.rs
universalmind303: This seemed trivial enough to just implement ourselves without either pulling in larger caching libraries like `moka`, or relying smaller unmaintained libraries like `r-cache`. 
I also wanted it to be self contained without having to spawn an external background job to clear the entries like is the case with r-cache.

crates/rpcsrv/src/util.rs
universalmind303: no, It should be easy enough to add in some resizing logic. Will update.

crates/rpcsrv/src/util.rs
tychoish: you do spawn the background job on line 48, so that seems roughly equivalent.

r-cache doesn't seem unmaintained: it's small and feature complete, there are no open issues, and they made changes in the last year. some things don't need much. 

r-cache also has tests which provide some extra confidence.

tychoish:
I was thinking about this some more, and from a higher level: how much do we expect sessions to ever be reused? 

I think given the compute model, we should assume that users can make connections to multiple glaredb backend workers, (the only cost of using more than one glaredb worker for a user is that we have to load the catalog more than once, and the operational flexibility of assuming that sessions aren't going to be pinned, is probably ultimately a strength of the platform. 

In this world does it make sense to cache the session at all? Is there state in the connection that would lead users to have an unexpected experience if their requests get routed to different connections (or get different results from different sessions) Particularly given that the proxy is in front of the connection: multiple connections to the proxy, could get mapped onto one connection between the proxy and the worker?

crates/rpcsrv/src/util.rs
universalmind303: if you feel strongly about using a library we can. This seemed trivial enough to not need one, so I didn't.

universalmind303:
> In this world does it make sense to cache the session at all? Is there state in the connection that would lead users to have an unexpected experience if their requests get routed to different connections (or get different results from different sessions) Particularly given that the proxy is in front of the connection: multiple connections to the proxy, could get mapped onto one connection between the proxy and the worker?

yes there is state, specifically with prepared statements.

crates/rpcsrv/src/flight/handler.rs
scsmithr, line 116: This can panic. `contains_key` doesn't check if the entry is expired, but `get` will return `None` if it is.

tychoish: I think 
- there's this approach (or some variant therein), with a single worker that goes through and cleans up stale connections;
- you could write a version where the worker thread is triggered by a condition variable and some other event (the drop method on the connection object could trigger a cleanup pass);
- you store last-used in the map along with an Arc wrapped connection, and have a monitor task for each connection (maybe isolated to a simple runtime if it matters) that either sleeps till the ttl or drops the reference or if there are no other holders of the session.

crates/rpcsrv/src/util.rs
```diff
@@ -1,7 +1,95 @@
...
+    }
+
+    pub fn insert(&self, key: K, value: V) {
+        self.map.insert(key, (value, Instant::now() + self.ttl));
+    }
+
+    pub fn get(&self, key: &K) -> Option<V> {
+        self.map.get_mut(key).and_then(|mut entry| {
+            if entry.value().1 > Instant::now() {
+                entry.value_mut().1 = Instant::now() + self.ttl;
```
tychoish, line 71: as I was thinking about this and writing another comment it strikes me that because we're only setting the TTL when sessions are _accessed_, a session could be held/in use for a period longer than the TTL at which point it would (eventually) get dropped from the map, which means we end up caching _flaky_ connections very well and really stable connections would get dropped/leaked. 

Maybe (if we keep this model) we need some kind of wrapper that sets the "last used" time of the object in the map as part of the `Drop` and if the session/context is cached and wrapped in an `Arc` you'd be able to see (and skip) items that had active users.

PR #2286, opened 2023-12-21
feat: jq function

universalmind303:


crates/sqlbuiltins/src/functions/scalars/jq.rs
```diff
@@ -0,0 +1,137 @@
...
+            let (f, errs) = jaq_parse::parse(&val, jaq_parse::main());
+            if !errs.is_empty() {
+                return Err(datafusion::common::DataFusionError::Execution(
+                    errs.into_iter()
+                        .map(|e| e.to_string())
+                        .collect::<Vec<_>>()
+                        .join("\n"),
+                ));
+            }
+            defs.compile(f.unwrap())
```
tychoish, line 104: are we recompiling the expression for every row/value examined?

crates/sqlbuiltins/src/functions/scalars/jq.rs
```diff
@@ -0,0 +1,137 @@
...
+            return Err(datafusion::common::DataFusionError::Execution(
+                "missing JQ query".to_string(),
+            ))
+        }
+    };
+
+    let doc: Value = match get_nth_scalar_value(args, 0) {
+        Some(ScalarValue::Utf8(Some(val))) | Some(ScalarValue::LargeUtf8(Some(val))) => {
+            serde_json::from_str(&val)
+                .map_err(|err| datafusion::common::DataFusionError::Execution(err.to_string()))?
```
tychoish, line 122: it seems inideal (but maybe inevitable for jaq) that we have to convert the string to json and then parse it?

crates/sqlbuiltins/src/functions/scalars/jq.rs
```diff
@@ -0,0 +1,137 @@
...
+
+use super::get_nth_scalar_value;
+
+#[derive(Clone)]
+pub struct Jq;
+
+impl ConstBuiltinFunction for Jq {
+    const NAME: &'static str = "jq";
+    const DESCRIPTION: &'static str = "Run a jq query on a JSON document. 
+Note: this uses the rust jaq library, which does not yet support the full jq language.";
```
tychoish, line 23: thoughts: should we call this `jaq` so that we don't run into problems if we eventually want to have "real" `jq"?

tychoish (commented):
My gut feeling is that we should use [jq-rs](https://github.com/onelson/jq-rs) because giving people "real jq" is better than giving people "jq knock off" in terms of people's expectations?

I think doing that (or at least making it easy to do that in the future?) gives us some good options.

universalmind303:
> My gut feeling is that we should use [jq-rs](https://github.com/onelson/jq-rs) because giving people "real jq" is better than giving people "jq knock off" in terms of people's expectations?
> 
> I think doing that (or at least making it easy to do that in the future?) gives us some good options.

jaq aims to be feature complete as far as supporting the jq language. I **strongly** feel that we should not use the jq library for a number of reasons. 

while yes, it is more feature complete than the rust implementation, it is also much slower, and either requires dynamic linking or installing the entire c binary. The jaq implementation is much lighter weight. the jq-rs library also has not published a new release in over 4 years. The jaq implemenation is very active.

crates/sqlbuiltins/src/functions/scalars/jq.rs
universalmind303: yeah it's not ideal, but i dont think there's another way atm. We can optimize later if needed.

crates/sqlbuiltins/src/functions/scalars/jq.rs
universalmind303: yeah, datafusion doesn't really give us any way to do this once for all operations.

universalmind303:
so this actually has to be done outside of the datafusion scalar udf pattern due to the dynamic return type. Closing this for now. Will look more into it later.

tychoish:
> jaq aims to be feature complete as far as supporting the jq language.

aspirational support is not exactly the same as having the same semantics, particularly because there isn't a formal specification of the JQ language. 

I think if we're not going to give people _actual `jq`_ then we shouldn't call it `jq`. It's fine, I think, to call the function `jaq` (and have `jq` return an error telling people to use `jaq`?)

> while yes, it is more feature complete than the rust implementation, it is also much slower, and either requires dynamic linking or installing the entire c binary. 

I suspect that the speed differential would be ameliorated by mnemonizing the compilation of query statements; it's also the case that rs-jq operates on the byte-stream and so avoids needing to process the input stream twice (once to convert into serde-json structures and another time to execute the query.) 

I think we'd want to statically link the library, which, is 300kb: i'm not worried about the implication of the extra build dependency or impact on the binary size. In no case is it shelling out to jq.

> The jaq implementation is much lighter weight. the jq-rs library also has not published a new release in over 4 years. The jaq implementation is very active.

rs-jq has had commits recently, and the maintainers comments make it seem like he's interested and willing to take the library in the right direction. active development, in cases like this worry me a bit because it means people's query results could/would change somewhat as the library evolves in ways that they may not be expecting (particularly if we sell it as JQ proper.

universalmind303:
```
error: failed to run custom build command for `jq-sys v0.2.2`

Caused by:
  process didn't exit successfully: `/Users/corygrinstead/Development/glaredb/target/debug/build/jq-sys-7c8dda530aa77d23/build-script-build` (exit status: 101)
  --- stdout
  running: cd "/Users/corygrinstead/Development/glaredb/target/debug/build/jq-sys-0460cdc5d4d09767/out/sources/jq" && "sh" "-c" "exec \"$0\" \"$@\"" "autoreconf" "-fi"
  running: cd "/Users/corygrinstead/Development/glaredb/target/debug/build/jq-sys-0460cdc5d4d09767/out/sources/jq" && "sh" "-c" "exec \"$0\" \"$@\"" "autoreconf" "-fi"
  running: cd "/Users/corygrinstead/Development/glaredb/target/debug/build/jq-sys-0460cdc5d4d09767/out/sources/jq" && "sh" "-c" "exec \"$0\" \"$@\"" "autoreconf" "-fi"

  --- stderr
  autoreconf: line 0: exec: autoreconf: not found
  thread '<unnamed>' panicked at /Users/corygrinstead/.cargo/registry/src/index.crates.io-6f17d22bba15001f/autotools-0.2.6/src/lib.rs:781:5:

  command did not execute successfully, got: exit status: 127

  build script failed, must exit now
  Build experienced some sort of failure. Retrying (1).
  autoreconf: line 0: exec: autoreconf: not found
  thread '<unnamed>' panicked at /Users/corygrinstead/.cargo/registry/src/index.crates.io-6f17d22bba15001f/autotools-0.2.6/src/lib.rs:781:5:

  command did not execute successfully, got: exit status: 127

  build script failed, must exit now
  Build experienced some sort of failure. Retrying (2).
  autoreconf: line 0: exec: autoreconf: not found
  thread '<unnamed>' panicked at /Users/corygrinstead/.cargo/registry/src/index.crates.io-6f17d22bba15001f/autotools-0.2.6/src/lib.rs:781:5:

  command did not execute successfully, got: exit status: 127

  build script failed, must exit now
  thread 'main' panicked at /Users/corygrinstead/.cargo/registry/src/index.crates.io-6f17d22bba15001f/jq-sys-0.2.2/build.rs:8:21:
  autotools build: ()
  ```
This is exactly why id rather not use c libraries & would prefer to use native rust implementations. They usually don't work out of the box and make our build process more difficult. Trying to just run through the example provided and i get a bunch of build errors. the native implementation works out of the box. 
  
I closed this PR as it requires more work that appeared. But it seems like we can't come to an agreement on implementation either.

PR #2288, opened 2023-12-22
chore: Remote `fn runtime_preference()` from `TableFunc` trait.

### LINKED ISSUE #2165: Chore: Clarify where table functions run
## Description

I'm currently running through adding some table funcs, and I'm unsure what gets run where.

The full interface:

```rust
#[async_trait]
pub trait TableFunc: Sync + Send {
    /// The name for this table function. This name will be used when looking up
    /// function implementations.
    fn name(&self) -> &str;
    fn runtime_preference(&self) -> RuntimePreference;
    fn detect_runtime(
        &self,
        _args: &[FuncParamValue],
        _parent: RuntimePreference,
    ) -> Result<RuntimePreference> {
        Ok(self.runtime_preference())
    }

    /// Return a table provider using the provided args.
    async fn create_provider(
        &self,
        ctx: &dyn TableFuncContextProvider,
        args: Vec<FuncParamValue>,
        opts: HashMap<String, FuncParamValue>,
    ) -> Result<Arc<dyn TableProvider>>;
    /// Return the signature for this function.
    /// Defaults to None.
    // TODO: Remove the default impl once we have `signature` implemented for all functions
    fn signature(&self) -> Option<Signature> {
        None
    }
}
```

From what I'm reading, `create_provider` _always_ gets run remotely if runtime preference is remote. If this is the case, then that should be documented.

vrongmeal:
We should only have 1 source of truth for runtime preference for the functions. Here, `runtime_preference` method is removed so the runtime is always calculated from `detect_runtime`.

Fixes #2165

### DIFF
--- a/crates/metastore/src/database.rs
+++ b/crates/metastore/src/database.rs
@@ -1302,10 +1302,10 @@ impl BuiltinCatalog {
             let schema_id = schema_names
                 .get(DEFAULT_SCHEMA)
                 .ok_or_else(|| MetastoreError::MissingNamedSchema(DEFAULT_SCHEMA.to_string()))?;
-            let mut entry = func.as_function_entry(oid, *schema_id);
-            entry.runtime_preference = func.runtime_preference();
 
+            let entry = func.as_function_entry(oid, *schema_id);
             insert_entry(oid, CatalogEntry::Function(entry))?;
+
             schema_objects
                 .get_mut(schema_id)
                 .unwrap()
--- a/crates/protogen/proto/metastore/catalog.proto
+++ b/crates/protogen/proto/metastore/catalog.proto
@@ -199,9 +199,10 @@ message FunctionEntry {
     LOCAL = 1;
     REMOTE = 2;
   }
+
   EntryMeta meta = 1;
   FunctionType func_type = 2;
-  RuntimePreference runtime_preference = 3;
+  reserved 3;
   Signature signature = 4;
 }
 
--- a/crates/protogen/proto/metastore/catalog.proto
+++ b/crates/protogen/proto/metastore/catalog.proto
@@ -202,7 +202,7 @@ message FunctionEntry {
 
   EntryMeta meta = 1;
   FunctionType func_type = 2;
-  reserved 3;
+  reserved 3; // Function runtime preference (static)
   Signature signature = 4;
 }
 
--- a/crates/sqlbuiltins/src/functions/table/iceberg/data_files.rs
+++ b/crates/sqlbuiltins/src/functions/table/iceberg/data_files.rs
@@ -35,8 +35,12 @@ impl ConstBuiltinFunction for IcebergDataFiles {
 
 #[async_trait]
 impl TableFunc for IcebergDataFiles {
-    fn runtime_preference(&self) -> RuntimePreference {
-        RuntimePreference::Remote
+    fn detect_runtime(
+        &self,
+        _args: &[FuncParamValue],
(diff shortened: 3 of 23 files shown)


crates/protogen/proto/metastore/catalog.proto
```diff
@@ -199,9 +199,10 @@ message FunctionEntry {
     LOCAL = 1;
     REMOTE = 2;
   }
+
   EntryMeta meta = 1;
   FunctionType func_type = 2;
-  RuntimePreference runtime_preference = 3;
+  reserved 3;
```
universalmind303, line 205: since this isn't public do we need to reserve this? is there any harm in just removing?

crates/metastore/src/database.rs
```diff
@@ -1302,10 +1302,10 @@ impl BuiltinCatalog {
             let schema_id = schema_names
                 .get(DEFAULT_SCHEMA)
                 .ok_or_else(|| MetastoreError::MissingNamedSchema(DEFAULT_SCHEMA.to_string()))?;
-            let mut entry = func.as_function_entry(oid, *schema_id);
-            entry.runtime_preference = func.runtime_preference();
```
universalmind303, line 1305: previously the only reason we needed different for loops for `table_funcs()` and `scalar_funcs()` was for the handling of `runtime_preference`. Since that is removed, we should follow up with cleaning up that logic.

crates/protogen/proto/metastore/catalog.proto
scsmithr: This is tricky since these protos also have definitions that get written to disk, and so we have to be careful to make sure we're not clobbering something accidentally.

In this case, `FunctionEntry` is only sent over the network and never persisted, so removing the reserve would be fine. But it's on the programmer to know that this specific entry isn't serialized, but something like `TunnelEntry` is. Since we don't have a clear distinction, I think always being safe and reserving ids is the preferred action.

(Also when we reserve, we should make a comment on what it was previously just to provide context when we look back at it)

scsmithr (approved):

crates/sqlbuiltins/src/functions/table/iceberg/data_files.rs
```diff
@@ -35,8 +35,12 @@ impl ConstBuiltinFunction for IcebergDataFiles {
...
 #[async_trait]
 impl TableFunc for IcebergDataFiles {
-    fn runtime_preference(&self) -> RuntimePreference {
-        RuntimePreference::Remote
+    fn detect_runtime(
+        &self,
+        _args: &[FuncParamValue],
+        _parent: RuntimePreference,
+    ) -> Result<RuntimePreference> {
+        Ok(RuntimePreference::Remote)
```
universalmind303, line 43: tangent: but shouldn't this technically check the url path to detect the runtime? Same with a few of these other object store related ones. 

The other ones have todo's so it's clear that it's not yet implemented. Can we add that same here? We should also open up an issue to properly detect the runtimes for these functions

tychoish (approved):

crates/sqlbuiltins/src/functions/table/iceberg/data_files.rs
vrongmeal: Yeah, I'll add a TODO here as well.

crates/metastore/src/database.rs
vrongmeal: Will open another PR for this (since it would require some more review).

Issue #2220, opened 2023-12-07
add `shard_results` and `hashtext` functions


tychoish:
## Description

I was working on something with GlareDB cloud and I wished to have these functions in `glaredb`. We can discus names, but basically the operations would be:

### `hashtext` 

This could easily be `fnv()` or `fnv1a()` wold take the raw data of a field, and return the hash value as an unsigned 64 bit integer. If we differ from the postgres function we shouldn't use the same name. I'm partial to 64bit fnv1a but any non-cryptographic hash is fine. 

### `shard_results`

`shard_results(<data>, <num_shards>, <shard_id>)` 

Data could be any type (we'll use it's byte sequence, no need to cast), `num_shards` is a positive non-zero integer, and `shard_id` is a number that is within the `[0,<num_shards>)` range. The function would return a boolean, and be used in a `WHERE` clause. 

The operation would be, basically `hash(<data>) % <num_shards> == <shard_id>`. 

This function should be implemented in terms of the first.

As future work, It would be interesting if for parquet data sources, to see if it would end up working so that we'd pull the column in question, do the filtering, and then pull the remaining data out?

## Use Case 

If you have multiple stateless application servers and you want to divide the output of query (which represents some work), into slices (shards) for each application servers, this function can help push that calculation into the database, and reduce the amount of data that's sent to the application.

PR #2298, opened 2023-12-23
feat: add non-cryptographic hash function

tychoish:
Part of #2220

Notes:

- While Rust's hashmaps use a siphash variant, they use a less less
  common and less robust variant (for speed) reasons [which makes
  sense for default hashmaps but less sense in other cases]. They also
  deprecated the access to the specific algorithims, in a way that
  makes me less confident that the undeprecated form will be
  semantically stable.

- I contemplated (and tried) to share the implementation, but: I think
  these ought to be different functions and becuause the hashers are
  stateful, I think we'd end up writing a lot more code with limited
  benefit, but perhaps I've missed something.


crates/sqlbuiltins/src/functions/scalars/hashing.rs
```diff
@@ -0,0 +1,100 @@
...
+use siphasher::sip::SipHasher24;
+
+use super::*;
+
+pub struct SipHash;
+
+impl ConstBuiltinFunction for SipHash {
+    const NAME: &'static str = "siphash";
+    const DESCRIPTION: &'static str =
+        "Calculates a 64bit non-cryptographic hash (SipHash24) of the value.";
```
universalmind303, line 13: is this row or columnar hashing? 

from the implementation, it looks like it's vectorized (columnar), but i'm not sure if that is clear from the slts or the description here.

testdata/sqllogictests/functions/hashing.slt
universalmind303, line 1: can you add some slts of testing this on a column with multiple values.

universalmind303 (changes requested):

crates/sqlbuiltins/src/functions/scalars/hashing.rs
```diff
@@ -0,0 +1,100 @@
...
+
+use fnv::FnvHasher;
+use siphasher::sip::SipHasher24;
+
+use super::*;
+
+pub struct SipHash;
+
+impl ConstBuiltinFunction for SipHash {
+    const NAME: &'static str = "siphash";
```
universalmind303, line 11: Some of our other functions such as `digest` have them all wrapped up into a single function with an algorithm argument. Wondering if we could/should do the same here. `hash(<value>, 'siphash' | 'fnv')`. I'm not partial to either way, just a thought.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
tychoish: I think it's row based (and it wouldn't make much sense if it weren't for the partitioning case,) and is consistent with the other hashing/checksum functions. My intention was that that the word "value" (rather than column or values) would indicate this, but if there's a better term of art, I'm not wed to this wording.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
tychoish: In the digest case the `digest` functionality is also provided by dedicated methods (like the ones in this PR.) 

My assumption is that the `digest` API mostly exists so you could (say) store or compute which hash algorithm to use, and therefore select the algorithm at runtime. This is a pretty narrow use case, and maybe makes sense for cryptographic hashes (maybe?) but makes much less sense for non-cryptographic hashes, but it would be easy to add later if someone wanted it.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
```diff
@@ -0,0 +1,96 @@
...
+            Volatility::Immutable,
+        ))
+    }
+}
+impl BuiltinScalarUDF for SipHash {
+    fn as_expr(&self, args: Vec<Expr>) -> Expr {
+        let udf = ScalarUDF {
+            name: Self::NAME.to_string(),
+            signature: ConstBuiltinFunction::signature(self).unwrap(),
+            return_type: Arc::new(|_| Ok(Arc::new(DataType::Utf8))),
```
universalmind303, line 30:
```suggestion
            return_type: Arc::new(|_| Ok(Arc::new(DataType::Int64))),
```

crates/sqlbuiltins/src/functions/scalars/hashing.rs
```diff
@@ -0,0 +1,96 @@
...
+            Volatility::Immutable,
+        ))
+    }
+}
+impl BuiltinScalarUDF for FnvHash {
+    fn as_expr(&self, args: Vec<Expr>) -> Expr {
+        let udf = ScalarUDF {
+            name: Self::NAME.to_string(),
+            signature: ConstBuiltinFunction::signature(self).unwrap(),
+            return_type: Arc::new(|_| Ok(Arc::new(DataType::Utf8))),
```
universalmind303, line 75:
```suggestion
            return_type: Arc::new(|_| Ok(Arc::new(DataType::Int64))),
```

testdata/sqllogictests/functions/hashing.slt
```diff
@@ -0,0 +1,108 @@
...
+
+statement ok
+select siphash(9001);
+
+statement ok
+select siphash(true);
+
+
+statement error
+select fnv(1, 2, 3);
```
universalmind303, line 21: I think this _technically_ violates the type signature of `TypeSignature::Any(1)`. Maybe we should use `TypeSignature::VariadicAny` instead

testdata/sqllogictests/functions/hashing.slt
```diff
@@ -0,0 +1,108 @@
...
+----
+8771948186893062792
+
+query I
+select siphash([0, 100, 100, 300, 500, 800]);
+----
+3773389192103504674
+
+query I
+select fnv();
```
universalmind303, line 64: I think this violate the type signature of `TypeSignature::Any(1)`. We either shouldn't allow this, or have the signature reflect this.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
```diff
@@ -0,0 +1,96 @@
...
+        ))
+    }
+}
+impl BuiltinScalarUDF for SipHash {
+    fn as_expr(&self, args: Vec<Expr>) -> Expr {
+        let udf = ScalarUDF {
+            name: Self::NAME.to_string(),
+            signature: ConstBuiltinFunction::signature(self).unwrap(),
+            return_type: Arc::new(|_| Ok(Arc::new(DataType::Utf8))),
+            fun: Arc::new(move |input| match get_nth_scalar_value(input, 0) {
```
universalmind303, line 31: I don't think this does what we want

```rust
fn get_nth_scalar_value(input: &[ColumnarValue], n: usize) -> Option<ScalarValue> {
    match input.get(n) {
        Some(input) => match input {
            ColumnarValue::Scalar(scalar) => Some(scalar.clone()),
            ColumnarValue::Array(arr) => ScalarValue::try_from_array(arr, 0).ok(),
        },
        None => None,
    }
}
```

If it's an array, it'll just use the first element. So this will likely produce invalid results when used on a column. 

sorry for being unclear in this comment. 
> can you add some slts of testing this on a column with multiple values.

What I meant to say was run the hashing function on a column with multiple **rows** of different values. I expect that they will all have the hash of the first value.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
universalmind303: FWIW, I think we can add a slt for this by passing the output into a function that only accepts `UInt64`. 

something like 
```sql 
statement error 
select initcap(fnv('hello world'))

statement ok
select trunc(fnv('hello world'), 2)
```

tychoish: The fact that this compiles at all is a bit concerning.

testdata/sqllogictests/functions/hashing.slt
tychoish: shouldn't the function dispatcher prevent this? Will convert it into an error

crates/sqlbuiltins/src/functions/scalars/hashing.rs
tychoish: > If it's an array, it'll just use the first element. So this will likely produce invalid results when used on a column.

The [docs here](https://arrow.apache.org/datafusion/library-user-guide/adding-udfs.html#adding-user-defined-functions-scalar-window-aggregate-table-functions) make it seem pretty clearly that scalar functions are only ever get one row [which is backed up by the example](https://github.com/apache/arrow-datafusion/blob/main/datafusion-examples/examples/simple_udf.rs) which wouldn't make sense if column references for scalar functions worked the way you're suggesting. (and the get_nth_value function [has been around for a while](https://github.com/GlareDB/glaredb/commit/d43669cb7ab49fe5af1b6eb3da80a938830ddc00#diff-1b7ec8a265c3a8ec87f20cafa22e741e7f658cbbc3c6883748d70bd14f65913dR261), so we've been doing this for a while, with no ill effects.

universalmind303: in the datafusion example, you can see that they are handling the array directly. https://github.com/apache/arrow-datafusion/blob/main/datafusion-examples/examples/simple_udf.rs#L80


The `get_nth_scalar_value` function was designed to unwrap **scalars** out of ColumnarValue, not unwrap all values. It doesn't properly extract the array from `ColumnarValue::Array` variant.


If we try this on an array, we can see that it is using the first value from the array, not operating on the elements in the array. 
```sql
> create table t (x int);
Table created
> insert into t values (1), (2), (3);
Inserted 3 rows
> select siphash(x) as actual, siphash(arrow_cast(1, 'Int32')) as one from t;
┌─────────────────────┬─────────────────────┐
│              actual │                 one │
│                  ── │                  ── │
│              UInt64 │              UInt64 │
╞═════════════════════╪═════════════════════╡
│ 1308865109194702177 │ 1308865109194702177 │
│ 1308865109194702177 │ 1308865109194702177 │
│ 1308865109194702177 │ 1308865109194702177 │
└─────────────────────┴─────────────────────┘
```

crates/sqlbuiltins/src/functions/scalars/hashing.rs
universalmind303: yeah the return types are only checked during planning, so they're kinda hard to catch if you miss them.

testdata/sqllogictests/functions/hashing.slt
universalmind303: Unless we are bypassing something, I'd expect datafusion's planner to handle this. Maybe we should create a MRE and open up an issue with them.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
tychoish: Interesting... we should probably have the `get_nth_scalar` function (and the friends I've written recently) either error for array values, or take a closure. 

Relatedly, this means that the `kdl` functions and maybe the postgres functions that use it (though perhaps more ambiguous) will have the same issue.

testdata/sqllogictests/functions/hashing.slt
tychoish: I haven't intentionally bypassed anything: you did more work in the function handling stuff recently than me.

scsmithr: I agree that providing nothing and having it not error is a bit weird. I don't know if this is datafusion allowing that through, or us not propagating the signature upwards correctly. But I think this is still good to merge as is since I think fixing what's allowed based on the signature to be a separate scope.

scsmithr (approved):

testdata/sqllogictests/functions/hashing.slt
tychoish: yeah, agreed.

crates/sqlbuiltins/src/functions/scalars/hashing.rs
tychoish: I've added a test case for this.

tests/tests/test_functions.py
```diff
@@ -0,0 +1,19 @@
...
+        with glaredb_connection.cursor() as cur:
+            cur.execute(operation)
+
+    with glaredb_connection.cursor() as cur:
+        cur.execute("select siphash(x) as actual, siphash(arrow_cast(1, 'Int32')) as one from t")
+        result = cur.fetchall()
+
+    assert result[0][0] == result[0][1]
+    assert result[1][0] != result[1][1]
+    assert result[2][0] != result[2][1]
```
universalmind303, line 19: not blocking, but I'd prefer if these were in the slts. They run on all protocols, while it appears that these are specific to using the postgres protocol.

universalmind303 (approved):

tests/tests/test_functions.py
tychoish: Fair enough, though I'm not particularly worried about execution level operations like this being impacted by service architecture aspects, and I _am_ more worried about the ways that multi-statement operations' statefullnes in SLTs can leak.

PR #2299, opened 2023-12-23
feat: partition results helper

### LINKED ISSUE #2220: add `shard_results` and `hashtext` functions

tychoish:
Adds partitioning (sharding) of result sets using the hashing method. 

Implemented on top of #2298.

Closes #2220


scsmithr (approved):

testdata/sqllogictests/functions/hashing.slt
```diff
@@ -106,3 +106,62 @@ query I
...
+statement ok
+select partition_results(100, 10.0, 1.0);
+
+statement error
+select partition_results(100, 10.5, 1.5);
+
+statement error
+select partition_results(16, 4, 4);
+
+query B
```
scsmithr, line 149: Technically there is no 'B'. Only 'I', 'R', and 'T'. For bools, I just use 'T'. I also don't think the types are actually checked, just that there's the correct number of them.

https://github.com/risinglightdb/sqllogictest-rs/blob/main/sqllogictest/src/column_type.rs#L11-L43

vrongmeal (approved):

Issue #2218, opened 2023-12-07
data source from JSON array data


tychoish:
Presently, we have `ndjson` support for reading files and endpoints with new-line delimited json. This is great for dealing with files that store sequences or streams of document-structured data; however, many JSON API's return sequences of documents as members of a JSON array, and we should add support for accessing these kinds of data sources, which will help connect GlareDB with potential users who have json data from other sources.

## Components

I've just split aspects out into components. 

- [ ] `read_arjson` ("array json"): table func: these should read the data source (local file or URL, or object store, consistent with other functions) into a memory table with one row for every element in the array. There should be an option for maximum data size, overwhich the table will result in an error.
- [ ] `COPY TO` support, [open question: handling of null values in tables (omit vs preserve), I lean towards omit]
- [ ] ability to add data source of `arjson` data in the catalog with the appropriate options.

## Semantics

- schema inference: read every document in the array, construct a schema that is the superset of all documents
- handling unexpected first level data structures: initially we should error if any element of the top-level array is either a primitive value (number string, etc.) or an array.
- nested data structures: arrays can become lists, documents can become structs or stay as json strings
- type mismatch: error/abort of a value appears as incompatible types in different documents, it can be an error, following existing rules is fine. 

## Future Work

- specifying some kind of common set of "reading documents into tables/dataframe" options, as schema inference is something we'll end up doing more of and we should make these arguments consistent.
- `read_json` function which reads the first character of the file, determine whether to use `ar` or `nd` rules.
- research: can streaming tables be used to have "partitions" with different schemas as part of the same addressable unit. 
- handling type mismatches: we should continue doing whatever we do now, but in the future it might be interesting to make the handling of situations where "field `id` is either a string or an integer"
- provide a column type that is document-structure aware (json, binary representation, etc.) with corresponding query options..
- advanced schema inference options: 
  - read the first `<n>` (configurable) documents and inferring the schema from those documents
  - schema inferred to be the common subset of all documents/fields.
  - some kind of random sampling, but otherwise the same as the section option.
  - user specified schema and types.
- nested type handling structures
  - documents/arrays as arrow structs (including top-level, for a sort of document-database feel) 
  - flatten documents into columns with dotted notation.
- type conversions or handling of different type-mismatches.
  - option to insert nulls for rows where the type is a mismatch
  - devolve to a document-encoding-as-string.
  - schema inference should upscale numeric types and permit upcasting
[...]

PR #2306, opened 2023-12-26
feat: data source from JSON array data

Lilit0x:
Closes: #2218

## How to Test
`SELECT * FROM read_arjson('path/to/file/or/urls', max_size, <creds>)`


tychoish (commented):
I'm super excited about this! 

I wonder if implementing the object store's file type ends up having you write more boilerplate than is needed. 

I had great luck recently using the `StreamingTable` infrastructure in the bson functionality, though I don't know if it makes a lot of sense (also your code supports gzipped inputs and the BSON code does not.)

crates/sqlbuiltins/src/functions/table/arr_json/exec.rs
```diff
@@ -0,0 +1,262 @@
...
+        let batch_size = self.batch_size;
+        let file_compression_type = self.file_compression_type.to_owned();
+        Ok(Box::pin(async move {
+            let r = store.get(file_meta.location()).await?;
+            let stream = stream! {
+                 match r.payload {
+                GetResultPayload::File(file, _) => {
+                    let decoder = file_compression_type.convert_read(file)?;
+                    let mut reader = BufReader::new(decoder);
+                    let values: Value = serde_json::from_reader(&mut reader)
```
universalmind303, line 241: I'm partial to using using simd-json here. It's generally faster out of the box for reads, and it has a [known-key](https://github.com/simd-lite/simd-json#known-key) feature that makes it even more performant when reading with a known set of keys. I also had much success using it as the base for the json readers in polars

Lilit0x:
> I wonder if implementing the object store's file type ends up having you write more boilerplate than is needed.
You mean the `FileFormat`? Yes, it did. Especially in the `create_physical_plan` function because of the execution. I tried to look for a way around that, but I couldn't since the default `JsonFormat` is new line delimited which gives more credibility to @universalmind303's suggestion to making the change upstream. For now though, I think it is okay like this, albeit, I'll still have to write more for the `create_writer_physical_plan`. 
The other option I had was to create a different type similar to `FileType` and then implement all the traits that datafusion's `FileType` did, but that would have been extremely longer and probably unnecessary. 

> I had great luck recently using the `StreamingTable` infrastructure in the bson functionality, though I don't know if it makes a lot of sense (also your code supports gzipped inputs and the BSON code does not.)
Hmm, I looked through it, I can't seem to see where and how it will fit this use case. Maybe I am missing something.

crates/sqlbuiltins/src/functions/table/arr_json/exec.rs
Lilit0x: This is interesting. And it is compatible with `serde` out of the box. 
Thank youuu!.

tychoish:
I think given that df419aa51d31d99f0d95eb2d2a4b788a3c010ae7 has landed, we should let this one drop. 

I think there's a lot of merit to this approach, and I'm super thankful that you did this work for us to learn from, and sorry that we didn't use the code directly.

PR #2307, opened 2023-12-27
feat: cli highlighting improvements.

universalmind303:
a number of improvements to the cli highlighter. 

- better support for types
- more support for various symbols `&|+-/%` 
- function highlighting!

demo

### DIFF
--- a/crates/glaredb/src/local.rs
+++ b/crates/glaredb/src/local.rs
@@ -143,7 +143,6 @@ impl LocalSession {
             .with_validator(Box::new(SQLValidator));
 
         let prompt = SQLPrompt {};
-        let mut scratch = String::with_capacity(1024);
 
         loop {
             let sig = line_editor.read_line(&prompt);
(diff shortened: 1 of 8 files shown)


crates/glaredb/src/local.rs
```diff
@@ -157,26 +156,14 @@ impl LocalSession {
...
-                        if second.is_some() {
-                            match self.execute(&scratch).await {
-                                Ok(_) => {}
-                                Err(e) => println!("Error: {e}"),
-                            };
-                            scratch.clear();
-                        } else {
-                            scratch.push(' ');
-                        }
+                        match self.execute(&buffer).await {
```
universalmind303, line 159: all of this is handled in the `Validator` now. Anything that makes it this point is already validated as `Complete`

testdata/sqllogictests/special_characters.slt
```diff
@@ -0,0 +1,9 @@
+# https://github.com/GlareDB/glaredb/issues/2309
+statement ok
+select ';';
+
+statement ok
+select ';"';
+
+statement error Schema error: No field named ";".
+select ";";
```
scsmithr, line 9: Since these are going of the postgres protocol, these queries should already be working correctly since there's no checks for ';' in that code path. 

It's likely we'll want a cli test for checking that changes in this pr fix https://github.com/GlareDB/glaredb/issues/2309.

universalmind303: I added a bunch of unit tests [here](https://github.com/GlareDB/glaredb/pull/2307/files#diff-99e5c07101a26683f4a227db867a7280be5cbbe8426bd9ed377ca8aa88e45cb3R304) that test the completion logic. 

I'll also add one or two integration tests just to be safe.

universalmind303: > Since these are going of the postgres protocol, 

also keep in mind, all of these tests are also ran over the flightsql protocol.

scsmithr (approved):

PR #2310, opened 2023-12-27
feat: Add `read_sqlserver` table returning function

vrongmeal:
```sql
SELECT * FROM read_sqlserver(
    -- Connection string
    'server=tcp:localhost,1433;user=SA;password=bigStrong@1Pwd;TrustServerCertificate=true',
    -- Schema name
    'dbo',
    -- Table name
    'my_table'
);
```


vrongmeal:
I am setting this to auto-merge. There is nothing new in the code, like other functions like `read_postgres`.

PR #2313, opened 2023-12-27
fix: bson type handling cleanup

tychoish:


tychoish:
I expect the mongodb tests might `fail,` because I removed an "empty-string-to-null" converter in the schema inference and an incompatibility between the CSV inference (empty cells are empty strings) and our assumption (here) that "empty cells are nulls". I think empty cells should probably be nulls, but this seems tractable to fix (potentially in testing).

scsmithr:
For reference, our current csv inference will put nulls for empty strings

```
> select * from 'test.csv';                                                                          
┌───────┬───────┬───────┐
│     a │     b │     c │
│    ── │    ── │    ── │
│ Int64 │ Int64 │ Int64 │
╞═══════╪═══════╪═══════╡
│     1 │     2 │     3 │
│  NULL │     4 │     5 │
│     6 │  NULL │     7 │
│     8 │     9 │  NULL │
└───────┴───────┴───────┘
```

test.csv:

```
a,b,c
1,2,3
,4,5
6,,7
8,9,
```

scsmithr (approved):

tychoish:
I realized that the bug wasn't _our_ csv handling but the way that the test fixture used `mongoimport` and so by adding a new option there I was able to remove the work around.

PR #2316, opened 2023-12-28
fix: Pushdown SQL Server `WHERE` clause

### LINKED ISSUE #2094: feat: SQL Server followup items
## Description

Initial support added in https://github.com/GlareDB/glaredb/issues/2093. This lays out some follow-up items.

- [x] `read_sqlserver` function: #2310
- [x] Push down expression filters (WHERE): #2316
- [x] `datatype` table SLT tests: #2317 
- [x] Virtual schema thing (for listing tables, column info): #2311

vrongmeal:
Fixes #2094

### DIFF
--- a/crates/datasources/src/sqlserver/mod.rs
+++ b/crates/datasources/src/sqlserver/mod.rs
@@ -22,7 +22,6 @@ use datafusion::physical_plan::{
     DisplayAs, DisplayFormatType, ExecutionPlan, Partitioning, RecordBatchStream,
     SendableRecordBatchStream, Statistics,
 };
-use datafusion::scalar::ScalarValue;
 use datafusion_ext::errors::ExtensionError;
 use datafusion_ext::functions::VirtualLister;
 use datafusion_ext::metrics::DataSourceMetricsStreamAdapter;
@@ -35,7 +34,7 @@ use tokio_util::compat::TokioAsyncWriteCompatExt;
 use tracing::warn;
 
 use std::any::Any;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::fmt::{self, Write};
 use std::pin::Pin;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


crates/datasources/src/sqlserver/mod.rs
```diff
@@ -341,6 +350,76 @@ impl TableProvider for SqlServerTableProvider {
...
+                return Ok(false);
+            }
+        }
+        Expr::IsFalse(expr) => {
+            if write_expr(expr, buf)? {
+                write!(buf, " IS FALSE")?;
+            } else {
+                return Ok(false);
+            }
+        }
```
scsmithr, line 404: Could we add an SLT with a query that would trigger this? SQL Server doesn't have a bool data type, so I'm not positive this would work.

crates/datasources/src/sqlserver/mod.rs
```diff
@@ -341,6 +350,76 @@ impl TableProvider for SqlServerTableProvider {
...
+            ss.push(buf);
+            buf = String::new();
+        }
+    }
+
+    Ok(ss.join(" AND "))
+}
+
+/// Try to write the expression to the string, returning true if it was written.
+fn write_expr(expr: &Expr, buf: &mut String) -> Result<bool> {
```
universalmind303, line 369:
```suggestion
fn try_write_expr(expr: &Expr, buf: &mut String) -> Result<bool> {
```

crates/datasources/src/sqlserver/mod.rs
vrongmeal: Yeah, I just raised a PR to run tests in CI (some issue with local). I will keep resolving stuff accordingly. There are tests in `basic.slti` that will validate this.

crates/datasources/src/sqlserver/mod.rs
vrongmeal: Will rename!

scsmithr (approved):

PR #2317, opened 2023-12-28
test: Add datatypes test for SQL Server

vrongmeal:


crates/datasources/src/sqlserver/mod.rs
```diff
@@ -143,20 +144,21 @@ impl SqlServerAccessState {
...
                 ColumnType::Null => DataType::Null,
-                ColumnType::Bit => DataType::Boolean,
+                ColumnType::Bit | ColumnType::Bitn => DataType::Boolean,
                 ColumnType::Int1 => DataType::Int8,
                 ColumnType::Int2 => DataType::Int16,
                 ColumnType::Int4 => DataType::Int32,
                 ColumnType::Int8 | ColumnType::Intn => DataType::Int64,
-                ColumnType::Float4 => DataType::Float32,
-                ColumnType::Float8 | ColumnType::Floatn => DataType::Float64,
+                ColumnType::Float4 | ColumnType::Floatn => DataType::Float32,
```
scsmithr, line 152: `Floatn` should convert to `Float64` since it can hold more than 32 bits (and by default gets set to 53).

scsmithr: Ooook, seems like sql server tries to be smart about this.

We likely will need to implement our own `Floatn` type like what's done for `Intn`: https://github.com/glaredb/glaredb/blob/e16355bc679b532bcdf454bbfbd19ac397322e0c/crates/datasources/src/sqlserver/mod.rs#L611-L636

vrongmeal: Done!

PR #2318, opened 2023-12-29
feat: json and bson document unwinding

tychoish:
It's nifty (and I think good!) to expose document-native tools for
filtering and selecting data stored in document formats (as we have
and may do more of.) But why not convert documents to the DF Struct
type so that we can just use normal SQL. 

## Remaining work

- [ ] testing, obviously.
- [ ] make sure we're good with `unwind` as a name for this. 
- [ ] (maybe?) do something recursive in JSON (this patch doesn't do that yet.)
- [ ] come to some agreement about how to hanlde (potentially heterogeneous) arrays.


crates/datasources/src/bson/errors.rs
```diff
@@ -1,8 +1,14 @@
+use datafusion::arrow::datatypes::DataType;
 use datafusion::error::DataFusionError;
 use datafusion_ext::errors::ExtensionError;
 
 use crate::object_store::errors::ObjectStoreSourceError;
 
+/// Recursion limit for inferring the schema for nested documents.
+///
+/// The MongoDB kernel rejects nesting of greater than 100.
+pub const RECURSION_LIMIT: usize = 100;
```
scsmithr, line 10: I don't know if this has always been 100, but do we want to match mongo's nesting limit here? This _could_ potentially be very expensive on our side, all for just inferring the schema (I also have no data to back up how fast schema inference is).

crates/datasources/src/bson/errors.rs
```diff
@@ -35,6 +41,9 @@ pub enum BsonError {
         datafusion::arrow::datatypes::DataType,
     ),
 
+    #[error("value was type {0}, expected {1}")]
+    IncorrectType(DataType, DataType),
```
scsmithr, line 45: Personally I'd make this a struct like this:

```suggestion
    #[error("value was type {got}, expected {expected}")]
    IncorrectType{got: DataType, expected: DataType},
```

Makes it unambiguous which field is which.

crates/datasources/src/bson/schema.rs
```diff
@@ -50,14 +44,15 @@ fn bson_to_arrow_type(depth: usize, bson: &Bson) -> Result<DataType> {
         Bson::Double(_) => DataType::Float64,
         Bson::Boolean(_) => DataType::Boolean,
         Bson::Null => DataType::Null,
-        Bson::Int32(_) => DataType::Float64,
-        Bson::Int64(_) => DataType::Float64,
+        Bson::Int32(_) => DataType::Int32,
+        Bson::Int64(_) => DataType::Int64,
```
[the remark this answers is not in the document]
scsmithr: nice https://github.com/GlareDB/glaredb/pull/715/files#diff-8f50fe2c8818660375e9fab64c9f444d0a99bfdda52db4d9094e7f35a38f9529R91-R92

crates/sqlbuiltins/src/functions/scalars/unwind.rs
```diff
@@ -0,0 +1,181 @@
...
+
+use super::*;
+use datasources::bson::scalar as bson_scalar;
+
+pub struct Json;
+
+impl ConstBuiltinFunction for Json {
+    const NAME: &'static str = "unwind_json";
+    const DESCRIPTION: &'static str =
+        "Converts a json document in a string feilld to a struct field.";
```
scsmithr, line 13:
```suggestion
        "Converts a json document in a string field to a struct field.";
```

scsmithr:
> make sure we're good with unwind as a name for this.

I'm good with unwind, I'm even more good with `parse_*` since I think that would match what people would expect.

tychoish:
For array handling (and we should normalize here and this elsewhere), the options are:
- reject heterogeneous arrays (what happens here for bson), and convert all arrays to lists.
- do not attempt to unwind arrays, and leave these in their serialized forms (this is what happens here for json). 
- treat arrays and objects/documents (e.g. structs with keys that are either indexes (integers) or stringified integers). This is the internal representation of bson documents (string keys).

There exists the _broken_ implementation in the current MongoDB and BSON table funcs (which share code), which is that we decide the type of the array for the schema based on the type of the first element and then convert all values to strings.

Having implemented options one and two, I'm inclined to convert everything to option 3: 
- pros: doesn't lose data, doesn't error
- cons: potentially unintuitive. potentially inconsistent with native JSON handling. 

Thoughts?

scsmithr:
> reject heterogeneous arrays (what happens here for bson), and convert all arrays to lists.



> For array handling (and we should normalize here and this elsewhere), the options are:
> 
>     * reject heterogeneous arrays (what happens here for bson), and convert all arrays to lists.
> 
>     * do not attempt to unwind arrays, and leave these in their serialized forms (this is what happens here for json).
> 
>     * treat arrays and objects/documents (e.g. structs with keys that are either indexes (integers) or stringified integers). This is the internal representation of bson documents (string keys).
> 
> 
> There exists the _broken_ implementation in the current MongoDB and BSON table funcs (which share code), which is that we decide the type of the array for the schema based on the type of the first element and then convert all values to strings.
> 
> Having implemented options one and two, I'm inclined to convert everything to option 3:
> 
>     * pros: doesn't lose data, doesn't error
> 
>     * cons: potentially unintuitive. potentially inconsistent with native JSON handling.
> 
> 
> Thoughts?

Is the problem we're trying to solve here around being able to accurately infer the type of an array?

If so, I think we'll be hitting some amount of difficulty with getting accurate types, especially in the case of heterogenous arrays no matter if we go with option 1 or 3.

I have no issue with option 3, and I'd be down to explore that, especially if it makes the type problems easier.

crates/datasources/src/bson/errors.rs
tychoish: I have a fair amount of confidence in the rust bson library from a performance perspective, so I'm not super worried about that code (and the message serialization work is probably more expensive than our work here.) 

I think if we had more strict limits than MongoDB that could lead us to reject data in ways that would be more expensive, and that feels like it could be bad for users. 

MongoDB has rejected documents with more than 100 levels since 2013 or so, and I'm not aware of any real complaints. about it, so I think we don't need a larger limit. 100 Is sort of arbitrary, not _actually_ that big and mostly prevents us from blowing up the stack or letting a user bomb the stack.

scsmithr: Sounds good, we can stick with 100 here.

tychoish:
> Is the problem we're trying to solve here around being able to accurately infer the type of an array?

Yes. If you have `[1, true, null, "hello world", 0.34, {"a": 1}]` what do you do with this?

> If so, I think we'll be hitting some amount of difficulty with getting accurate types, especially in the case of heterogenous arrays no matter if we go with option 1 or 3.

Option 1 is just to error (and this would cause the users' query to fail. I think we _should_ do this rather than just ignore these values, _but_ it's bad. 

Option 3 couldn't error in this case. What is the difficulty you're referring to?

> I have no issue with option 3, and I'd be down to explore that, especially if it makes the type problems easier.

I think it does. The main questions/problems that it raises: 

- It might be unexpected to users.
- It might (probably?) would be inconsistent with how the JSON table functions and external data provider function.

crates/datasources/src/bson/errors.rs
tychoish: (I just duplicated this constant here, so it's consistent with the existing behavior)

crates/datasources/src/bson/errors.rs
tychoish: This does seem like a better solution, and I think I did this because I'm a `n00b`.

This duplicates an error in the `datasources::bson::errors::BsonError` enum, and so I'm inclined to come back and change it later when/where I can do both in a very boring PR.

tychoish:
> I'm good with unwind, I'm even more good with parse_* since I think that would match what people would expect.

I was mostly just stealing [this term](https://www.mongodb.com/docs/manual/reference/operator/aggregation/unwind/), which isn't a good fit, admittedly. 

Particularly if we do the thing with recursively parsing arrays into structs then it is closer to unwinding.

There's future work if people like this to let people project here, which would be kind of boss.

scsmithr:
> Yes. If you have [1, true, null, "hello world", 0.34, {"a": 1}] what do you do with this?

Idk, I remember doing the initial mongo implementation and just punting on it since I didn't have a good answer then (or now). So I'm definitely down to explore option 3 to get that to work.

> Option 1 is just to error (and this would cause the users' query to fail. I think we should do this rather than just ignore these values, but it's bad.
> 
> Option 3 couldn't error in this case. What is the difficulty you're referring to?

I meant to say that I think we would have the same amount of difficulty if we tried to use Lists or Structs for this since they'll both end up with some complicated types, because we _could_ generate a list data type that includes all possible field types with structs and unions.

> I think it does. The main questions/problems that it raises:
>
>    It might be unexpected to users.
>    It might (probably?) would be inconsistent with how the JSON table functions and external data provider function.

I'd want to think on this a bit more, I don't have a strongly formed opinion on this yet, and I'd want to see what other databases (like postgres) do here.

crates/datasources/src/bson/schema.rs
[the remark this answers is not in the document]
tychoish: 😂 

https://github.com/GlareDB/glaredb/actions/runs/7357998053/job/20030565482?pr=2318#step:8:804

tychoish:
> I meant to say that I think we would have the same amount of difficulty if we tried to use Lists or Structs for this since they'll both end up with some complicated types, because we could generate a list data type that includes all possible field types with structs and unions.

Oh I hadn't thought about the list/union situation. This is gross, but maybe cool. Let's call this option 4. 

> I'd want to think on this a bit more, I don't have a strongly formed opinion on this yet, and I'd want to see what other databases (like postgres) do here.

postgres supports heterogeneous arrays, probably doing something like option 4.

I'm still worried about sort of just having to copy what datafusion does. 

------

Regardless, I think we should probably do the better array handling as part of another PR.

scsmithr:
> Regardless, I think we should probably do the better array handling as part of another PR.

Completely agree with this.

crates/sqlbuiltins/src/functions/scalars/unwind.rs
```diff
@@ -0,0 +1,180 @@
...
+use datafusion::arrow::datatypes::Fields;
+use serde_json::{from_str, Map, Value};
+
+use super::*;
+use datasources::bson::scalar as bson_scalar;
+
+pub struct Json;
+
+impl ConstBuiltinFunction for Json {
+    const NAME: &'static str = "unwind_json";
```
universalmind303, line 11: I think grouping the `json` functions under a `json_` prefix improves discovery & autocomplete ux
```suggestion
    const NAME: &'static str = "json_unwind";
```

crates/sqlbuiltins/src/functions/scalars/unwind.rs
```diff
@@ -0,0 +1,180 @@
...
+        ))
+    }
+}
+
+impl BuiltinScalarUDF for Json {
+    fn as_expr(&self, args: Vec<Expr>) -> Expr {
+        let udf = ScalarUDF {
+            name: Self::NAME.to_string(),
+            signature: ConstBuiltinFunction::signature(self).unwrap(),
+            return_type: Arc::new(|_| Ok(Arc::new(DataType::Struct(Fields::empty())))),
```
universalmind303, line 34: I'm pretty sure this'll break during logical planning due to datafusion's logical planner not having the correct return type.

datafusion doesn't yet support dynamic return types for udfs. 

See https://github.com/apache/arrow-datafusion/issues/8624

This will likely have to be done outside of the 
`BuiltinScalarUDF` similar to how [arrow_cast](https://github.com/apache/arrow-datafusion/blob/b85a39739e754576723ff4b1691c518a86335769/datafusion/sql/src/expr/arrow_cast.rs#L55-L56) is handled.

There is a chance that datafusion planner doesn't check struct fields during logical planning in which case we are ok, but we'll want to extensively test this.

tychoish: noted.

crates/sqlbuiltins/src/functions/scalars/unwind.rs
tychoish: We're pretty inconsistent with this historically: `read_x` vs `scan_x`, etc. I think `read_x` won out anyway (and this is consistent with Duck). 

function names should generally be verb phrases... "do_thing" ... rather than a description of the the outcome.... Functions are named `add` or `add_integers` not `integer_add`.

In many cases autocomplete is a fuzzy search anyway so I think it's fine.

universalmind303:
> > make sure we're good with unwind as a name for this.
> 
> I'm good with unwind, I'm even more good with `parse_*` since I think that would match what people would expect.

+1 for `parse_*`. 

FWIW, that's what snowflake calls this function as well.

tychoish:
Going to close. I don't think writing an arrow_cast-type function makes sense right now, and despite some poking, I don't think there's much to be done.

PR #2334, opened 2023-12-31
fix: bson builder should handle schema flexibly

tychoish:
The `RecordStructBuilder` for bson previously assumed (as is the case
with our MongoDB implementation) that it would never see bson
documents with fields that didn't appear in the schema/projection and
would error otherwise.

While it does mean that the schema inference (which for MongoDB is
done somewhat probabalistically) controls the projection, it does mean
that you will never see documents that have fields that aren't in the
schema. 

For handling the pure-BSON document streams, it _does_ mean that it
would be very easy to see a document that has fields that aren't in
the schema. This increases with #2333, but could easily happen
otherwise.


scsmithr:
> While it does mean that the schema inference (which for MongoDB is
done somewhat probabalistically) controls the projection, it does mean
that you will never see documents that have fields that aren't in the
schema.

Is this saying that the current implementation in main would skip bson documents that have unexpected schemas? And the fix here is we can now sanely handle those documents? What happens to the extra columns/fields? Are those just dropped?

Also would we be able to throw together a quick test for this?

crates/datasources/src/bson/builder.rs
```diff
@@ -102,10 +100,48 @@ impl RecordStructBuilder {
...
+    }
+
+    pub fn project_and_append(&mut self, doc: &RawDocument) -> Result<()> {
+        let mut cols_set: BitVec<u8, Lsb0> = BitVec::repeat(false, self.fields.len());
+
+        for iter_result in doc {
+            match iter_result {
+                Ok((key, val)) => {
+                    if let Some(idx) = self.field_index.get(key) {
+                        let idx = idx.to_owned();
```
scsmithr, line 118:
```suggestion
                    if let Some(&idx) = self.field_index.get(key) {
```

Since it's a pattern.

Also for `&usize` (and `&i64`, etc) you can dereference with `*idx` instead of having to do `let idx = idx.to_owned()`.

tychoish:
> Is this saying that the current implementation in main would skip bson documents that have unexpected schemas? And the fix here is we can now sanely handle those documents? What happens to the extra columns/fields? Are those just dropped?

I believe that the current mongodb implementation doesn't have this problem because we project out the fields that we don't expect to see, so that we're never in a case where a document would have a field we don't expect (caveat: nested documents/arrays are an edge case that I'd want to think about.) 

The failure mode, is that the query will fail, because we'll be building a result set, see a field we don't expect to see and return an error, which could happen with bson-files. 

For MongoDB if the schema inference does not pick up a field, the _query_ will remove that field from the result set, so we _are_ already dropping data silently (not great, but the solution is to not infer schema, (addressable with #2333),) so this just brings the implementations on par with each other.

> Also would we be able to throw together a quick test for this?

So the bug, (which to be fair, isn't released), requires the schema inference to be "wrong," (or incomplete) relative to our expectations. Building a test case that has that would be pretty fragile (all of the new code is hit by the existing test, so I'm not worried about this being worse than baseline.)

universalmind303:
can we add some tests for this.

tests/tests/test_bson.py
```diff
@@ -88,3 +88,96 @@ def test_read_bson(
                 assert len(row) == 5
                 assert row["beatle_name"] in beatles
                 assert beatles.index(row["beatle_name"]) == row["beatle_idx"] - 1
+
+
+def test_read_bson(
```
universalmind303, line 93: from looking at the implementation, i'd expect that we'd be able to write a test that has duplicate fields and assert that the value is the first one, but not the second.

tychoish: it's not possible to use to python bson library to build these. I wrote a little go script and generated a file that I committed. 

I'm not sure if we should commit/run it, but it's there to prove the point.

tests/tests/dupes.bson
scsmithr, line 1: I'd prefer having this in `testdata/bson` as to match the other test files we have.

tests/tests/scripts/make_bad_bson.go
scsmithr, line 1: I think it's good to commit this, but not run in CI, so we have _something_ to trace the lineage of test data. I committed `testdata/generate_iceberg.py` for a similar reason.

tests/tests/dupes.bson
scsmithr: Only sticking point, rest looks good to me.

tychoish: match how?

scsmithr: Meant that's just where we keep testdata in various formats.

tychoish:
In any case, I wrote the duplicate fields always picks the first one test as a rust unittest.

scsmithr (approved):

PR #2338, opened 2024-01-02
docs: improve readme

universalmind303:
So i think our readme has been getting a bit cluttered. This is an attempt to strip it down a little bit & keep only the most relevant information.

Most importantly, it moves our "about" section to the header, and adds some of the most important links to there as well. 

the `## Your first data source` section was removed as it's already included in our docs. 

Note:
it does use some github markdown specific syntax such as `[!TIP]`, so you should definitely review it with that in mind.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,33 +2,29 @@
 <!-- markdownlint-disable first-line-h1 -->
 <div align="center">
   <a href="https://glaredb.com#gh-light-mode-only">
-    <img src="https://docs.glaredb.com/assets/logo.svg" height="44">
+    <img src="https://docs.glaredb.com/assets/logo.svg" height="60q">
   </a>
   <a href="https://glaredb.com#gh-dark-mode-only">
-    <img src="https://glaredb.com/logo.svg" height="44">
+    <img src="https://glaredb.com/logo.svg" height="60">
   </a>
 </div>
-
-<!-- Adds some spacing between logo and badges. -->
-<p></p>
+<br/>
+<p align="center">
+Data exists everywhere: on laptops, in databases like Postgres and Snowflake, and as files in S3, in various formats including Parquet, CSV, and JSON. Accessing and analyzing this data typically involves multiple steps across different systems.<br/> <br/> <b>GlareDB is designed to query your data wherever it lives using SQL that you
+already know.</b>
+</p>
+<p align="center">
+<a href="https://docs.glaredb.com">Documentation</a> •
+<a href="https://console.glaredb.com>">GlareDB Cloud</a> •
+<a href="https://glaredb.com/blog">Blog</a>
+</p>
 
 <div align="center">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,11 +34,12 @@ Install/update `glaredb` in the **current directory**:
 curl https://glaredb.com/install.sh | sh
 ```
 
-It may be helpful to install the binary in a location on your `PATH`. For
-example, `~/.local/bin`.
+> [!TIP]
+> It may be helpful to install the binary in a location on your `PATH`. For example, `~/.local/bin`.
+>
+> If you prefer manual installation, download, extract and run the GlareDB binary
+> from a release in our [releases page].
```
tychoish, line 41:
```suggestion
> If you prefer manual installation, download, extract, and run the GlareDB binary
> from a release in our [releases page].
```

README.md
```diff
@@ -94,17 +85,16 @@ Read our [announcement on Hybrid Execution] for more information.
...
 2. Import and use `glaredb`.
 
-     ```python
-     import glaredb
-     con = glaredb.connect()
-     con.sql("select 'hello world';").show()
-     ```
+   ```python
+   import glaredb
+   glaredb.sql("select 'hello world';").show()
```
tychoish, line 96: I think I'd inline/chain a connect method: without this we don't establish that you can connect to cloud, and when people start doing stateful things, not being bound to a connection object is confusing.

README.md
```diff
@@ -115,103 +105,46 @@ con = glaredb.connect("glaredb://user:pass@host:port/deployment")
...
-When launched as a server process, GlareDB can be reached on port `6543` using a
-Postgres client. The following example uses `psql` to connect to a locally
-running server:
 
 ```shell
+# Connect to the server using psql client
 psql "host=localhost user=glaredb dbname=glaredb port=6543"
 ```
 
-## Your first data source
```
tychoish, line 162: I wouldn't drop this (and might even move it above 'running in server mode') as it's our main chance to demonstrate what glaredb is and what makes it cool.

README.md
```diff
@@ -2,34 +2,30 @@
...
+    <img src="https://glaredb.com/logo.svg" height="60">
   </a>
 </div>
-
-<!-- Adds some spacing between logo and badges. -->
-<p></p>
+<br/>
+<p align="center">
+Data exists everywhere: on laptops, in databases like Postgres and Snowflake, and as files in S3, in various formats including Parquet, CSV, and JSON. Accessing and analyzing this data typically involves multiple steps across different systems.<br/> <br/> <b>GlareDB is designed to query your data wherever it lives using SQL that you
+already know.</b>
```
tychoish, line 14: I don't think I would be inclined to put this here, but if you are, can you use paragrpah tags and hard wrap it like the rest of the file, rather than doing the `br` thing.

tychoish (changes requested):

Issue #2315, opened 2023-12-27
Clickhouse followup items


scsmithr:
Follow up items for Clickhouse support added in https://github.com/GlareDB/glaredb/pull/2300.

- [x] Virtual listing (and related SLTs)
- [x] Handle empty tables 

    Currently we require that tables have at least one item since we're detecting the table schema from the blocks that clickhouse returns. Clickhouse does not seem to return blocks on queries that return nothing.

    https://github.com/GlareDB/glaredb/blob/c50dda2d61ffd162f05cbc08c536f934b09488b5/crates/datasources/src/clickhouse/mod.rs#L81-L84

- [x] Add `WHERE` clause in the scan
- [x] Add `LIMIT` clause in the scan

    https://github.com/GlareDB/glaredb/blob/c50dda2d61ffd162f05cbc08c536f934b09488b5/crates/datasources/src/clickhouse/mod.rs#L230-L232

- [x] Data types SLTs
- [ ] Ensure this works with clickhouse cloud

PR #2340, opened 2024-01-04
feat: Clickhouse followup items

vrongmeal:
See tests for changes to read function (schema optional).

To connect to cloud:

```
> CREATE EXTERNAL DATABASE click_cloud
::: FROM clickhouse (
::: connection_string 'clickhouse://default:<PWD>@gfvi4aovgg.us-east1.gcp.clickhouse.cloud/?secure'
::: );
Database created
```

Fixes: #2315

### DIFF
--- a/crates/datasources/src/clickhouse/mod.rs
+++ b/crates/datasources/src/clickhouse/mod.rs
@@ -28,6 +28,7 @@ use std::any::Any;
 use std::borrow::Cow;
 use std::fmt::{self, Display, Write};
 use std::sync::Arc;
+use std::time::Duration;
 use url::Url;
 
 use crate::clickhouse::stream::BlockStream;
@@ -86,8 +87,54 @@ impl ClickhouseAccessState {
             )));
         }
 
-        let addr = conn_str.clone(); // Let it take the address from original URL
-        let mut opts = Options::new(addr).pool_min(1).pool_max(1);
+        let mut secure = false;
+        for (key, _val) in conn_str.query_pairs() {
+            if key.as_ref() == "s" || key.as_ref() == "secure" {
+                secure = true;
+                break;
+            }
+        }
+
+        let addr = {
+            let port = if let Some(port) = conn_str.port() {
+                port
+            } else if secure {
+                9440
+            } else {
+                9000
+            };
+            if port == 9440 && !secure {
+                // We assume that if someone is trying to connect to port 9440,
+                // they are trying to connect to the Clickhouse Cloud on the
+                // TLS port. Due to a panic, we error early. In a rare scenario,
+                // this might be incorrect. So we should fix and push the fix
+                // upstream.
+                //
+                // Issue to track: https://github.com/GlareDB/glaredb/issues/2360
+                return Err(ClickhouseError::String(
+                    "Cannot connect to SSL/TLS port without secure parameter enabled.
+Enable secure param in connection string:
+    `clickhouse://<user>:<password>@<host>:<port>/?secure`"
+                        .to_string(),
+                ));
+            }
+            let mut addr = if let Some(host) = conn_str.host_str() {
+                format!("tcp://default@{host}").parse::<Url>()?
+            } else {
+                "tcp://default@127.0.0.1".parse::<Url>()?
+            };
+            addr.set_port(Some(port)).map_err(|_| {
+                ClickhouseError::String("unable to set port for clickhouse URL".to_string())
+            })?;
+            addr
+        };
+
+        let mut opts = Options::new(addr)
+            .pool_min(1)
+            .pool_max(1)
+            .secure(secure)
+            .ping_timeout(Duration::from_secs(5))
+            .connection_timeout(Duration::from_secs(5));
 
         if let Some(mut path) = conn_str.path_segments() {
             if let Some(database) = path.next() {
@@ -102,12 +149,6 @@ impl ClickhouseAccessState {
 
         if let Some(password) = conn_str.password() {
             opts = opts.password(password);
-        }
-
-        for (key, _val) in conn_str.query_pairs() {
-            if key.as_ref() == "s" || key.as_ref() == "secure" {
-                opts = opts.secure(true);
-            }
         }
 
         let pool = Pool::new(opts);
--- a/crates/datasources/src/clickhouse/mod.rs
+++ b/crates/datasources/src/clickhouse/mod.rs
@@ -149,8 +149,9 @@ impl ClickhouseAccessState {
                     SqlType::Float64 => DataType::Float64,
                     SqlType::String | SqlType::FixedString(_) => DataType::Utf8,
                     // Clickhouse has both a 'Date' type (2 bytes) and a
-                    // 'Date32' type (4 bytes). I think they're both represented
-                    // with this one variant.
+                    // 'Date32' type (4 bytes). The library doesn't support
+                    // 'Date32' type yet.
+                    // TODO: Maybe upstream support for Date32?
                     SqlType::Date => DataType::Date32,
                     SqlType::DateTime(DateTimeType::DateTime32) => {
                         DataType::Timestamp(TimeUnit::Second, None)
--- a/crates/datasources/src/clickhouse/mod.rs
+++ b/crates/datasources/src/clickhouse/mod.rs
@@ -31,6 +31,7 @@ use std::sync::Arc;
 use url::Url;
 
 use crate::clickhouse::stream::BlockStream;
+use crate::common::util;
 
 #[derive(Debug, Clone)]
 pub struct ClickhouseAccess {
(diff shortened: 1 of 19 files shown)


crates/datasources/src/clickhouse/mod.rs
```diff
@@ -69,28 +75,54 @@ struct ClickhouseAccessState {
...
-            None => {
-                return Err(ClickhouseError::String(
-                    "unable to determine schema for table, no blocks returned".to_string(),
-                ))
-            }
-        };
+        let block = client
+            .query(format!("SELECT * FROM {table_ref} LIMIT 0"))
+            .fetch_all()
+            .await?;
```
scsmithr, line 125: Just want to confirm, this is actually working with the `LIMIT 0` part? When I was trying to stream blocks, I wasn't getting anything. Is `fetch_all` behaving differently here?

crates/datasources/src/clickhouse/mod.rs
```diff
@@ -361,3 +454,40 @@ impl fmt::Debug for ClickhouseExec {
             .finish_non_exhaustive()
     }
 }
+
+#[derive(Debug, Clone)]
+pub struct ClickhouseTableRef<'a> {
+    schema: Option<Cow<'a, str>>,
```
scsmithr, line 460: I'd make this field `database` so that it's consistent with clickhouse terminology (and we just have a comment saying a database in clickhouse is similar to a schema elsewhere).

crates/datasources/src/clickhouse/mod.rs
vrongmeal: Yeah, this works. `fetch_all` does return a block, no matter how many rows.

scsmithr: Nice, happy that this works.

crates/datasources/src/clickhouse/mod.rs
```diff
@@ -117,32 +150,18 @@ impl ClickhouseAccessState {
...
-                        };
-
-                        let tz: Arc<str> = tz.to_string().into_boxed_str().into();
-
-                        DataType::Timestamp(unit, Some(tz))
+                    SqlType::DateTime(DateTimeType::DateTime64(_precision, tz)) => {
+                        // We store all timestamps in nanos, so the precision
+                        // here doesn't matter. clickhouse_rs crate handles this
+                        // for us.
+                        DataType::Timestamp(TimeUnit::Nanosecond, Some(tz.to_string().into()))
```
tychoish, line 164: it matters, I think, more what _they_ store timestamps as; and it doesn't look like we're giving (or getting) information from `clickhouse_rs` about what they store as.

crates/datasources/src/clickhouse/mod.rs
```diff
@@ -117,32 +150,18 @@ impl ClickhouseAccessState {
...
                     // Clickhouse has both a 'Date' type (2 bytes) and a
-                    // 'Date32' type (4 bytes). I think they're both represented
-                    // with this one variant.
+                    // 'Date32' type (4 bytes). The library doesn't support
+                    // 'Date32' type yet.
+                    // TODO: Maybe upstream support for Date32?
                     SqlType::Date => DataType::Date32,
                     SqlType::DateTime(DateTimeType::DateTime32) => {
-                        DataType::Timestamp(TimeUnit::Second, None)
+                        DataType::Timestamp(TimeUnit::Nanosecond, None)
```
tychoish, line 158: 32 bits of nanoseconds is not very much time. this might be right, but it _looks_ wrong, and the comment doesn't clarify enough.

are there tests over this?

tychoish (commented):
I find myself being confused by the timestamp casting, and I think some more clarification there might be useful.

crates/datasources/src/clickhouse/mod.rs
vrongmeal: So, we essentially convert everything to `chrono::NaiveDateTime` and get `seconds` OR `nanos` from there and we store it in 64 bit (even though clickhouse returns data as 32 bits here).

crates/datasources/src/clickhouse/mod.rs
vrongmeal: Nanos is the most precise clickhouse can go. It’s essentially a design choice that we are making here. It’s easier if the whole codebase uses nanos instead of anything else. We return accurate values but yeah, we don’t exactly translate clickhouse. But this doesn’t seem like an issue (to me). Wondering… are there any real disadvantages to this?

scsmithr (approved):

crates/datasources/src/common/util.rs
```diff
@@ -87,6 +88,14 @@ pub fn encode_literal_to_text(
             let naive = Utc.timestamp_nanos(*v * 1_000).naive_utc();
             encode_utc_timestamp(buf, &naive, tz.is_some())?;
         }
+        ScalarValue::TimestampMillisecond(Some(v), tz) => {
+            let naive = Utc.timestamp_nanos(*v * 1_000_000).naive_utc();
```
universalmind303, line 92:
```suggestion
            let naive = Utc.timestamp_millis_opt(*v).unwrap().naive_utc();
```

crates/datasources/src/common/util.rs
```diff
@@ -87,6 +88,14 @@ pub fn encode_literal_to_text(
...
+            encode_utc_timestamp(buf, &naive, tz.is_some())?;
+        }
+        ScalarValue::TimestampSecond(Some(v), tz) => {
+            let naive = Utc.timestamp_nanos(*v * 1_000_000_000).naive_utc();
```
universalmind303, line 96:
```suggestion
            let naive = Utc.timestamp_opt(*v).unwrap().naive_utc();
```

scripts/create-test-clickhouse-db.sh
```diff
@@ -45,9 +45,51 @@ done
...
+
+      -- Dates and times
+      '1999-09-30',
+      -- '1999-09-30', -- Date32
+      '1999-09-30 16:32:34',
+      '1999-09-30 16:32:34.123456',
+      -- with timezones
+      '1999-09-30 16:32:34',
+      '1999-09-30 16:32:34.123456'
+  )"
```
universalmind303, line 84: can we move these INSERT statements into `setup-clickhouse.sql`

universalmind303 (changes requested):
A few small changes, but overall looks good!

crates/datasources/src/clickhouse/mod.rs
tychoish: What gets easier if our entire codebase returns nanos instead of anything else?

I think we should maintain whatever precision the data source natively has and not try to cast numbers around to make things more consistent..

It seems like we could get ourselves into unexpected overflows or have users get unexpected results if their timestamp resolution changes, mostly for the cost of a few arms on a match statement, but I might be missing something.

crates/datasources/src/clickhouse/mod.rs
tychoish: I would think that this would break if we/when we support the write path.

crates/datasources/src/clickhouse/mod.rs
vrongmeal: > What gets easier if our entire codebase returns nanos instead of anything else?

There’s a lot of stuff (IIRC) that DF has issues with (when it comes to types). Even this change is something I made when I encountered an error. But that’s a good point that we might get a weird overflow someday so I’ll create a different issue for now.

This might not be very convincing but we do have: https://github.com/GlareDB/glaredb/blob/bf86b29349fd319754ef4b7dca76162fc232fd5c/crates/datasources/src/common/util.rs#L141-L163

This existed for a good reason which in hindsight I should have noted down…

crates/datasources/src/clickhouse/mod.rs
vrongmeal: What “write path”?

crates/datasources/src/clickhouse/mod.rs
vrongmeal: Opened https://github.com/GlareDB/glaredb/issues/2369

testdata/sqllogictests_clickhouse/datatypes.slt
```diff
@@ -0,0 +1,17 @@
...
+statement ok
+CREATE EXTERNAL TABLE datatypes
+	FROM clickhouse
+	OPTIONS (
+		connection_string = '${CLICKHOUSE_CONN_STRING}',
+		table = 'datatypes'
+	);
+
+# Check if we can fetch contents of the datatype table.
+query ITIIIIIIIIRRTTTTTTT rowsort
```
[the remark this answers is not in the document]
vrongmeal: Oh okay! Will add that.

vrongmeal:
@universalmind303 addressed your comments.

***

Is this good to merge?

crates/datasources/src/clickhouse/mod.rs
tychoish: like, inserting into clickhouse: if we change the types, round-tripping data will be hard (and also if their underlying type is in seconds, things could overflow in glaredb that wouldn't overflow in the source.)

tychoish (approved):

testdata/sqllogictests_clickhouse/datatypes.slt
```diff
@@ -0,0 +1,47 @@
...
+	arrow_typeof(c6),
+	arrow_typeof(c7),
+	arrow_typeof(c8),
+	arrow_typeof(c9),
+	arrow_typeof(c10),
+	arrow_typeof(c11),
+	arrow_typeof(c12),
+	arrow_typeof(c13),
+	arrow_typeof(c14),
+	-- arrow_typeof(c15),
```
universalmind303, line 39: why is this one skipped?

universalmind303 (approved):

testdata/sqllogictests_clickhouse/datatypes.slt
vrongmeal: There isn't a column called `c15` due to lack of support in the clickhouse library.

Issue #2345, opened 2024-01-05
feat(cassandra): `EXTERNAL TABLE` support for cassandra


universalmind303:
## Description

follow up to https://github.com/GlareDB/glaredb/issues/2332

Allow for `create external table` syntax for cassandra tables

PR #2361, opened 2024-01-05
feat: external table for cassandra

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2345


crates/protogen/proto/metastore/options.proto
```diff
@@ -229,6 +230,12 @@ message TableOptionsClickhouse {
   string table = 2;
 }
 
+message TableOptionsCassandra {
+  string connection_string = 1;
```
scsmithr, line 234: I'd prefer `host` here since this isn't really a connection string, but a host name/ip to a known node.

scsmithr (commented):
Structure looks good, just want to change "connection_string" to "host"

vrongmeal (approved):
Looks good to me.

Issue #2362, opened 2024-01-05
chore(ci): move cassandra slt's to their own stage


universalmind303:
## Description
cassandra SLT's are pretty slow. Let's pull them out into their own job like we did with snowflake

PR #2363, opened 2024-01-05
ci: speed up ci by introducing precompiled slt binary

universalmind303:
billable time & wall time are both down
# Before
image
image

image

# After
image


crates/testing/src/main.rs
```diff
@@ -0,0 +1,27 @@
...
+        .test_files_dir(test_dir)?
+        // Rust tests
+        .test("sqllogictests/ssh_keys", Box::new(SshKeysTest))?
+        .test("pgproto/binary_encoding", Box::new(PgBinaryEncoding))?
+        // Add hooks
+        .hook("*", Arc::new(AllTestsHook))?
+        // SSH Tunnels hook
+        .hook("*/tunnels/ssh", Arc::new(SshTunnelHook))?
+        .run()
+}
```
universalmind303, line 27: we could shave another minute off our builds if we want to include this runner into the glaredb binary directly.

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2362

crates/testing/src/main.rs
tychoish: I was initially viscerally opposed to this idea, but as I look at this code and also think about it a bit more, I kind of like it. 

Potential caveats: 
- I think the reset of the team should definitely sign off on this. 😁 😓
- I think we should hide it from any help menus 
- I think there should be a pre-flight log warning message that says "the slt test runner is not supported and does not have the same stability guarantees as the rest of glaredb." 

Otherwise, I think maybe it should be in everything, including the release build, because:
- I'm a little afraid of having the dev build drift/vary that much from the test build. 
- making it super easy for users to run/validate a simple reproduction case using slt on their hardware/systems seems like it might be useful.

.github/workflows/ci.yaml
```diff
@@ -112,12 +148,24 @@ jobs:
...
             ~/.cargo/
+          key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
+      - uses: actions/cache@v3
+        name: build cache
+        with:
+          path: |
             target/
             !target/**/glaredb
+            !target/**/slt
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
```
tychoish, line 169: - I have a memory of each individual cache action invocation taking time, and so I remember in the past I combined them. Could you just check that we're not running into that? (It's possible that I'm being superstitious) 
- I think it makes sense to cache `~/.cargo` against a hash of the `Cargo.lock` file; but I'm not sure that makes the same amount of sense to cache the target directory against the same hash. Should we use the branch name so that different builds (are less likely) to overwrite each other?

tychoish (commented):
this looks great, a few questions really.

.github/workflows/ci.yaml
universalmind303: > I have a memory of each individual cache action invocation taking time,

they do take _some_ time, but its pretty negligible. The build cache is the biggest one, and takes like 40 seconds. the other caches add a total of <15 seconds to the build. 

But what they add up front, they save in the slt pipelines. Now since it is precompiled, we can skip all of the caches except the one containing the binary `slt cache`. So it really allows us to parallelize the stages a lot more without getting penalized by pulling down the entire cache every time. 

> but I'm not sure that makes the same amount of sense to cache the target directory against the same hash. Should we use the branch name so that different builds (are less likely) to overwrite each other?

I was thinking about this too, but without the build cache set to the same key as the cargo cache, the initial build wont benefit from incremental compilation. So it ends up being much slower. 

I definitely think we should experiment further with the caches. I'd be pretty happy with a <10 minute pipeline

crates/testing/src/main.rs
universalmind303: > Otherwise, I think maybe it should be in everything, including the release build,

I think this would be a pretty good candidate for a feature flag, as it's something thats unlikely to change very often & has a very clear boundary on when/why you'd want to enable the feature. But I'm open for suggestions.

.github/workflows/ci.yaml
tychoish: > I was thinking about this too, but without the build cache set to the same key as the cargo cache, the initial build wont benefit from incremental compilation.

I'm sort of just thinking outloud but could we handle `target/**/incremental` with the cargo cache and the rest of the build/target cache with the branch?

crates/testing/src/main.rs
```diff
@@ -0,0 +1,27 @@
+use std::sync::Arc;
+
+use clap::Parser as _;
```
scsmithr, line 3:
```suggestion
use clap::Parser;
```

crates/testing/src/main.rs
scsmithr: Would including the the runner in glaredb _just_ be for build times, or would there be some other benefits as well.

Some other questions:
- Why is this binary not built during the "build" step? Are we just missing a cargo profile or something?

testdata/sqllogictests/functions/csv_scan.slt
```diff
@@ -33,7 +33,7 @@ select count(*) from csv_scan(
 
 # Relative path
 query I
-select count(*) from csv_scan('../../testdata/sqllogictests_datasources_common/data/bikeshare_stations.csv')
+select count(*) from csv_scan('${PWD}/testdata/sqllogictests_datasources_common/data/bikeshare_stations.csv')
```
scsmithr, line 36: Some of these tests purposely use relative path segments to ensure we work with both relative and absolute paths.

crates/testing/src/main.rs
universalmind303: > Would including the the runner in glaredb just be for build times, or would there be some other benefits as well.

I think build times are one, but not just for CI, but also dev workflow. I often find myself in the workflow of: 
run slts
run local 
execute some queries for debugging
make small changes
repeat

which is not great because `run local` and `slt` trigger different build paths so it usually slows down development quite a bit going back and forth between the two. if slt was part of the main binary, it'd speed up this cycle quite a bit. 

> Why is this binary not built during the "build" step? Are we just missing a cargo profile or something?

We could throw it into the "build" step, but that wouldn't really add any benefits.  Since it's a separate crate/bin, we need to compile it separately.

testdata/sqllogictests/functions/csv_scan.slt
universalmind303: hmm... I'll have to think on that one a bit. Making it into a binary forced it to not use a fixed path, but instead use one passed in from the command line.

universalmind303: @scsmithr This should be resolved now. It uses the repo root instead of the path from `crates/testing/`, which is arguably better anyways.

.github/workflows/ci.yaml
universalmind303: i'm not sure if we'd want to try to cache certain parts of `target/`, it may break some things, but not entirely sure. We should follow up in a separate issue/pr.

justfile
```diff
@@ -113,6 +113,14 @@ rpc-tests: protoc
     'sqllogictests/describe_rpc' \
     'sqllogictests/allowed_operations'
 
+# Build a pre-compiled slt runner
+build-testrunner *args:
+  cargo build --bin slt -- {{args}}
+
+# Run SQL Logic Tests with a pre-compiled slt runner
+slt-bin *args:
```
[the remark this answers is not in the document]
universalmind303: So i think the dev workflow is slightly different as we usually want the most updated code when running the tests, but within the CI, we only want to build the code once.

crates/testing/src/slt/cli.rs
```diff
@@ -145,7 +145,8 @@ impl Cli {
         let mut tests: Vec<_> = if let Some(patterns) = &self.tests_pattern {
             let patterns = patterns
                 .iter()
-                .map(|p| glob::Pattern::new(p))
+                .map(|p| p.trim_end_matches(".slt"))
```
scsmithr, line 148: What's this for? Is it to allow for things like `just slt "sqllogictests/*.slt"`?

testdata/sqllogictests/catalog/functions.slt
```diff
@@ -54,5 +54,5 @@ select
 from glare_catalog.functions
 where function_name = 'read_parquet';
 ----
-read_parquet   table   Utf8/List<Utf8>   t   SELECT * FROM read_parquet('./my_data.parquet')   Returns a table by scanning the given Parquet file(s).
-read_parquet   table   Utf8/List<Utf8>   t   SELECT * FROM read_parquet('./my_data.parquet')   Returns a table by scanning the given Parquet file(s).
+read_parquet table Utf8/List<Utf8> t SELECT * FROM read_parquet('./my_data.parquet') Returns a table by scanning the given Parquet file(s).
+read_parquet table Utf8/List<Utf8> t SELECT * FROM read_parquet('./my_data.parquet') Returns a table by scanning the given Parquet file(s).
```
scsmithr, line 58: I was separating by 3 spaces to make the distinction clearer between fields (especially since the description is a string containing spaces).

testdata/sqllogictests/create_table.slt
```diff
@@ -49,7 +49,7 @@ select * from ctas1;
 1
 
 statement ok
-create table ctas2 as select * from '../../testdata/parquet/userdata1.parquet';
+create table ctas2 as select * from '${PWD}/testdata/parquet/userdata1.parquet';
```
scsmithr, line 52: Is there a way for us to leave these as relative paths?

crates/testing/src/slt/cli.rs
universalmind303: yeah, i use tab completion a lot, which always puts in the `.slt`. This previously failed, now it just ignores the extension if its there.

testdata/sqllogictests/catalog/functions.slt
universalmind303: for some reason this was failing locally. Will look into it more.

scsmithr (commented):
If passes, lgtm

tychoish (approved):

PR #2364, opened 2024-01-07
chore: Allow specifying aliases for table functions

scsmithr:
Instead of duplicating the function definition itself.

Also a few cleanups.

Existing SLTs assert that this works as expected. SLTs for `parquet_scan`, etc are just copies of the `read_parquet` ones.


crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -19,6 +19,7 @@ use table::{BuiltinTableFuncs, TableFunc};
 
 /// Builtin table returning functions available for all sessions.
 static BUILTIN_TABLE_FUNCS: Lazy<BuiltinTableFuncs> = Lazy::new(BuiltinTableFuncs::new);
+// TODO: Why is this separate?
```
universalmind303, line 22: The arrow cast function has special handling in datafusion due to the dynamic return type. So it isn't a 'scalar_udf' and needs to be handled a bit differently.

scsmithr: This constant wasn't used, so removed it. Also added a comment when generating the hashmap about the arrow_cast special case.

PR #2367, opened 2024-01-08
chore: Misc builtin function refactoring

scsmithr:
- Adds back in examples/descriptions for some of the read_* functions
- Adds some tests asserting that we're actually getting entries in the functions table.
- Fixes panic when using non-table function as a table factor.

### DIFF
--- a/testdata/sqllogictests/catalog/functions.slt
+++ b/testdata/sqllogictests/catalog/functions.slt
@@ -42,7 +42,7 @@ where function_name = 'sum';
 sum   aggregate   Int8/Int16/Int32/Int64/UInt8/UInt16/UInt32/UInt64/Float32/Float64   t   sum(a)   Returns the sum of a column
 
 # Assert an arbitrary glaredb table function exists.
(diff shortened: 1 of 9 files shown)


scsmithr:
Punting duplicate stuff to this issue: https://github.com/GlareDB/glaredb/issues/2368

testdata/sqllogictests/catalog/functions.slt
```diff
@@ -28,31 +28,31 @@ where function_name = 'repeat';
...
+    function_type,
+    parameters,
+    builtin,
+    example,
+    description
+from glare_catalog.functions
+where function_name = 'read_parquet';
+----
+read_parquet   table   Utf8/List<Utf8>   t   SELECT * FROM read_parquet('./my_data.parquet')   Returns a table by scanning the given Parquet file(s).
+read_parquet   table   Utf8/List<Utf8>   t   SELECT * FROM read_parquet('./my_data.parquet')   Returns a table by scanning the given Parquet file(s).
```
universalmind303, line 58: it's because we are adding 2 entries. one for the alias, and one for the actual.

Issue #2368, opened 2024-01-08
Fix duplicate function entries in the catalog


scsmithr:
## Description

Currently we're inserting multiple functions with the same name due to the entry generation not accounting for our namespace rules here: https://github.com/GlareDB/glaredb/blob/608cc7174a0d2fe8f70cca3a30b51bd616231ac3/crates/sqlbuiltins/src/functions/mod.rs#L194-L222

Closing this issue requires this test to pass: https://github.com/GlareDB/glaredb/blob/608cc7174a0d2fe8f70cca3a30b51bd616231ac3/crates/metastore/src/database.rs#L1409-L1435

PR #2370, opened 2024-01-08
chore: Add alias map for functions

scsmithr:
Closes https://github.com/GlareDB/glaredb/issues/2368

Futzing around with function aliases. This replaces `HashMap` with `AliasMap` to allow multiple keys to point to the same object. Also adds some tests.

I needed to add a workaround for 'array_to_string' for one of the iterators, see https://github.com/GlareDB/glaredb/issues/2371.

### DIFF
--- a/testdata/sqllogictests/catalog/functions.slt
+++ b/testdata/sqllogictests/catalog/functions.slt
@@ -42,7 +42,6 @@ where function_name = 'sum';
 sum   aggregate   Int8/Int16/Int32/Int64/UInt8/UInt16/UInt32/UInt64/Float32/Float64   t   sum(a)   Returns the sum of a column
 
 # Assert an arbitrary glaredb table function exists.
-# TODO: Why are there two? No clue.
 query TTTTTT
 select
     function_name,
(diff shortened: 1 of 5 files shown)


testdata/sqllogictests/catalog/functions.slt
```diff
@@ -55,4 +54,35 @@ from glare_catalog.functions
...
+    function_name,
+    function_type,
+    parameters,
+    builtin,
+    example,
+    description
+from glare_catalog.functions
+where function_name = 'parquet_scan';
+----
+parquet_scan   table   Utf8/List<Utf8>   t   SELECT * FROM read_parquet('./my_data.parquet')   Returns a table by scanning the given Parquet file(s).
```
universalmind303, line 70: so i didn't have a chance to comment on it in the previous [pr](https://github.com/GlareDB/glaredb/pull/2367), but to me it seems wrong that the function name is 'parquet_scan' and the example uses 'read_parquet'.

universalmind303 (commented):
IDK, I'm not entirely convinced we should deduplicate the entries using the aliases. It seems valid to keep them as separate entries & is what other databases seem to do. 


I do think we need to address the duplicates showing up in the catalog but I don't know if de-duplicating based on the resolved name is the answer.

To me, an entry with a different `function_name` should be a valid separate entry.

testdata/sqllogictests/catalog/functions.slt
scsmithr: I agree with this, but I didn't feel like getting examples exactly correct was the biggest priority since the previous version was incorrect here too (on prod):
Screenshot 2024-01-08 at 1 31 49 PM

scsmithr:
> IDK, I'm not entirely convinced we should deduplicate the entries using the aliases. It seems valid to keep them as separate entries & is what other databases seem to do.
> 
> [...]
> 
> To me, an entry with a different `function_name` should be a valid separate entry.

This is an implementation detail. Separate entries are still exposed through the catalog:

```
> select * from glare_catalog.functions where function_name like '%parquet%';
┌────────┬────────────┬───────────────┬───────────────┬───────────────────┬─────────┬─────────────────────────────────────────────────┬───────────────────────────────────┐
│    oid │ schema_oid │ function_name │ function_type │ parameters        │ builtin │ example                                         │ description                       │
│     ── │         ── │ ──            │ ──            │ ──                │ ──      │ ──                                              │ ──                                │
│ UInt32 │     UInt32 │ Utf8          │ Utf8          │ List<Utf8>        │ Boolean │ Utf8                                            │ Utf8                              │
╞════════╪════════════╪═══════════════╪═══════════════╪═══════════════════╪═════════╪═════════════════════════════════════════════════╪═══════════════════════════════════╡
│  16521 │      16386 │ read_parquet  │ table         │ [Utf8/List<Utf8>] │ true    │ SELECT * FROM read_parquet('./my_data.parquet') │ Returns a table by scanning the … │
│  16520 │      16386 │ parquet_scan  │ table         │ [Utf8/List<Utf8>] │ true    │ SELECT * FROM read_parquet('./my_data.parquet') │ Returns a table by scanning the … │
└────────┴────────────┴───────────────┴───────────────┴───────────────────┴─────────┴─────────────────────────────────────────────────┴───────────────────────────────────┘
> 
```

And both can be called by the user without needing to know which is the primary function name, or the alias. It all behaves the same.

> I do think we need to address the duplicates showing up in the catalog but I don't know if de-duplicating based on the resolved name is the answer.

This isn't deduplicating based on the name, it's deduplicating based on the implementation. Both `read_parquet` and `parquet_scan` point to the same implementation. 

> [...] resolved name [...]

My dream (at some point) is we don't resolve functions here by name, but by arbitrary ids, and keep the name resolution for everything centralized in sqlexec. Names can be ambiguous to work with.

universalmind303 (approved):

PR #2376, opened 2024-01-08
fix: add signature for read_excel

### LINKED ISSUE #2373: Parameters for `read_excel` are empty
```
> select function_name, parameters from glare_catalog.functions WHERE function_name = 'read_excel';
┌───────────────┬────────────┐
│ function_name │ parameters │
│ ──            │ ──         │
│ Utf8          │ List<Utf8> │
╞═══════════════╪════════════╡
│ read_excel    │ []         │
└───────────────┴────────────┘
>
```

universalmind303:
closes #2373 

```sql
> select function_name, parameters from glare_catalog.functions WHERE function_name = 'read_excel';
┌───────────────┬─────────────────────────────────────────────────────────────────────────┐
│ function_name │ parameters                                                              │
│ ──            │ ──                                                                      │
│ Utf8          │ List<Utf8>                                                              │
╞═══════════════╪═════════════════════════════════════════════════════════════════════════╡
│ read_excel    │ [Utf8, Utf8, sheet_name: Utf8, infer_rows: UInt64, has_header: Boolean] │
└───────────────┴─────────────────────────────────────────────────────────────────────────┘
```

### DIFF
--- a/crates/sqlbuiltins/src/functions/table/excel.rs
+++ b/crates/sqlbuiltins/src/functions/table/excel.rs
@@ -1,5 +1,7 @@
 use async_trait::async_trait;
+use datafusion::arrow::datatypes::{DataType, Field, Fields};
 use datafusion::datasource::TableProvider;
+use datafusion::logical_expr::{Signature, TypeSignature, Volatility};
 use datafusion_ext::errors::{ExtensionError, Result};
 use datafusion_ext::functions::{FuncParamValue, TableFuncContextProvider};
 use datasources::common::url::DatasourceUrl;
@@ -8,6 +10,7 @@ use ioutil::resolve_path;
 use protogen::metastore::types::catalog::{FunctionType, RuntimePreference};
 use std::collections::HashMap;
 use std::sync::Arc;
+use tracing::field;
 
 use super::{table_location_and_opts, TableFunc};
 use crate::functions::ConstBuiltinFunction;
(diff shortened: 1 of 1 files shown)


scsmithr (approved):

greyscaled:
In the PR description, why is there two UTF8's in a row?

```
[Utf8, Utf8, sheet_name: Utf8, infer_rows: UInt64, has_header: Boolean]
```

greyscaled (commented):
Quick Q:

In the PR description, why is there two Utf8's in a row?

```
[Utf8, Utf8, sheet_name: Utf8, infer_rows: UInt64, has_header: Boolean]
```

The first seems to be for required path arg, but what's the second?

crates/sqlbuiltins/src/functions/table/excel.rs
```diff
@@ -21,6 +23,22 @@ impl ConstBuiltinFunction for ExcelScan {
...
         "SELECT * FROM read_excel('file:///path/to/file.xlsx', sheet_name => 'Sheet1')";
     const FUNCTION_TYPE: FunctionType = FunctionType::TableReturning;
+    fn signature(&self) -> Option<Signature> {
+        let options: Fields = vec![
+            Field::new("sheet_name", DataType::Utf8, true),
+            Field::new("infer_rows", DataType::UInt64, true),
+            Field::new("has_header", DataType::Boolean, true),
+        ]
+        .into_iter()
+        .collect();
```
tychoish, line 33: From is implemented for `Vec<Field>` and Fields, so I think FromIterator handlers, which might make this simpler?

To be clear, this is fine, but I definitely stopped and read this three times because it *looks* wrong. "why are we making a vector, converting it to an iterator and collecting it back into a vector without doing anything else?" and while it compiles and does the thing, mds

If it's easy enough to change I think that would make it better and less likely to confuse other people in the future.

universalmind303:
> Quick Q:
> 
> In the PR description, why is there two UTF8's in a row?

Datafusion doesn't really give us that good formatting out of the box for the signatures. 

so the first signature is `Utf8` -> `read_excel('path')`, the second is `Utf8, sheet_name: Utf8, infer_rows: UInt64, has_header: Boolean` -> `read_excel('path', options)`

greyscaled:
> > Quick Q:
> > In the PR description, why is there two UTF8's in a row?
> 
> Datafusion doesn't really give us that good formatting out of the box for the signatures.
> 
> so the first signature is `Utf8` -> `read_excel('path')`, the second is `Utf8, sheet_name: Utf8, infer_rows: UInt64, has_header: Boolean` -> `read_excel('path', options)`
> 
> @greyscaled I agree that the formatting choices are not great though.

all good, I appreciate the explanation

Issue #2375, opened 2024-01-08
parameters for read_csv are now wrong


universalmind303:
## Description
https://github.com/GlareDB/glaredb/pull/2365 changed the signature to now allow a delimiter. 

```sql
> select distinc function_name, parameters from glare_catalog.functions WHERE function_name = 'read_csv';
┌───────────────┬───────────────────┐
│ function_name │ parameters        │
│ ──            │ ──                │
│ Utf8          │ List<Utf8>        │
╞═══════════════╪═══════════════════╡
│ read_csv      │ [Utf8/List<Utf8>] │
└───────────────┴───────────────────┘
```

scsmithr:
Going to add a headers arg for csv, so will add this in while I'm there.

PR #2380, opened 2024-01-08
feat: Allow reading csv files without a header

scsmithr:
```
> select * from read_csv('./testdata/csv/headerless.csv', header => 'false');
┌──────────┬──────────┬──────────┬──────────┐
│ column_1 │ column_2 │ column_3 │ column_4 │
│       ── │ ──       │ ──       │       ── │
│    Int64 │ Utf8     │ Utf8     │  Float64 │
╞══════════╪══════════╪══════════╪══════════╡
│        1 │ hello    │ world    │  3.90000 │
│        2 │ HELLO    │ WORLD    │  4.90000 │
└──────────┴──────────┴──────────┴──────────┘
```

Also does things with the signature. Closes https://github.com/GlareDB/glaredb/issues/2375


testdata/sqllogictests/functions/read_csv.slt
```diff
@@ -74,3 +80,26 @@ select * from read_csv('./testdata/csv/delimiter.csv', delimiter => ';');
...
+# Headers
+
+query ITTR rowsort
+select * from read_csv('./testdata/csv/headerless.csv', header => false);
+----
+1   hello   world   3.9
+2   HELLO   WORLD   4.9
+
+query ITTR rowsort
+select * from read_csv('./testdata/csv/headerless.csv', header => 'false');
```
universalmind303, line 93: i don't think we should do implicit conversions. I'd expect any string value to error out here.

crates/sqlbuiltins/src/functions/table/object_store.rs
```diff
@@ -56,6 +58,13 @@ pub struct CsvOptionReader;
 impl OptionReader for CsvOptionReader {
     type Format = CsvFormat;
 
+    const OPTIONS: &'static [(&'static str, DataType)] = &[
+        // Specify delimiter between fields. Default: ','
+        ("delimiter", DataType::Utf8),
+        // Try to read a header. Default: true
+        ("header", DataType::Boolean),
```
universalmind303, line 65: We should probably be consistent here. In excel, we've been calling this `has_header`. I think `has_header` is a bit more explicit. I don`t have strong feelings either way as long as they are consistent.

crates/sqlbuiltins/src/functions/table/object_store.rs
```diff
@@ -174,11 +193,20 @@ impl<Opts: OptionReader> BuiltinFunction for ObjScanTableFunc<Opts> {
...
             vec![
-                DataType::Utf8,
-                DataType::List(Arc::new(Field::new("item", DataType::Utf8, false))),
+                TypeSignature::Exact(vec![DataType::Utf8]),
+                TypeSignature::Exact(vec![DataType::Utf8, DataType::Struct(opts.clone())]),
+                TypeSignature::Exact(vec![DataType::new_list(DataType::Utf8, false)]),
+                TypeSignature::Exact(vec![
+                    DataType::new_list(DataType::Utf8, false),
+                    DataType::Struct(opts),
+                ]),
```
universalmind303, line 209:
```suggestion
                // read_csv('path')
                TypeSignature::Exact(vec![DataType::Utf8]),
                // read_csv('path', ...options)
                TypeSignature::Exact(vec![DataType::Utf8, DataType::Struct(opts.clone())]),
                // read_csv(['path1', 'path2'])
                TypeSignature::Exact(vec![DataType::new_list(DataType::Utf8, false)]),
                // read_csv(['path1', 'path2'], options)
                TypeSignature::Exact(vec![
                    DataType::new_list(DataType::Utf8, false),
                    DataType::Struct(opts),
                ]),
```

testdata/sqllogictests/functions/read_csv.slt
scsmithr: I disagree with this, I think it's reasonable to to expect 'false' gets automatically cast to boolean here. And this auto cast feels consistent with other sql systems.

Postgres will convert 'false' to a boolean if the context expects a boolean: https://www.postgresql.org/docs/current/datatype-boolean.html

crates/sqlbuiltins/src/functions/table/object_store.rs
scsmithr: Will update

testdata/sqllogictests/functions/read_csv.slt
universalmind303: fair enough.

universalmind303 (approved):

Issue #2390, opened 2024-01-10
stackoverflow when using tab completion in psql 


universalmind303:
## Description

```sh
# terminal 1
> glaredb server
12:36:27.542  INFO metastore::local: Starting in-memory metastore
12:36:27.544  INFO metastore::srv: Creating new Metastore service process_id=c5c943ea-de19-46f6-9e9d-39e578311e75
12:36:27.549  INFO glaredb::server: Starting GlareDB 0.7.1

Connect via Postgres protocol: postgresql://0.0.0.0:6543
```

```sh
# terminal 2
psql "host=localhost user=glaredb dbname=glaredb port=6543"
psql (15.3 (Homebrew), server 15.1)
Type "help" for help.

glaredb=>  select * from <KC_TAB>
```

```sh
# terminal 1
thread 'server-thread-9' has overflowed its stack
fatal runtime error: stack overflow

universalmind303:
It may be unrelated, but I'm seeing the same error when trying to e2e test the `pg_type` table. The server gives the same response when connecting via the `adbc_postgres` driver.

scsmithr:
DF issues related to stack blowing up: https://github.com/apache/arrow-datafusion/issues?q=is%3Aissue+stack+overflow+is%3Aopen

Issue #747, opened 2023-03-13
Catalog builtin for pg_type


RustomMS:
## Summary

Enable Metabase by providing builtin for [`pg_type`](https://www.postgresql.org/docs/current/catalog-pg-type.html)

Issue #1450, opened 2023-08-01
(DBeaver): Failed connection: Error during planning: Unable to fetch table provider for 'pg_catalog.pg_type'


greyscaled:
## Context

If I try to connect to my GlareDB cloud-hosted db with DBeaver, it gives two options:

- Host
- URL

1. If I try Host with org_name as a subdomain, I get:

> FATAL: cloud auth: missing org ID: pass it as an option, or subdomain in proxy or in database as '<org>/<db>'
FATAL: cloud auth: missing org ID: pass it as an option, or subdomain in proxy or in database as '<org>/<db>'

2. If I try Host with org_id as a subdomain, I get:

To me, this indidcates a "successful" connection, minus support for `pg_type`

> ERROR: Error during planning: Unable to fetch table provider for 'pg_catalog.pg_type': Failed to dispatch to table: Missing catalog entry; schema: pg_catalog, name: pg_type
> ERROR: Error during planning: Unable to fetch table provider for 'pg_catalog.pg_type': Failed to dispatch to table: Missing catalog entry; schema: pg_catalog, name: pg_type
>   ERROR: Invalid number of values provided. Expected: 0, got: 1

I get this same response if I set `--options` in their UI

3. If I try `<org_name>/<db_name>`, I get:

(under-the-hood,  it uses JDBC, and it invalidates URLs that we understand):

> Unable to parse URL jdbc:postgresql://proxy.qa.glaredb.com:6543/grey_new/grey_new_db
Unable to parse URL jdbc:postgresql://proxy.qa.glaredb.com:6543/grey_new/grey_new_db

I get a very similar error if I try `<org_id>/<db_name>`, I get:

## Expected

I'm able to connect to GlareDB with DBeaver. It's worth noting JDBC doesn't play nicely with some of our formats, but there are always workarounds. However, given how frequently I've run into this, we may want to consider how we present connection strings down the line.

## Actual

I cannot successfully hookup GlareDB isntance with DBeaver

greyscaled:
Very similar class of bug to #1445

YuriyGavrilov:
+1

PR #2392, opened 2024-01-10
implement `pg_catalog.pg_type` table

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/747
closes https://github.com/GlareDB/glaredb/issues/2390

Note, I don't think this quite yet closes the other [connector issues](https://github.com/GlareDB/glaredb/issues?q=is%3Aissue+is%3Aopen+pg_type). Many of them rely on transactions to work properly (such as https://github.com/GlareDB/glaredb/issues/1450)


crates/glaredb/src/commands.rs
```diff
@@ -281,7 +281,22 @@ impl RunCommand for MetastoreArgs {
...
+    //
+    // Note that Sean observed the stack size only reaching ~300KB when
+    // running in release mode, and so we don't need to bump this
+    // everywhere. However there's definitely improvements to stack
+    // usage that we can make.
+    // see <https://github.com/GlareDB/glaredb/issues/2390>
+    #[cfg(not(release))]
+    builder.thread_stack_size(4 * 1024 * 1024);
+
+    let runtime = builder
```
universalmind303, line 299: the sql queries from some of the connectors was causing stack overflows in dev due to referencing many views.

tychoish (commented):
is there any kind of test we should write? 

is the nested view stack problem the kind of thing that we should raise with datafusion? It seems like we don't really have enough control or tools to push the stack size back down?

universalmind303:
> is there any kind of test we should write?
> 
> is the nested view stack problem the kind of thing that we should raise with datafusion? It seems like we don't really have enough control or tools to push the stack size back down?

release already uses different optimizations & removes all debug symbols, so this isn't really an issue outside of development. Since this is something we've already done with our SLT for the exact same reason, it feels reasonable to introduce it here as well. 

its really only an issue in develop, and it is a somewhat known one. There are a few issues already [open](https://github.com/apache/arrow-datafusion/issues?q=is%3Aissue+stack+overflow+is%3Aopen) in datafusion wrt SO during planning.

scsmithr (commented):
This looks good. Let's add an SLT to assert we can query this without erroring.

vrongmeal (approved):

PR #2400, opened 2024-01-10
ci: downgrade upload-artifact v4->v3

greyscaled:


scsmithr (commented):
Need to downgrade download-artifact too

greyscaled:
> Need to downgrade download-artifact too

I appreciate that, I wouldn't have realized they were intertwined right away. 👍🏻

scsmithr (approved):

Issue #2378, opened 2024-01-08
Chore: decision on slt as part of glaredb, or separate binary


universalmind303:
## Description
https://github.com/GlareDB/glaredb/pull/2363

I think the only thing we didn't decide on in this PR was whether or not to include `slt` in the binary, or keep it separate. 

My vote is for moving it in and feature flagging it to either a new feature, or `!release`

tychoish:
I'm a vote for:
- _yes_ moving it in
- **no** feature flag
- _yes_ warning log messages for release builds.
- _yes_ hide slt from help.

scsmithr:
I don't have a strong opinion here.

If we don't feature flag, I'd just want to make sure that it doesn't significantly increase binary size, and that a user won't accidentally stumble upon on it during typical usage (help).

vrongmeal:
I think a `!release` here makes more sense.

vrongmeal:
My argument is more towards: Are there any advantages of including a test runner in the release build? And an instant answer is no for me.

If a user faces any issue, they are likely to provide an SQL script to reproduce the issue, SLTs are only useful to us.

For me, they would only result in increasing the binary size. The only thing I can think of in favour of including it is that we keep the release and debug binary as close as possible (which IDK if in this case is a very strong argument).

Hiding it/ Not including it is quite similar (except for the binary size).

universalmind303:
I agree with @vrongmeal's sentiment here. It's a pretty narrowly scoped feature, so I'm not so concerned about configuration drift between dev/prod. It seems like if we are already going to hide it for prod, why not just remove it entirely?

tychoish:
> If a user faces any issue, they are likely to provide an SQL script to reproduce the issue, SLTs are only useful to us.

I'm thinking less of "users writing slts" and more of the case where we write an slt to validate the behavior that we expect on someone else's system, we can get a clear and controlled response, as well as more coverage than running a sql script would provide us. 

It also seems that, because we don't really test on many platforms, being able to rerun (most?) of our test suite to recertify that we do in fact work as expected on an arbitrary platform, is a good property to have.

>  The only thing I can think of in favour of including it is that we keep the release and debug binary as close as possible (which IDK if in this case is a very strong argument).

To be honest, this is pretty compelling to me, and the main thing that I'm worried about,  distributing a binary that we aren't actually testing, or using locally ourselves, seems like it incurs a pretty non trivial risk (different binary, different code layout, different compilation process that we don't test,) 

[...]

PR #2401, opened 2024-01-10
chore: move slt runner into main binary

tychoish:
Addresses #2378

See discussion on the ticket.


universalmind303 (commented):
I won't be a blocker for this, but I'd still much prefer feature flagging instead of just hiding. While the extra binary size isn't a lot, every little bit slowly adds up. 

Since it is exclusively a development feature, it seems natural to feature flag as such.

tychoish:
> I won't be a blocker for this, but I'd still much prefer feature flagging instead of just hiding. While the extra binary size isn't a lot, every little bit slowly adds up.

by my read this shouldn't increase the binary by more than a few kb 

> Since it is exclusively a development feature, it seems natural to feature flag as such.

let's get it working first :joy: and talk about it on friday.

.github/workflows/ci.yaml
```diff
@@ -227,15 +227,15 @@ jobs:
           if [[ "${{ matrix.protocol }}" == "rpc" ]]; then
             just rpc-tests
           else
-            just slt-bin -v --protocol=${{ matrix.protocol }} 'sqllogictests/*'
+            just slt-bin --protocol=${{ matrix.protocol }} 'sqllogictests/*'
```
universalmind303, line 230: this branch enables debug logging if you rerun the job in debug mode. we should keep this.

crates/glaredb/src/commands.rs
```diff
@@ -33,6 +42,9 @@ pub enum Commands {
     /// Starts an instance of the Metastore.
     #[clap(hide = true)]
     Metastore(MetastoreArgs),
+    /// Runs SQL Logic Tests
+    #[clap(hide = true)]
```
universalmind303, line 46: can we alias this to `slt`

```suggestion
    #[clap(hide = true, alias="slt")]
```

tychoish:
Sorry for the really noisy PR, I've been trying to iron and and do a bit of cleanup with the dependencies just in awareness of the binary size. My conclusion generally:
- it's probably the flightsql client (this is the main thing that's in slt that wasn't in glaredb before. 
- we might get more out of doing some flightsql RPC tests using a non-rust client (because testing it with go or python might reveal something tonic-specific that we've done, that the existing tests wouldn't catch.)
- the flightsql client will become moot when we start using flightsql as our internal IPC for dist exec anyway, so it's not super important now.
- we can always flag this out of the build later if we want, but I'd like to start with giving it a try. 
- I think the most compelling argument for keeping this in main is that it means we can build glaredb for aarch64 and then run the SLTs with the binary from the package on the actual hardware, rather than just trusting that all of the cross compilation works.

Do folks agree?

PR #2413, opened 2024-01-11
Fix : Adding keywords for PRQL for higlighting

### LINKED ISSUE #2114: syntax Highlighting doesnt work properly for prql dialect
## Description
When the dialect is set to prql, it is still using the sql highlighting. 

image

Instead it should use different highlighting logic specific to the prql language.

hemanth94:
Fixes #2114 
A dedicated highlighting logic for PRQL queries is desired in future.


universalmind303 (approved):

tychoish:
My only concern about this is that we'll now highlight PRQL keywords in SQL statements where they are not special. 

We should definitely keep a ticket open (perhaps not this one,) to resolve the above issue by implementing the dedicated formatter.

Issue #1870, opened 2023-10-02
drop tables doesn't delete native table files.


universalmind303:
## Description

files are not dropped when tables are dropped.
likely related to #1855

```
> \open ./db
> create table t1 (a int, b int);
try_insert_table_namespace
Table created
> insert into t1 values (1, 2);
Inserted 1 row
> select * from t1;
┌───────┬───────┐
│     a │     b │
│    ── │    ── │
│ Int32 │ Int32 │
╞═══════╪═══════╡
│     1 │     2 │
└───────┴───────┘
> drop table t1;
Table(s) dropped
> Ctrl+D
glaredb  ❯ ls db/databases/00000000-0000-0000-0000-000000000000/tables/                                                                                                                       10/02/2023 04:14:11 PM
╭───┬────────────────────────────────────────────────────────────────┬──────┬───────┬──────────────╮
│ # │                              name                              │ type │ size  │   modified   │
├───┼────────────────────────────────────────────────────────────────┼──────┼───────┼──────────────┤
│ 0 │ db/databases/00000000-0000-0000-0000-000000000000/tables/20000 │ dir  │ 128 B │ a minute ago │
╰───┴────────────────────────────────────────────────────────────────┴──────┴───────┴──────────────╯
```

scsmithr:
Is there anything inside that directory? We have https://github.com/GlareDB/glaredb/issues/1411 open for not deleting the directory itself though (only for local, s3/gcs work as expected I believe).

universalmind303:
it doesn't delete any of the files 

<img width="894" alt="image" src="https://github.com/GlareDB/glaredb/assets/21327470/0aab73f9-6bd8-4b84-a5bb-5ff886cf808f">

PR #2415, opened 2024-01-11
fix: actually drop delta tables

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/1870

I wanted to open this up to make sure that others are aligned on the direction I took with this before getting too far. This works as is, but could use a bit of cleanup in areas. 

It uses our scheduler to spawn a system_exec to remove the delta tables. Currently it just logs the output, but we could put them in a table or something else.. Logging seemed to be fine for now.. I didn't want to over-engineer and throw away a bunch of work if this isn't the direction we want to take.


universalmind303:
so this is technically working fine, but it isn't working inside of a test, likely because of the known issues with the scheduler. The task gets scheduled, but never picked up. Even when putting a long sleep in the test, the task never gets picked up. I'll keep this in a WIP until we can get our scheduler working properly

tychoish:
> so this is technically working fine, but it isn't working inside of a test, likely because of the known issues with the scheduler. The task gets scheduled, but never picked up. Even when putting a long sleep in the test, the task never gets picked up. So since we can't write an integration test for this, I'll keep this in a WIP until we can get our scheduler working properly

When you say technically working fine, you mean that you can observe it working correctly in manual testing but that because the distributed execution scheduler is not working in the testing environment that your tests don't pass? 

I think keeping it draft until distexec gets more ironed out seems fine. It also doesn't seem necessary to use the distexec scheduler for this because:

- why not just have the remove operation run in the calling thread? having the calling operation return before the delete operation returns seems like it could lead to some confusions (and failed deletes) 
- why not spawn a task for doing this directly? The scheduler mostly exists to provide fairness and limitations when we hav a lot of very small tasks (as we would in a dist-exec situation,) but that doesn't seem like a huge problem here. 
- using the distexec scheduler for the "non-dist-exec" code paths, seems potentially confusing, and might make it harder operationally to turn things off, or understand things.

Having said that, all of the potential bugs or issues from using the distexec scheduler before _everything_ becomes distexec is just that we'd regress on this issue, which is already happening today, so I wouldn't worry about it too hard, though obviously this is all moot until the scheduler works better.

universalmind303:
> When you say technically working fine, you mean that you can observe it working correctly in manual testing but that because the distributed execution scheduler is not working in the testing environment that your tests don't pass?

yeah, it functionally works, but not in the testing suite.

crates/protogen/src/sqlexec/physical_plan.rs
```diff
@@ -252,6 +252,8 @@ pub struct DropTablesExec {
     pub tbl_references: Vec<FullObjectReference>,
     #[prost(bool, tag = "3")]
     pub if_exists: bool,
+    #[prost(message, repeated, tag = "4")]
+    pub tbl_entries: Vec<TableEntry>,
```
scsmithr, line 256: If we have table entries, do we need table references? Can we just use the info provided in entries?

universalmind303: I originally wanted to go that route, but we don't have access to an entry resolver inside the execution context, only during planning.

crates/sqlexec/src/planner/physical_plan/drop_tables.rs
```diff
@@ -101,16 +105,20 @@ async fn drop_tables(
...
-    // // on the session until transaction commit.
-    // self.background_jobs.add_many(jobs)?;
+    // only after the catalog is updated, we can delete the delta tables
+    // TODO: this should be done in the scheduler.
+    let sys_exec =
+        SystemOperationExec::new(DeleteDeltaTablesOperation::new(plan.tbl_entries.clone()).into());
+    let _ = sys_exec
+        .execute(0, context.clone())?
+        .collect::<Vec<_>>()
+        .await;
```
tychoish, line 121: what are the items in the iterator that we're throwing away? 

The first time I read this it seemed like we were throwing away an error, and now I'm unsure.

tychoish (commented):
seems reasonable, one question.

crates/sqlexec/src/planner/physical_plan/drop_tables.rs
universalmind303: We just discard the stream of recordbatches. The error is still propagated if the execution fails, but we don't care about the results.

PR #2432, opened 2024-01-16
chore: Remove openssh from slt crate cargo.toml

scsmithr:
Was causing build failures on windows. Didn't seem to be used anyways.

Note that we conditionally use openssh in the datasources crate to get around this:

```
[target.'cfg(any(target_os = "linux", target_os = "macos"))'.dependencies]
openssh = "0.10.2"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7187,7 +7187,6 @@ dependencies = [
  "logutil",
  "metastore",
  "object_store",
- "openssh",
  "pgrepr",
  "pgsrv",
  "regex",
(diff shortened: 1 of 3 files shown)


greyscaled (approved):

universalmind303 (approved):

tychoish (approved):
what are we using this for anyway if it's only on mac?

scsmithr:
> what are we using this for anyway if it's only on mac?

It's mac and linux. It's for ssh tunnels for postgres and mysql. The windows implementation just returns an error saying unsupported on windows.

PR #2433, opened 2024-01-16
chore: Post release build results to slack

scsmithr:
Closes https://github.com/GlareDB/glaredb/issues/2431

### DIFF
--- a/.github/workflows/nodejs-release.yml
+++ b/.github/workflows/nodejs-release.yml
@@ -8,6 +8,8 @@ on:
     # Run every Monday at 12:00
     - cron: 0 12 * * 1
   push:
+    branches:
+      - "sean/slack"
     tags:
       - "*"
   workflow_dispatch:
@@ -29,6 +31,8 @@ env:
   APP_NAME: "glaredb"
   MACOSX_DEPLOYMENT_TARGET: "10.13"
   PROTOC: "${{ github.workspace }}/deps/protoc/bin/protoc"
+  SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+  SLACK_WEBHOOK_TYPE: INCOMING_WEBHOOK
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/nodejs-release.yml
```diff
@@ -8,6 +8,8 @@ on:
     # Run every Monday at 12:00
     - cron: 0 12 * * 1
   push:
+    branches:
+      - "sean/slack"
```
greyscaled, line 12: should this be removed before merge?

scsmithr: yes

greyscaled (commented):
This is great

tychoish (approved):
is there an easy way to have there be one post per release/commit, with comments/updates about each specific release/workflow

greyscaled (approved):

scsmithr:
> is there an easy way to have there be one post per release/commit, with comments/updates about each specific release/workflow

I'm sure there's a way. The slack action readme says you can do things like `update-ts: ${{ steps.slack.outputs.ts }}` but it seems that needs a reference to a previous step and I don't know of an easy of getting this "ts" across workflows.

I just went with the quickest/easiest/straightforwardest way to get us feedback on the builds. Iterating on this to have a single message sounded kinda painful. What if one of the node js jobs passed, then a python job failed? Would we be able to preserve that context in the message?

tychoish:
> I'm sure there's a way. The slack action readme says you can do things like `update-ts: ${{ steps.slack.outputs.ts }}` but it seems that needs a reference to a previous step and I don't know of an easy of getting this "ts" across workflows.
> 
> I just went with the quickest/easiest/straightforwardest way to get us feedback on the builds. Iterating on this to have a single message sounded kinda painful. What if one of the node js jobs passed, then a python job failed? Would we be able to preserve that context in the message?

Legit! agree that it's good to merge. I think combining into one release workflow with three jobs wouldn't be too bad, and then it'd be a single workflow.

Issue #2440, opened 2024-01-17
ci(node): just rate limit


greyscaled:
run: https://github.com/GlareDB/glaredb/actions/runs/7560954189

Error:
```
Status: Downloaded newer image for ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian
+ curl --proto =https --tlsv1.2 -sSf https://just.systems/install.sh
+ bash -s -- --to .
curl: (22) The requested URL returned error: 403
```

Cause:
https://github.com/GlareDB/glaredb/blob/690b74f287e45108d9ca28c91608359f07441d25/.github/workflows/nodejs-release.yml#L54-L55

Referring to https://just.systems/man/en/chapter_5.html it seems like we need to add `--tag`:

> Note that install.sh may fail on GitHub actions, or in other environments where many machines share IP addresses. install.sh calls GitHub APIs in order to determine the latest version of just to install, and those API calls are rate-limited on a per-IP basis. To make install.sh more reliable in such circumstances, pass a specific tag to install with --tag.

scsmithr:
Fixed in https://github.com/GlareDB/glaredb/pull/2442

greyscaled:
Still seeing this

tychoish:
@greyscaled link to a recent failure?

greyscaled:
> @greyscaled link to a recent failure?

https://github.com/GlareDB/glaredb/actions/runs/7610842049/job/20725034938

PR #2442, opened 2024-01-18
chore(ci): pin just version

tychoish:
Addresses #2440


greyscaled:
@tychoish still seeing #2440 from this spot:

https://github.com/GlareDB/glaredb/blob/abd3f5546d79fe1cbe9a878f62f128eb238760f0/.github/workflows/nodejs-release.yml#L52-L59

PR #2443, opened 2024-01-18
feat: Notices on transactions

scsmithr:
Emit notices on transaction commands warning about our incomplete transaction support.

psql:
```
sean@Seans-Air glaredb % psql postgres://localhost:6543
psql (15.4, server 15.1)
Type "help" for help.

sean=> begin;
WARNING:  GlareDB does not support proper transactional semantics. Do not rely on transactions for correctness. Transactions are stubbed out to enable compatability with existing Postgres tools.
BEGIN
sean=> 
```

CLI (colorful):
Screenshot 2024-01-18 at 2 26 37 PM

Some pg protocol tests added to assert the notices.

### DIFF
--- a/crates/glaredb/src/local.rs
+++ b/crates/glaredb/src/local.rs
@@ -169,7 +169,7 @@ impl LocalSession {
                         // non-interactive fashion which and having notice
                         // messages interspersed with the output would be
                         // annoying.
-                        for notice in self.sess.drain_notices() {
+                        for notice in self.sess.take_notices() {
                             eprintln!(
                                 "{}: {}",
                                 match notice.severity {
--- a/crates/glaredb/src/local.rs
+++ b/crates/glaredb/src/local.rs
@@ -14,6 +14,7 @@ use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::physical_plan::SendableRecordBatchStream;
 use futures::StreamExt;
 use pgrepr::format::Format;
+use pgrepr::notice::NoticeSeverity;
 use reedline::{FileBackedHistory, Reedline, Signal};
 use std::collections::HashMap;
 
@@ -160,6 +161,25 @@ impl LocalSession {
                             Ok(_) => {}
                             Err(e) => println!("Error: {e}"),
                         };
+
+                        // Print out notices as needed.
+                        //
+                        // Note this isn't being called in the above `execute`
+                        // function since that can be called in a
+                        // non-interactive fashion which and having notice
+                        // messages interspersed with the output would be
+                        // annoying.
+                        for notice in self.sess.drain_notices() {
+                            println!(
+                                "{}: {}",
+                                match notice.severity {
+                                    s @ (NoticeSeverity::Warning | NoticeSeverity::Error) =>
+                                        s.to_string().red(),
+                                    other => other.to_string().blue(),
+                                },
+                                notice.message
+                            );
+                        }
                     }
                 },
                 Ok(Signal::CtrlD) => break,
--- a/crates/glaredb/src/local.rs
+++ b/crates/glaredb/src/local.rs
@@ -170,7 +170,7 @@ impl LocalSession {
                         // messages interspersed with the output would be
                         // annoying.
                         for notice in self.sess.drain_notices() {
-                            println!(
+                            eprintln!(
                                 "{}: {}",
                                 match notice.severity {
                                     s @ (NoticeSeverity::Warning | NoticeSeverity::Error) =>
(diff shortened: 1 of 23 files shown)


crates/glaredb/src/local.rs
```diff
@@ -160,6 +161,25 @@ impl LocalSession {
...
+
+                        // Print out notices as needed.
+                        //
+                        // Note this isn't being called in the above `execute`
+                        // function since that can be called in a
+                        // non-interactive fashion which and having notice
+                        // messages interspersed with the output would be
+                        // annoying.
+                        for notice in self.sess.drain_notices() {
+                            println!(
```
tychoish, line 173: do we want to write to stderr?

crates/pgrepr/src/error.rs
```diff
@@ -32,6 +32,9 @@ pub enum PgReprError {
 
     #[error("Internal error: {0}")]
     InternalError(String),
+
+    #[error("{0}")]
+    String(String),
```
tychoish, line 37: I dislike strongly typed errors that end up just being strings: but I don't have a better suggestion

crates/sqlexec/src/context/local.rs
```diff
@@ -320,6 +324,28 @@ impl LocalSessionContext {
...
 
+    pub(crate) fn push_notice(&mut self, notice: Notice) {
+        self.notices.push(notice)
+    }
+
+    /// Drain all notices from the session.
+    ///
+    /// This will take into account the 'client_min_messages' session var to
+    /// filter out notices that shouldn't be sent to the client.
+    pub(crate) fn drain_notices(&mut self) -> Vec<Notice> {
```
tychoish, line 335: I would have expected this to be `Optional<Vec<Notice>>`. but it sort of seems like it shouldn't even be that, maybe just a bool, or a count of how many there were? if you've drained them,

crates/sqlexec/src/context/local.rs
```diff
@@ -320,6 +324,28 @@ impl LocalSessionContext {
...
+        // Behind an if since this is called every query, and session vars is behind  a lock :)
+        // Let's get rid of the lock.
```
tychoish, line 337: does this mean that we're constructing an empty vector for every iteration?

tychoish (approved):

crates/glaredb/src/local.rs
scsmithr: We could. Since this is in the interactive portion, I don't know that it matters. But we could also move this to the `execute` func and eprintln there.

crates/pgrepr/src/error.rs
scsmithr: I agree, but I've also found creating specific variants to cause friction when developing. That combined with reusing certain variants sometimes makes debugging trickier than needed. String isn't elegant, but it's practical for constructing debuggable errors.

crates/sqlexec/src/context/local.rs
scsmithr: > but it sort of seems like it shouldn't even be that, maybe just a bool, or a count of how many there were? if you've drained them,

I don't understand what this means. `drain_notices` takes the notices from the session so that they can be sent somewhere else higher up (pg protocol or cli). What would bool here do?

Also I picked `Vec<Notice>` since it makes it easier to do `for notice in sess.drain_notices()` elsewhere.

crates/sqlexec/src/context/local.rs
scsmithr: Yes, but empty vecs don't allocate.

crates/sqlexec/src/context/local.rs
tychoish: I sort of assumed that "drain noticed" did more than just copy them by looking at it so the bool/usize options were really more "just report that you're done" rather than "copy them out".

I think this is fine, and is mostly that "drain" lead me to assume the wrong thing. I'm still a touch worried about constructing an empty vector more often than we need, but otherwise this is fine.

crates/sqlexec/src/context/local.rs
tychoish: wat. ok.

crates/glaredb/src/local.rs
tychoish: stderr feels right for this kind of thing, (if only because it makes it easier to write scripts.) but this doesn't need to block.

crates/sqlexec/src/context/local.rs
scsmithr: I'll rename this to `take_notices` to make it clearer.

Originally had this return an iterator to follow Vec's drain method (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain) but changed it since that if makes it a bit harder to return an iterator.

crates/glaredb/src/local.rs
scsmithr: updated

PR #2447, opened 2024-01-19
chore: add nightly rustfmt rules

tychoish:
Per conversation in slack


crates/arrow_util/src/pretty.rs
```diff
@@ -293,8 +295,8 @@ pub fn fmt_dtype(dtype: &DataType) -> String {
             .map(|f| format!("{}: {}", f.name(), fmt_dtype(f.data_type())))
             .collect::<Vec<_>>()
             .join(", "),
-        DataType::Decimal128(_, _) => "Decimal128".to_string(),
-        DataType::Decimal256(_, _) => "Decimal256".to_string(),
+        DataType::Decimal128(..) => "Decimal128".to_string(),
```
scsmithr, line 298: I prefer the other way here.

crates/datafusion_ext/src/metrics.rs
```diff
@@ -161,22 +174,22 @@ pub trait DataSourceMetricsOptsType: 'static + Debug + Clone + Sync + Send {
 #[derive(Debug, Clone)]
 pub struct ReadWriteDataSourceMetricsOptsType;
 impl DataSourceMetricsOptsType for ReadWriteDataSourceMetricsOptsType {
-    const OPTS: DataSourceMetricsOpts = DataSourceMetricsOpts::read_write();
     const DISPLAY_NAME_PREFIX: &'static str = "ReadWrite";
+    const OPTS: DataSourceMetricsOpts = DataSourceMetricsOpts::read_write();
```
scsmithr, line 178: If this is being sorted alphabetically by some rule, I'd prefer if we didn't since there's definitely cases where these are ordered according to what the developer thinks is important.

crates/metastore/src/database.rs
```diff
@@ -682,7 +706,7 @@ impl State {
             Mutation::CreateTunnel(create_tunnel) => {
                 validate_object_name(&create_tunnel.name)?;
                 match self.tunnel_names.get(&create_tunnel.name) {
-                    Some(_) if create_tunnel.if_not_exists => return Ok(()), // Already exists, nothing to do.
+                    Some(_) if create_tunnel.if_not_exists => return Ok(()), /* Already exists, nothing to do. */
```
scsmithr, line 709: I prefer the other way.

tychoish:
[@scsmithr condensed on the current set of options over slack]

crates/datafusion_ext/src/planner/expr/mod.rs
```diff
@@ -26,23 +26,41 @@ mod substring;
...
+    BinaryExpr,
+    BuiltinScalarFunction,
+    Cast,
+    Expr,
+    ExprSchemable,
+    GetFieldAccess,
+    GetIndexedField,
+    Like,
+    Operator,
+    TryCast,
```
universalmind303, line 49: Not a big deal as the imports are easily folded, but I prefer having blocks instead of splitting every item into a newline

crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -36,50 +42,104 @@ use object_store::aws::AmazonS3ConfigKey;
...
+    TableOptionsMysql,
+    TableOptionsObjectStore,
+    TableOptionsPostgres,
+    TableOptionsS3,
+    TableOptionsSnowflake,
+    TableOptionsSqlServer,
+    TunnelOptions,
+    TunnelOptionsDebug,
+    TunnelOptionsInternal,
+    TunnelOptionsSsh,
```
universalmind303, line 98: same comment about using blocks, but much more noticeable here

crates/sqlbuiltins/src/functions/scalars/postgres.rs
```diff
@@ -1,10 +1,10 @@
 use datafusion::logical_expr::expr::ScalarFunction;
 use pgrepr::compatible::server_version_with_build_info;
 
+use super::df_scalars::array_to_string;
+use super::*;
```
universalmind303, line 5: I thought we wanted to forbid usage of `super::*`

crates/sqlexec/src/planner/logical_plan/mod.rs
```diff
@@ -57,6 +42,17 @@ pub use create_table::*;
 pub use create_temp_table::*;
 pub use create_tunnel::*;
 pub use create_view::*;
+use datafusion::arrow::datatypes::{DataType, Schema as ArrowSchema};
```
universalmind303, line 45: Idk if rustfmt can do it, but I think having `pub use` seperate from `use` would be preferred. IMO, it's easier to grok what's being exported if all of those are on top. Same applies for `pub mod` vs `mod`

crates/datafusion_ext/src/planner/expr/mod.rs
tychoish: yeah, I agree that there are cases where it's a bit extreme, but in most cases I think the expanded form ends up being more readable, and easier to grep.

crates/sqlexec/src/planner/logical_plan/mod.rs
tychoish: I agree with you, and also rustfmt doesn't support it. 😦

crates/sqlbuiltins/src/functions/scalars/postgres.rs
tychoish: we do, but rustfmt doesn't support that. 

I will see if there's a clippy rule or something that we can use.

tychoish: https://rust-lang.github.io/rust-clippy/rust-1.56.0/index.html <- looks like there's something. will do that separately.

tychoish:
Holding back on this pr till the DF upgrade lands.

scsmithr (approved):

PR #2448, opened 2024-01-19
feat: pg_catalog.version(), standard_conforming_strings session var

scsmithr:
Adds the `pg_catalog.version()` function (which provides more info that the `server_version` session var). The format provides compatibility with sqlalchemy/great expectations.

Also adds the `standard_conforming_strings` session var since great expectations was looking for this too.

I included the python script I was running against to get great expectations working, and it seems to connect without issue. But I haven't tried anything more than that.

cc @talagluck


crates/sqlbuiltins/src/functions/scalars/postgres.rs
```diff
@@ -475,3 +476,36 @@ impl BuiltinScalarUDF for PgArrayToString {
...
+        Some(Signature::exact(vec![], Volatility::Stable))
+    }
+}
+
+impl BuiltinScalarUDF for PgVersion {
+    fn as_expr(&self, _: Vec<Expr>) -> Expr {
+        Expr::Literal(ScalarValue::Utf8(Some(
+            server_version_with_build_info().to_string(),
+        )))
+    }
```
scsmithr, line 506: @universalmind303 I was originally trying to use `session_var("server_version")` here, but was getting a "Unknown variable: server_version" error.

Using server_version here would be incorrect anyways, but where would I look to enable using it with `session_var`?

tychoish (commented):
I would not put the great expectations dependency or example in the python bindings (it's all via the pg protocol and so is orthogonal. otherwise seems legit.

scsmithr:
> I would not put the great expectations dependency or example in the python bindings (it's all via the pg protocol and so is orthogonal. otherwise seems legit.

I _believe_ this is just needed for the examples, and the actual requirements for the glaredb library is defined in pyproject.toml

vrongmeal (approved):

scsmithr:
> > I would not put the great expectations dependency or example in the python bindings (it's all via the pg protocol and so is orthogonal. otherwise seems legit.
> 
> I _believe_ this is just needed for the examples, and the actual requirements for the glaredb library is defined in pyproject.toml

I just confirmed on pypi, `requirements.txt` doesn't impact any of the metadata for the bindings that get uploaded. Going to merge so this gets into qa soon so we can start testing more tools.

PR #2460, opened 2024-01-19
ci: fix timings.csv

universalmind303:


benchmarks/tpch/justfile
```diff
@@ -44,6 +44,7 @@ run n_runs="1" $SCALE_FACTOR="1": requirements
     just _run_private duckdb {{SCALE_FACTOR}}
     ((n++))
   done
+  exit 0
```
tychoish, line 47: I think we should probably set `errexit` on this script, so that if there's a problem (for whatever reason) that we get some signal for this.

Having said that, given that errexit *isn't set* (unless just is doing something weird) it sort of seems unlikely that this ever exited non-zero because of bash's continue on error defaults.

tychoish (commented):
code seems good, one question about shell scripting.

benchmarks/tpch/justfile
universalmind303: yeah idk why just isn't executing this properly, but it does exit 1 if `exit 0` isnt added. If i take the inner part into it's own bash file & execute it, it exits 0. 🤷‍♂️. 

Seems like `just` is handling it oddly.

PR #2472, opened 2024-01-21
chore: Get most of the rpc tests working

scsmithr:


justfile
```diff
@@ -72,50 +72,10 @@ slt-bin-debug *args:
...
-    'sqllogictests/vars' \
-    'sqllogictests/views' \
-    'sqllogictests/virtual_catalog' \
-    'sqllogictests/xlsx' \
-    'sqllogictests/prql' \
-    'sqllogictests/describe_rpc' \
-    'sqllogictests/allowed_operations'
+  just slt --protocol=rpc "sqllogictests/*" \
+    --exclude "sqllogictests/functions/cache_external_database_tables" \
+    --exclude "sqllogictests/functions/kdl" \
```
tychoish, line 77: I have no idea why this wouldn't work. Seems suspect.

tychoish (approved):
I have read this in a "bearing witness" sort of way. Nothing seems objectionable.

justfile
scsmithr: We're just passing in the default function registry here: https://github.com/glaredb/glaredb/blob/f0da39cb5c2d2453b7aee3e5ea27f14af785b108/crates/rpcsrv/src/session.rs#L58-L62

It's a trait that we can implement, but the `TaskContext` also implements it (which you get from a datafusion session context).

PR #2476, opened 2024-01-22
fix: download protoc without just during the release

tychoish:


.github/workflows/nodejs-release.yml
```diff
@@ -51,8 +51,10 @@ jobs:
             docker: ghcr.io/napi-rs/napi-rs/nodejs-rust:lts-debian
             build: |-
               set -eux &&
-              curl --proto '=https' --tlsv1.2 -sSf https://just.systems/install.sh | bash -s -- --to . &&
-              ./just protoc &&
+              curl -L https://github.com/protocolbuffers/protobuf/releases/download/v23.1/protoc-23.1-linux-x86_64.zip -o protoc.zip
```
greyscaled, line 54: seems like the v23.1 is 'hard-coded' here, thought I'd ask:

- is this something that would need to get bumps periodically,
- (if yes): does anything catch/detect it?

Because we have our nightlies, I'm not overly concerned, but wanted to ask in case there's a way to share this version throughout the repo or pin it in some fashion.

That said, I'm happy to merge, as it fixes our issue

greyscaled (approved):

.github/workflows/nodejs-release.yml
tychoish: it's super unlikely that this will never really need to change, though it certainly could. There's no security implications (it's a build dependency, maybe improvements in protoc would generate tighter code (unlikely) or resolve a bug that we are unlikely to hit, and protobufs are extremely stable as a format. (This is also probably, I think, just used to parse a protobuf and either generate rust code, or generate an AST that rust code then uses to generate rust code, either way, very likely to either work fine or blow up and then we'll fix it.

scsmithr (approved):

PR #2483, opened 2024-01-24
fix: python dependency specification

tychoish:


tychoish:
I expect the environment/optional dependency thing something we should discuss (do we want to force pulling in pandas and polars? Probably dev dependencies are going to lead to some kind of problem.

bindings/python/pyproject.toml
```diff
@@ -12,6 +12,30 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: CPython",
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
+dependencies = [
+  "maturin==1.0.1",
+  "ruff==0.0.275",
+  "mypy==1.3.0",
+  "typos==1.15",
+  "pyarrow",
+]
```
scsmithr, line 21: None of these should be required. I'd expect pyarrow to be in optional dependencies.

talagluck:
I don't think we want to force pulling in polars and I'm not sure about pandas. The more dev dependencies we have the harder this will be to navigate (ran into this a lot at a previous place) and pip's dependency resolver can be kind of heavy, so unless something is strictly necessary for core functionality, it should be optional

universalmind303 (changes requested):
in addition to making things optional dependencies, we should also properly handle when the deps are not installed.

It should error instead of panicking


this is what polars does if you try `df.to_arrow()` without pyarrow installed

```py
Traceback (most recent call last):
  File "/Users/me/Development/glaredb/test_pl.py", line 3, in <module>
    pl.DataFrame({"foo": [1, 2, 3]}).to_arrow()
  File "/Users/me/Development/glaredb/.venv/lib/python3.11/site-packages/polars/dataframe/frame.py", line 1970, in to_arrow
    record_batches = self._df.to_arrow()
                     ^^^^^^^^^^^^^^^^^^^
ModuleNotFoundError: No module named 'pyarrow'
```

we instead just panic.

tychoish:
> in addition to making things optional dependencies, we should also properly handle when the deps are not installed.
> 
> It should error instead of panicking

Totally agree. I think we should probably do that work as part of a seperate ticket just to fix the bug, incase we do a point release early next week. 

Do you, @universalmind303 or @talagluck, want to take this over? g

talagluck:
I'm not sure I know what needs to be done for that, but I can review the PR for it when it comes through.

tychoish:
I spent a couple hours playing with this, and I have something that maybe works:
- installs build time dependencies and test/runtime dependencies into different virtualvens that are scoped to the bindings. 
- installs the extra dependencies (I think we want to have this off by default? 
- the install succeeds but it has a patchelf error message. I was able to run something locally, but I didn't kick the tires very hard. 

I'm not sure how to do the better error messages you're interested in Cory, but that'd definitely be a good thing to add.

bindings/python/justfile
```diff
@@ -1,29 +1,35 @@
 set fallback
-VENV := env_var_or_default("VENV", "../../.venv")
+VENV := env_var_or_default("VENV", ".venv")
 VENV_BIN := VENV / "bin"
 
+VENV_DEV := env_var_or_default("VENV", ".venv-dev")
+VENV_DEV_BIN := VENV_DEV / "bin"
```
universalmind303, line 6: why would we want two `.venv`'s? why not just put all of the dev dependencies into the same virtual env? Seems unnecessary, but maybe i'm missing something.

universalmind303: So initially I had the `bindings/python` with their own `.venv`, but it was moved to the root _(not entirely sure why)_. Why not have a `bindings/python/.venv` and a `testing/.venv`? That is much more common & seems like it'd be the most sensible to me.

tychoish: I've collapsed into one venv for the bindings, but this comment is a bit out of date with the current state of the repo:

- there is a top-level `.venv` environment that installs poetry so that the pytests don't depend on system-poetry and don't have a bootstrapping problem. 
- `tests` has it's own venv with it's own dependencies. 
- `bindings/python` has it's own venv with it's own dependencies.

bindings/python/justfile
```diff
@@ -1,29 +1,26 @@
...
-build *args: requirements
+build *args: venv
   @unset CONDA_PREFIX
-  {{VENV_BIN}}/maturin develop {{args}}
+  {{VENV_BIN}}/maturin develop --pip-path="{{VENV_BIN}}/pip" --extras="pandas,polars,pyarrow" {{args}}
 
-test: requirements
-  {{VENV_BIN}}/pytest -s
+test: venv
+  {{VENV_BIN}}/pytest --rootdir={{invocation_directory()}}/bindings/python
```
universalmind303, line 17: this way it'll work regardless of the directory you are in. 

```suggestion
  {{VENV_BIN}}/pytest --rootdir={{justfile_directory()}}
```

bindings/python/requirements.txt
```diff
@@ -1,14 +1,12 @@
-maturin==1.0.1
-virtualenv==20.23.1
-ruff==0.0.275
+# this requirements.txt is for development deps only
+maturin[patchelf]==1.4.0
+patchelf
```
universalmind303, line 3: we shouldn't need this. The `patchelf` dependency breaks my install on mac.

bindings/python/justfile
```diff
@@ -1,29 +1,26 @@
...
-  {{VENV_BIN}}/pip --quiet install -r requirements.txt
+  {{VENV_BIN}}/python -m pip install --upgrade pip
+  {{VENV_BIN}}/pip install -r requirements.txt
 
 ## Compile and install py-glaredb for development
-build *args: requirements
+build *args: venv
   @unset CONDA_PREFIX
-  {{VENV_BIN}}/maturin develop {{args}}
+  {{VENV_BIN}}/maturin develop --pip-path="{{VENV_BIN}}/pip" --extras="pandas,polars,pyarrow" {{args}}
```
universalmind303, line 14: what is this doing? Everything works fine on my end by just running `maturin develop`

tychoish:
@universalmind303 could you take another look.

bindings/python/pyproject.toml
```diff
@@ -13,6 +13,30 @@ classifiers = [
     "Programming Language :: Python :: Implementation :: PyPy",
 ]
 
+[tool.poetry]
+name = "glaredb"
+version = "0.0.0"
+description = "glaredb bindings"
+authors = ["GlareDB <hello@glaredb.com>"]
+readme = "README.md"
```
universalmind303, line 21: this shouldn't be under `tool.poetry`

If we want to keep this information, it should go under the `[project]` label. 

polars [py-project.toml](https://github.com/pola-rs/polars/blob/main/py-polars/pyproject.toml) 
datafusion-python [py-project.toml](https://github.com/apache/arrow-datafusion-python/blob/main/pyproject.toml)

universalmind303 (changes requested):

bindings/python/requirements.txt
universalmind303: @tychoish the `patchelf` dependency is still breaking things locally. We need to remove this.

tychoish: done

greyscaled (commented):
I just ran this from the `root` dir - should we add a dependency on `venv` for `pytest` in the root Justfile?

```shell
greyb ~/Repos/glaredb [tycho/python-deps] $ just pytest
.venv/bin/poetry -C tests lock --no-update
sh: .venv/bin/poetry: No such file or directory
error: Recipe `pytest` failed on line 131 with exit code 127
```

greyscaled (commented):
After the previous comment, I ran `just venv` and got this:

```shell
greyb ~/Repos/glaredb [tycho/python-deps] $ just venv
python3 -c "import virtualenv" || python3 -m pip --quiet install virtualenv
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'virtualenv'
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try brew install
    xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-brew-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip.
    
    If you wish to install a non-brew packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
error: Recipe `venv` failed on line 124 with exit code 1
```

I recognize that it's prompting me to install `virtualenv` but I'm wondering if this is expected, and if so, can we document this somewhere (if it's not already)

bindings/python/justfile
```diff
@@ -1,29 +1,25 @@
...
 
 @venv:
-  python3 -c "import virtualenv" || python3 -m pip --quiet install virtualenv
-  python3 -m virtualenv {{VENV}} --quiet
-
-## Set up virtual environment and install requirements
-@requirements: venv
-  {{VENV_BIN}}/python -m pip --quiet install --upgrade pip
-  {{VENV_BIN}}/pip --quiet install -r requirements.txt
+  python3 -m venv {{VENV}}
```
universalmind303, line 6: `virtualenv` was used because it has support for more terminals.

tychoish: is that material in this case where we're just calling the paths' directly with these targets rather than using he `activate` script?

bindings/python/justfile
```diff
@@ -1,29 +1,26 @@
...
 
 @venv:
   python3 -c "import virtualenv" || python3 -m pip --quiet install virtualenv
   python3 -m virtualenv {{VENV}} --quiet
-
-## Set up virtual environment and install requirements
-@requirements: venv
-  {{VENV_BIN}}/python -m pip --quiet install --upgrade pip
-  {{VENV_BIN}}/pip --quiet install -r requirements.txt
+  {{VENV_BIN}}/python -m pip install --upgrade pip
```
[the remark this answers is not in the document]
tychoish: inside of a virutalenv all versions of python are the same thing... if it's python3, then you get both `python` and `python3` 

(the main source of this confusion is that archlinux went ahead and made `/usr/bin/python` python3 very early, and every other distro/system did not, because they had system level python dependencies (whereas arch's system tools are all in C or bash) so while the "standard" is to have `python3` and friends, and `python` within virtualenvs, it's kind of... a mess)

bindings/python/justfile
universalmind303: It does force everything to go through the justfile which is mildly annoying. I  `activate` and run things locally a lot so this disrupts my current workflow. I'd prefer to avoid that disruption.

greyscaled: If using `virualenv`, can we install it via `venv` first. This works on MacOS for me:

```shell
# Installs python dependencies for testing
venv:
  python3 -m venv .venv_temp
  ./.venv_temp/bin/pip3 install virtualenv
  ./.venv_temp/bin/python3 -m virtualenv {{VENV}}
  {{VENV_BIN}}/python -m pip install --upgrade pip
  {{VENV_BIN}}/python -m pip install poetry
```

greyscaled: I think I accidentally complicated matters.

I do need a path forward though as I currently am having trouble even running `just pytest`...whatever gets me there without interference of others' workflows is good enough

tychoish: Because the virtualenv helper doesn't matter functionally, and because @greyscaled's issue was preexisting, I've made this conditional (so you get `virtualenv` if you have it installed and otherwise you get `venv`). I think we should address Grey's issue/environment separately...

This patch is meant to fix the issue with the implicit dependencies in the pypi package, and I think it does a decent job of that.

tychoish:
@talagluck @universalmind303 can you take another look at this?

talagluck (approved):
Overall the approach feels reasonable, and I was able to successfully run the code, but I'll defer to Cory for the final sign-off.

PR #2485, opened 2024-01-24
fix: bson should handle nulls

### LINKED ISSUE #2484: Invalid null handling in mongo collections
#### mongo shell
```sh
test> db.createCollection("nulls_test")
{ ok: 1 }
test> db.nulls_test.insertMany([{ value: 1}, {value: null}])

```

#### glaredb
```sql
> SELECT * FROM read_mongodb('mongodb://localhost:27017/test', 'test', 'nulls_test');
Error: External error: External error: Unhandled element type to arrow type conversion; Null, Int32
```

## Expected
```
┌─────────────────┐
│ value           │
│              ── │
│           Int32 │
╞═════════════════╡
│               1 │
│            NULL │
└─────────────────┘
```

tychoish:
Closes #2484


tests/tests/test_bson.py
```diff
@@ -85,3 +85,22 @@ def test_read_bson(
...
+    
+    with open(data_path, "wb") as f:
+        for i in range(100):
+            f.write(bson.encode({"a": 1}))
+            
+        for i in range(10):
+            f.write(bson.encode({"a": None}))
+
+    with glaredb_connection.cursor() as curr:
+        curr.execute(f"select count(*) from '{data_path}'")
```
universalmind303, line 104: the `count(*)` query was previously working, We should explicitly check the values

universalmind303 (changes requested):

tests/tests/test_bson.py
tychoish: I ran this test without the patch and it failed. 

Because this uses the streaming bson table/file code, which uses the same bson-to-recordbatch as mongodb, we would expect the count to break here too.

scsmithr (approved):

universalmind303 (changes requested):
can we add the example in the original issue as a slt.

tychoish:
I think adding this as an SLT is pretty fragile and doesn't provide a lot of value. The code gets exercised. Given the quality of the mongo docker container we use and our mechanisms for getting data into it for SLTs fixtures, it seems precarious. I'm also sort of dubious about keeping the SLT fixtures in a shell script that's pretty far away from the actual tests in question. (local testing is hard, knowing where to fix it is hard, shell scripts like this are kind of grenades, etc.) 

I think there's a project to be had about adding better support for fixtures in the SLTs, though I'm not sure where it would make sense to prioritize. And in light of that, I'm not convinced that we need to bodge in tests just because...

universalmind303:
> I think adding this as an SLT is pretty fragile and doesn't provide a lot of value. 

I disagree. As a general rule, If an issue is opened with a specific reproducible example, I include that example in the tests to ensure that that specific case is tested against. This indirectly. Implicitly, i know that this bson stuff _should_ fix the mongo issue, but it doesn't explicitly test the example I provided.

I agree that we shouldn't be creating a bunch of test data in those shell scripts, but we can create a bson file with that data & include it in our `testdata` dir, and do a mongoimport. In our testdata, we have a script to generate iceberg data, then that data is checked into source control. We should do something similar with bson.

tychoish:
I've written an additional test.

PR #2478, opened 2024-01-22
fix: add auth for cassandra

tychoish:
someone else should test this more thoroughly.

PR #2486, opened 2024-01-24
test(cassandra): require auth

greyscaled:
## Description

This builds on top of #2478 and shouldn't merge until after that does.

This configures our cassandra test fixture to require auth. It just uses the default passwordAuthenticator, and only has the default superuser role (username = cassandra, password = cassandra).

Adding this in means we don't have a test for the "unauthenticated". To add a test for the "unauthenticated" scenario would be a bit of a pain given how it's configured through `cassandra.yaml` and requires node restarts. If we do add in a test for that scenario, I'd suggest creating a second cassandra container that doesn't mount in our custom `cassandra.yaml` and running a really basic smoke test, since the only thing we're testing that alters from the auth tests is that we can create data sources and use our table func without auth params.

### What are the changes to the script?

At a high level:

0. Build a custom image that mounts in `cassandra.yaml` that requires user/pass auth
1. Instead of creating a second container to seed data, just mount the files in the first time and use `docker exec`
2. Additional waits for the authentication role to apply
3. `-u cassandra` and `-p cassandra` to `cqlsh`

### Why do we need a new Dockerfile?

It's very hard to change `YAML` values in a running service (which requires restarting the service).
One might think "ah ok volume mount", which is the first thing I tried. Except each time I ran the
script, it would change some networking values. Having the git working tree dirty after each run
felt odd.

### What's with the `cassandra.yaml` ?

I ran the former script and copied the default config. I changed one value `authenticator` after following the documentation https://cassandra.apache.org/doc/stable/cassandra/operating/security.html and pasted it.

I tried pasting only a subset of it, and the container would fail to start. Felt it was safest to pass the entire thing in with that one value changed. This is why the diff is so large :(

### DIFF
--- a/scripts/create-test-cassandra-db.sh
+++ b/scripts/create-test-cassandra-db.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 #
 # Spins up a Cassandra docker container and loads it with test data.
-# To access the database, the use the default user/password 'cassandra'.
+# To access the database, use the default user/password 'cassandra'.
 
 set -e
 CONTAINER_NAME="glaredb_cassandra_test"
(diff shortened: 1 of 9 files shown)


scripts/create-test-cassandra-db.sh
```diff
@@ -1,40 +1,54 @@
 #!/usr/bin/env bash
-
-# Spins up a test cassandra docker container and loads it with data.
 #
-# By default, the container will start up a 'default' database an not require a
-# username or password.
+# Spins up a Cassandra docker container and loads it with test data.
+# To access the database, the use the default user/password 'cassandra'.
```
greyscaled, line 4:
```suggestion
# To access the database, use the default user/password 'cassandra'.
```

tychoish (approved):
g

Issue #2487, opened 2024-01-24
Error creating table from Pandas DataFrame


greyscaled:
If I switch `to_pandas` to `to_polars`, everything works as expected.

```python
import glaredb
import pandas as pd
import polars as pl

con = glaredb.connect("glaredb://...")

df = con.sql("SELECT * FROM './b.parquet'").to_pandas()

con.execute("create table test as SELECT * FROM df")
```

> External error: Execution error: failed to create table in storage: Schema error: Invalid data type for Delta Lake: Null

---

I did not test if this is specific to Parquet, and so it might be useful to just create a simple df in pandas and try the same thing.

I suggest we add a [Python test](https://github.com/GlareDB/glaredb/tree/main/tests/tests) covering this case.

universalmind303:
@greyscaled do you have an example of the parquet file you used for this?

greyscaled:
[parquet_file.zip](https://github.com/GlareDB/glaredb/files/14087628/parquet_file.zip)

universalmind303:
hmm, yeah i'm not seeing the same error, but a different one

```
exceptions.ExecutionException: Error during planning: Unable to fetch table provider for 'test': Not a Delta table: no log files
```

greyscaled:
@universalmind303 ah, this is my bad. The error referenced in the original post is when I attempt to query the table.

The Python script exits without error, but `SELECT * FROM ...` does. My apologies.

PR #2488, opened 2024-01-25
test(python): create table from pandas df

greyscaled:


greyscaled:
See #2487

bindings/python/tests/test_pandas.py
```diff
@@ -115,3 +115,43 @@ def test_execute():
 
     assert out.equals(expected)
     con.close()
+
+def test_create_table_from_dataframe():
```
greyscaled, line 119: I propose we merge with this test skipped, and it can be a task to unskip it when resolving #2487

universalmind303 (approved):

PR #2494, opened 2024-01-25
feat: streaming json document/array data 

tychoish:
As an alternate implementation of #2306 for #2218. 

I'm sort of partial to this implementation, as it requires a bunch less DF boiler plate, but would love other opinions.

There's some more work that should be done, a lot of which 
- [ ] testing
- [ ] discussion about what the API should be. We should probably write a function that chooses between ndjson and json based on the first non-whitespace character of a file/stream.

Other notes:
- I don't think this implementation is suitable for all json data, (though it will hande "one json document as a one row table" and "an array of json documents" quite well) even if we could get it to handle ndjson, without some work to sort of the schema inference: 
- my intention was that we _always_ read the entire dataset, get the superset of the schema and then return the data. We could do schema inference.
- as a result I sort of arbitrarily added some chunking and batch sizing. I definitely picked these randomly, and if there are better options, let me know. 
- while it does support globbing, it still reads everything into memory always.


crates/datasources/src/json/stream.rs
```diff
@@ -0,0 +1,82 @@
...
+use datafusion::execution::TaskContext;
+use datafusion::physical_plan::streaming::PartitionStream;
+use datafusion::physical_plan::{RecordBatchStream, SendableRecordBatchStream};
+use futures::Stream;
+use futures::StreamExt;
+use serde_json::{Map, Value};
+
+pub struct JsonStream {
+    schema: Arc<Schema>,
+    stream: Pin<Box<dyn Stream<Item = Result<RecordBatch, DataFusionError>> + Send>>,
```
scsmithr, line 19: Datafusion has a `SendableRecordBatchStream` alias for this which is usually easier to write.

crates/datasources/src/json/stream.rs
```diff
@@ -0,0 +1,82 @@
...
+
+impl RecordBatchStream for JsonStream {
+    fn schema(&self) -> SchemaRef {
+        self.schema.clone()
+    }
+}
+
+pub struct JsonPartitionStream {
+    schema: Arc<Schema>,
+    stream: Mutex<Option<Pin<Box<dyn Stream<Item = Result<RecordBatch, DataFusionError>> + Send>>>>,
```
scsmithr, line 38: Typically prefer parking_lot over std mutex.

scsmithr (approved):
Test and ship!

tychoish:
> Test and ship!

patience. 

I also kinda want to have external tables for this too? 

Also, how do we solve the ndjson/json disambiguation? have an option with a default?

crates/datasources/src/json/stream.rs
tychoish: much to my great displeasure, `SendableRecordBatchStream` is this just record batches, no results.

crates/datasources/src/json/stream.rs
tychoish: I tried to use parking_lot and the `.take()` trick that you used here (because I stole it from the bson thing you did) wouldn't compile. Not super sure.

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+
+    let schema = Arc::new(Schema::new(
+        field_set
+            .iter()
+            .map(|(k, v)| Field::new(k, v.to_owned(), true))
+            .collect::<Vec<_>>(),
+    ));
+
+    let chunks = data
+        .chunks(100)
```
universalmind303, line 61: 100 chunks seems like a pretty small default.

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+                schema.clone(),
+                chunk.to_vec().to_owned(),
+            ))
+        })
+        .collect::<Vec<_>>();
+
+    Ok(Arc::new(StreamingTable::try_new(schema.clone(), chunks)?))
+}
+
+fn push_unwind_json_values(
```
universalmind303, line 73: I'm partial to using `simd_json` over `serde_json`. Not something we need to do now, but the perf benefits of using `simd_json` make it quite appealing.

crates/datasources/src/json/stream.rs
```diff
@@ -0,0 +1,83 @@
...
+            stream: partition,
+        })
+    }
+}
+
+impl JsonPartitionStream {
+    pub fn new(schema: Arc<Schema>, chunk: Vec<Map<String, Value>>) -> Self {
+        let stream_schema = schema.clone();
+        let stream = futures::stream::iter(chunk)
+            .chunks(25)
```
universalmind303, line 67: why 25?

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+    let mut data = Vec::new();
+    for obj in list {
+        let blob = store.get(&obj.location).await?.bytes().await?.to_vec();
+        let dejson = serde_json::from_slice::<serde_json::Value>(blob.as_slice())?.to_owned();
+        push_unwind_json_values(&mut data, dejson)?;
+    }
+
+    let mut field_set = indexmap::IndexMap::<String, DataType>::new();
+    for obj in &data {
+        for key in obj.keys() {
```
universalmind303, line 45: Doesn't this iterate over all rows? 

We already are iterating over it when its passed into arrow's decoder. 

It seems reasonable that we could deserialized the json in a single pass instead of 2.

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+            .map(|(k, v)| Field::new(k, v.to_owned(), true))
+            .collect::<Vec<_>>(),
+    ));
+
+    let chunks = data
+        .chunks(100)
+        .map(|chunk| -> Arc<dyn PartitionStream> {
+            Arc::new(JsonPartitionStream::new(
+                schema.clone(),
+                chunk.to_vec().to_owned(),
```
universalmind303, line 65: we shouldn't be cloning these chunks here

crates/sqlbuiltins/src/functions/table/json.rs
```diff
@@ -0,0 +1,67 @@
...
+    const FUNCTION_TYPE: FunctionType = FunctionType::TableReturning;
+}
+
+#[async_trait]
+impl TableFunc for JsonScan {
+    fn detect_runtime(
+        &self,
+        args: &[FuncParamValue],
+        _parent: RuntimePreference,
+    ) -> Result<RuntimePreference, ExtensionError> {
```
universalmind303, line 32: Id expect this to error if you try to provide non matching urls. `read_json(['file://...', 'https://...'])`

crates/sqlbuiltins/src/functions/table/json.rs
```diff
@@ -0,0 +1,67 @@
...
+use protogen::metastore::types::catalog::RuntimePreference;
+
+use crate::functions::table::{table_location_and_opts, TableFunc};
+use crate::functions::{ConstBuiltinFunction, FunctionType};
+
+#[derive(Debug, Clone, Copy, Default)]
+pub struct JsonScan;
+
+impl ConstBuiltinFunction for JsonScan {
+    const NAME: &'static str = "read_json";
```
universalmind303, line 20: I'm fine with just `read_json` and `read_ndjson`. This is explicit & doesn't leave any ambiguity.

universalmind303 (changes requested):

crates/datasources/src/json/table.rs
tychoish: isn't this "chunks of 100" (likely I think there aren't likely to be relatively few documents in any of these tables)

crates/datasources/src/json/table.rs
tychoish: this package is mostly just leaning on the record batch builders from arrow_json which mostly expose a serde_json interface, is simd_json shimable here?

I think the ndjson stuff would benefit a lot from this; but this stuff, we're talking about, at most, megabytes of data, for this source (and you have to have the data buffered in memory, so even slow things are going to be fast by most of our standards...)

crates/datasources/src/json/stream.rs
tychoish: it's less than 100; Other numbers could be good too.

crates/datasources/src/json/table.rs
tychoish: the loop on 37 iterates over all files/urls (in the case of a glob) and in most cases will only run once. 

`push_unwind_...` doesn't iterate over the document, just builds the list of documents, erroring if we get something that isn't useable (a scalar). This is either a single document, or an array of documents, and unwinds the array when needed. We could extract the schema in this function,  _but_:
- I think the code would get more complicated, and this is pretty easy to understand.
- in the future where you might specify a schema to the table provider, the code is, in its current form easier to modify in that direction
- it would be easy to end up with code where you were were iterating less but doing more allocations to temporary data structures to keep track as you go (particularly if you have multiple files), and I feel like two loops are likely to be faster than one loop with a lot of allocation pressure. 
- my baseline assumption is that it's always going to be relatively small amounts of data (a few megs tops).
- in the error case you end up building a schema that you're never going to use, and errors might end up being pretty common just because of malformed REST apis.

crates/sqlbuiltins/src/functions/table/json.rs
tychoish: none of our other data sources have this semantic. 

I agree, though.

crates/sqlbuiltins/src/functions/table/json.rs
tychoish: the duckdb json interface is a bit cleaner in terms of format autodetection, which this really doesn't have and that might be confusing.

If you take a newline delimited json file or glob you'll get one document per-file, with this code.

crates/datasources/src/json/table.rs
tychoish: how big should chunks be? 

I think in general, being able to split this into a number of partitions is good, and means processing could be meaningfully async, but it'd be reasonable to just have one too.

crates/datasources/src/json/stream.rs
tychoish: Do you have a suggestion?

crates/sqlbuiltins/src/functions/table/json.rs
universalmind303: They did at one point. I remember writing the logic for this. 

If that got removed, we should open up an issue as this is a regression.

crates/sqlbuiltins/src/functions/table/json.rs
universalmind303: So i think the autodetection is nice, but I think making users aware of the implications of using ndjson vs json is more important. 

We can do a lot more optimizations with ndjson than we can with json, and the code paths are very different. We had some pretty lengthy discussions over at polars when adding ndjson functionality, and ended up making them two distinct functions for this reason.

crates/sqlbuiltins/src/functions/table/json.rs
scsmithr: The object store one does: https://github.com/glaredb/glaredb/blob/c8275d47909a5f5b0380055b5b192510f9b340c7/crates/sqlbuiltins/src/functions/table/object_store.rs#L221-L236

Which calls into this: https://github.com/glaredb/glaredb/blob/c8275d47909a5f5b0380055b5b192510f9b340c7/crates/sqlbuiltins/src/functions/table/object_store.rs#L309-L348

crates/datasources/src/json/table.rs
universalmind303: when working on the polars json parser, we ended up landing on `1000` as the default chunk size. I don't have the benchmarks anymore, but it seemed to be the sweet spot in partition sizes for most cases.

crates/datasources/src/json/stream.rs
universalmind303: I guess i don't understand why we chunk it, then chunk it again?

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    // for consistent results, particularly for the sample, always
+    // sort by location
+    list.sort_by(|a, b| a.location.cmp(&b.location));
+
+    let mut data = Vec::new();
+    for obj in list {
+        let blob = store.get(&obj.location).await?.bytes().await?.to_vec();
+        let dejson = serde_json::from_slice::<serde_json::Value>(blob.as_slice())?.to_owned();
```
universalmind303, line 39: do we need to do `to_owned` here? 

We should be able to implement some copy-on-write semantics instead of immediately copying to a buffer. We are doing another `to_owned` later on `l:65`, so this'll result in multiple copies of data. 

It'd be preferred to act on the borrowed values instead.

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    };
+    Ok(())
+}
+
+fn type_for_value(value: &Value) -> DataType {
+    match value {
+        Value::Array(v) => {
+            if v.is_empty() {
+                DataType::List(Arc::new(Field::new("", DataType::Null, true)))
```
universalmind303, line 106: I think datafusion prefers using "item" for list field names

crates/datasources/src/json/stream.rs
```diff
@@ -0,0 +1,83 @@
...
+            .map(move |objs| {
+                let mut decoder =
+                    ReaderBuilder::new(stream_schema.clone().to_owned()).build_decoder()?;
+                decoder
+                    .serialize(&objs)
+                    .map_err(|e| DataFusionError::External(Box::new(e)))?;
+                Ok(decoder.flush()?.unwrap())
```
universalmind303, line 74: this seems really inefficient. We first deserialize the json buffer to a `serde_json::Value`, then we re-serialize just to deserialize it again as a RecordBatch?

so it goes from `&[u8]` -> `Vec<Map<String, Value>>` -> `arrow_json::Tape` -> `RecordBatch`

crates/datasources/src/json/table.rs
universalmind303: > my baseline assumption is that it's always going to be relatively small amounts of data (a few megs tops).

I don't think we should be making any assumptions about size of data, especially considering the globbing support. I could see someone trying to read in a directory of hundreds or thousands of small/medium json files.

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+                continue;
+            }
+            field_set.insert(key.to_string(), type_for_value(obj.get(key).unwrap()));
+        }
+    }
+
+    let schema = Arc::new(Schema::new(
+        field_set
+            .iter()
+            .map(|(k, v)| Field::new(k, v.to_owned(), true))
```
universalmind303, line 56:
```suggestion
            .into_iter()
            .map(|(k, v)| Field::new(&k, v), true))
```

crates/datasources/src/json/table.rs
universalmind303: > I think in general, being able to split this into a number of partitions is good, and means processing could be meaningfully async, but it'd be reasonable to just have one too.

Don't the chunks already contain the deserialized data, which is the most expensive part. Isn't the stream just creating recordbatches from the chunks? If we were to defer the cpu bound tasks (deserialization) to the stream, then I think this would make more sense to chunk. In this scenario, I think the chunking may introduce more partitioning overhead than benefits. _(we'd probably want to benchmark it to be sure, but that's just my suspicion)._

crates/datasources/src/json/stream.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+impl JsonPartitionStream {
+    pub fn new(schema: Arc<Schema>, chunk: Vec<Map<String, Value>>) -> Self {
+        let stream_schema = schema.clone();
+        let stream = futures::stream::iter(chunk)
+            .chunks(25)
+            .map(move |objs| {
+                let mut decoder =
+                    ReaderBuilder::new(stream_schema.clone().to_owned()).build_decoder()?;
```
universalmind303, line 70: why are we doing `.clone().to_owned()`? can't we just do `.clone()`

crates/datasources/src/json/table.rs
universalmind303, line 39: should avoid multiple allocations for the slice. 

```suggestion
        let blob = store.get(&obj.location).await?.bytes().await?;
        let dejson = serde_json::from_slice::<serde_json::Value>(&blob)?.to_owned();
```

crates/datasources/src/json/table.rs
tychoish: I'm down for 1000

crates/datasources/src/json/table.rs
```diff
@@ -0,0 +1,135 @@
...
+
+fn type_for_value(value: &Value) -> DataType {
+    match value {
+        Value::Array(v) => {
+            if v.is_empty() {
+                DataType::List(Arc::new(Field::new("", DataType::Null, true)))
+            } else {
+                DataType::List(Arc::new(Field::new(
+                    "",
+                    type_for_value(v.first().unwrap()),
```
universalmind303, line 110: This'll definitely cause some issues later on. I did this in the polars json parser & we ended up getting a bunch of issues from parsing errors due to schema mismatching. We ended up needing to infer the super type for the values in the array. 

for example, an array of `[null, "some string", "some other string"]` will result in `List<Null>` instead of `List<Utf8>`.

similarly, an array of `[0, 1.0, 2.2]` will get parsed as `List<Int64>` instead of `List<Float64>`

universalmind303: At a minimum, we should error out if the values in the array are of different types.

crates/datasources/src/json/stream.rs
tychoish: these streaming tables work by collecting a number of streams (from files, though in this case we will have read them all into memory because we need/want to get the schema (and do that losslessly)). Then the stream produces record batches.

The first chunk, gives us the partitions, the second gives us the size of the record batch. 

I (following from your earlier comment) made the recordbatch size 1000, and the stream size as 10,000.

crates/datasources/src/json/table.rs
tychoish: There are a couple of assumptions here:
- the primary use case will be pulling information off of HTTP endpoints. It's just convenient to use to object store stuff. We want to say "point glaredb at <any rest api that produces json> and you get a SQL table." files with arbitrary data is fine too but a sort of just something that comes for free.
- there are a lot of things that are kind of inefficient here because I want the experience in the above use case to not lose data (e.g. drop something from later documents because it's not in the schema). When we let folks pass schema clauses into functions and external tables, we can update these functions and also provide schema inference, and then we should change the features.

Unless we want to shelve this for a while while we figure out how to handle the schema, then all the data have to be in memory, or we have to parse all the files twice. I opted for the former based on the assumptions about the use case. I'm not opposed to addressing the "many small json files" or "data size larger than (I dunno) a gig" use cases, but I'm also fine if we punt on those, as there are workarounds (use polars) or (convert to ndjson), and it would not be difficult to add schema inference or explicit schema and/or lazy file reading to this implementation.

crates/sqlbuiltins/src/functions/table/json.rs
tychoish: It makes sense here for this function, and I'm willing to go with this naming, but we don't have a good answer for `CREATE EXTERNAL TABLE` which calls `ndjson` ... `json`.

crates/datasources/src/json/stream.rs
tychoish: you are correct.

you can't make a record batch or or an arrow-json decoder without a schema, and you need to examine the document's keys/values to get the schema. 

on the plus side, it's all in memory and the amount of time you spend fiddling with the rust abstractions is definitely a fraction of the time you spend reading the data off the filesystem or out of object_store/http.

crates/datasources/src/json/table.rs
universalmind303: I mostly agree with these assumptions, but I do think there is another path forward _(particularly for the multi file case)_. I don't think the assumption of *all the data have to be in memory* is correct. 

At any given time, we only need to hold a **single file** in memory. All of our file based readers expect the same schema across multiple files/globs. So we can apply that same logic here, and only use the first file to extract the schema. From there, we can defer fetching & parsing of the remaining files into the streams. 

This would be especially useful for _(eventually)_ limit pushdowns. If you only want `limit 1000` of thousands of files and each file is 1000, you could stop after the first file, saving a lot of resources (both IO + CPU).

crates/datasources/src/json/table.rs
tychoish: yeah, I thought of this and decided to not care initially. There are a lot of edge cases that we'll continue to not handle very well (`[nul, "foo", 1]` plus arrays of objects...

I can do something replaces nulls with something that isn't a null (that'd cover a lot of cases) sort of everything else gets much harrier.

crates/datasources/src/json/table.rs
tychoish: The streaming table handling basically implicitly pushes down limits because the streams are iterated lazily. 

I implemented the "all of the first file, nothing more" schema inference. I think this will be ok for now.

The globbing definitely doesn't work (or at least I couldn't provoke it to work); it's probably a regression, but the bson code is subject to the same flaw, and we're not doing anything except using the library functions, which at least for now renders the multi-file use case somewhat moot. 

We can dig into this next week.

universalmind303:
can we please add some slt's. I don't think we should use pytests as the default way of writing tests. The SLT's ensure that the functionality works across multiple protocols. Additionally, they are subjectively easier to read and maintain than the pytests.

tychoish:
> can we please add some slt's. I don't think we should use pytests as the default way of writing tests. The SLT's ensure that the functionality works across multiple protocols. 

I don't think you need full coverage across data sources and protocols: 
- you need to test that different query shapes and operations work over different protocols, but there are no new query shapes here.
- we're not consistent with "testing across different protocols" already: none of the service integration tests currently run on postgres, or flightsql; same with snowflake.

Ideally there should be a small core of tests that run across all protocols with the remainder of the tests executed in whatever way is fast. At the moment there's no real downside to running a bunch of local tests/files with the same tests over different protocols, but if the tests took a long time, I wouldn't hesitate to drop things like the minio flight tests or the sqlserver flight test (actually, we should do this one, because we wait 6 minutes a build for it.)

> Additionally, they are subjectively easier to read and maintain than the pytests.

I understand your preference here, but The SLTs frankly have terrible ergonomics for anything related to the filesystem, the fixtures are unintuitive, and the assertions are awkward enough. Though not applicable in this particular case (yet) SLTs don't provide any reasonable way to test `COPY TO`. We end up testing, effectively, GlareDB _with itself_ and we end up with a lot of blindspots as a result. 

The maintenance thing is tricky: the fixtures get setup in a lot of cases pretty far from the the actual assertions and there's no language server for SLTs, so you just have to grep wildly and hope than you've found the right thing. This is complicated by a lot of copy and pasted test cases. There's also no fixture _cleanup_ within a "suite". This is often fine because people have made sure to make tests not collide (but people make mistakes), but moving tests around can't always be safe and the execution order and statefulness assumptions in a lot of the integration-type tests are a weakness both of the assertion and of the tests themselves. 

There are lots of tests that shouldn't be tested with the pytests, and the SLTs are great for testing the semantics of SQL, (which makes sense) but pretty poorly suited to integration testing, and multistatement operations. In any case, the product is better for having multiple testing modalities and tools.

tychoish:
testing modality aside, I think the real question is how we square the fact that `ndjson` is `json` when creating an external table or for file extension sniffing, and this would confuse that.

Issue #2436, opened 2024-01-17
Use glaredb with dbt/great expectations: fix postgres functionality


talagluck:
## Description
When attempting to create a connection from dbt to a glaredb cloud instance using a postgresql connection string, I hit the following error:
```terminal
14:37:27  1 check failed:
14:37:27  dbt was unable to connect to the specified database.
The database returned the following error:

  >Database Error
  error with status PGRES_EMPTY_QUERY and no message from the libpq

Check your database credentials and try again. For more information, visit:
https://docs.getdbt.com/docs/configure-your-profile
```

Given that dbt should work fine with a postgres connector, I would expect this to work, but there's a mismatch here. Sean mentioned that he tested this a few months back and it worked, so there may have been a regression here.

I also get the same error when attempting to run Great Expectations with GlareDB - Great Expectations uses SqlAlchemy under the surface.

talagluck:
Steps to reproduce for Great Expectations:
```terminal
pip install "great_expectations[postgresql]"
```

In a notebook (or python script) run:
```python
import great_expectations as gx
context = gx.get_context() # gets a great expectations project context 
ds = context.sources.add_postgres(name="glaredb", connection_string=<CONNECTION_STRING>)
```
The connection string I used was `"postgresql://6AhiEN7GQDmo:<PASSWORD>@o_PRocU0j.proxy.glaredb.com:6543/rough_glitter"`

tychoish:
I'm _pretty sure_ that this is #2273, and if you can you specify `autocommit=True` in some way that it gets set on the psycopg2 (I found [some docs](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#empty-dsn-connections-environment-variable-connections) but I don't know the best way here.) and see if it helps, that'd be cool.

scsmithr:
Good news, I think I got further (thanks to correcting the transaction stubs and connecting to qa). Bad news, seems great expectation is expecting a `pg_catalog.version()` function:

```
~/Code/github.com/glaredb/glaredb/bindings/python [1] % just example ge
../../.venv/bin/python examples/ge.py
Traceback (most recent call last):
  File "/Users/sean/Code/github.com/glaredb/glaredb/.venv/lib/python3.11/site-packages/sqlalchemy/engine/base.py", line 1910, in _execute_context
    self.dialect.do_execute(
[...]

PR #2523, opened 2024-01-29
fix: add pg_matview table

universalmind303:
@talagluck this should fix the great-expectations issue you referenced in https://github.com/GlareDB/glaredb/issues/2436

### DIFF
--- a/crates/sqlbuiltins/src/builtins.rs
+++ b/crates/sqlbuiltins/src/builtins.rs
@@ -689,9 +689,9 @@ FROM (VALUES (NULL, NULL, NULL, NULL,
     NULL, NULL, NULL, NULL, NULL)) WHERE false",
 });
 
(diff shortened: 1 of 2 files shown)


crates/sqlbuiltins/src/builtins.rs
```diff
@@ -689,6 +689,20 @@ FROM (VALUES (NULL, NULL, NULL, NULL,
...
+    schema: POSTGRES_SCHEMA,
+    name: "pg_matviews",
+    sql: "
+    SELECT '' as schemaname,
+    '' as matviewname,
+    '' as matviewowner,
+    '' as tablespace,
+    false as hasindexes,
+    false as ispopulated,
+    '' as definition
```
scsmithr, line 702:
```suggestion
    SELECT '' as schemaname,
    '' as matviewname,
    '' as matviewowner,
    '' as tablespace,
    false as hasindexes,
    false as ispopulated,
    '' as definition
    WHERE false
```

To avoid returning a row.

PR #2524, opened 2024-01-29
chore: python cleanup

universalmind303:
- isolates the python virtualenvs (root, tests, tpch, bindings/py)
- adds python formatting to the top level `just fmt` 
- applies python formatting to all py files.


universalmind303:
i have no idea why this stage is failing. It works fine locally, but get random unrelated errors in CI.

ruff.toml
```diff
@@ -0,0 +1,9 @@
+line-length = 88
```
tychoish, line 1:
```suggestion
line-length = 100
```

ruff.toml
```diff
@@ -0,0 +1,9 @@
...
+
+[lint]
+ignore = [
+  "F841", # unused variables
+  "F401", # unused imports,
+  "F811", # redefined variables
```
tychoish, line 7: can we put these ignores at a more tightly scoped level?

justfile
```diff
@@ -118,18 +126,15 @@ protoc:
...
-  {{VENV_BIN}}/poetry -C tests install
-
+  
 
-# Runs pytest in the tests directory.
-pytest *args: 
-	{{VENV_BIN}}/poetry -C tests run pytest --rootdir={{invocation_directory()}}/tests {{ if args == "" {'tests'} else {args} }}
+## Set up virtual environment and install requirements
+@requirements: venv
+  {{VENV_BIN}}/python -m pip --quiet install --upgrade pip
```
tychoish, line 136: I think this checks pypi for an up-to-date pip every time, including on every `fmt` call, right?

.github/workflows/ci.yaml
```diff
@@ -312,11 +312,8 @@ jobs:
         with:
           path: |
             tests/.venv/
-          key: ${{ runner.os }}-poetry-${{ hashFiles('**/Cargo.lock') }}
-      - name: setup pytest
-        run: just venv
-      - name: run pytest
-        run: just pytest
+          key: ${{ runner.os }}-poetry-${{ hashFiles('**/poetry.lock') }}
```
tychoish, line 315: If we make this change, I think we need to add python caching to the fmt task.

justfile
```diff
@@ -83,12 +87,15 @@ rpc-tests:
     --exclude "sqllogictests/functions/postgres"
 
 #  Check formatting.
-fmt-check: protoc
+fmt-check: protoc requirements
   cargo +nightly fmt --check
+  {{VENV_BIN}}/ruff check .
```
tychoish, line 92: I have a (very small) amount of worry that this will make the CI results a bit more confusing.

I also.... don't know that I'd block a build because of a formatting error in the (non-bindings) python code? It's probably fine, and I think having a formatter around is good, but I also wouldn't care if it didn't block CI.

tychoish (commented):
I'm a little concerned that putting all of our python execution in one virtual environment will cause what are fundamentally different projects to (potentially) interact in confusing ways. 

Virtual environments are pretty cheap: we can have one for the bindings, one for the pytests, one for the benchmarks, and one for development tooling (or avoid centralized formatting

PR #2525, opened 2024-01-29
chore: update fake-gcs-server image

greyscaled:


greyscaled:
errors, looks like we need the fork

PR #2526, opened 2024-01-30
chore: skip sqlserver+flight combination.

tychoish:
This (on top of #2518) should get wall clock time to about 10 minutes for an entire PR run, at the same cost. 

It's likely redundant to the other suite. We could also switch datasource integration to just postgres as there's less postgres protocol coverage.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -40,9 +40,7 @@ jobs:
             target/
             !target/**/glaredb
           key: ${{ runner.os }}-cargo-${{ hashFiles('**/Cargo.lock') }}
-
       - run: just build
-
       - uses: actions/cache/save@v4
         name: glaredb cache
         with:
@@ -77,6 +75,7 @@ jobs:
       - run: rustup component add rustfmt --toolchain nightly
       - run: just fmt-check
 
+
   lint:
     name: Lint (clippy)
     runs-on: ubuntu-latest-8-cores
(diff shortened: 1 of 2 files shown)


tychoish:
Screenshot Capture - 2024-01-30 - 01-38-17

Not nothing. This is almost best case (good caches,) there were some scheduling delays, but we're mostly hitting the "compile+bindings" DAG limitation (which means there's another 4ish minutes of wall clock time that we could get. 

The runner is one of the small ones so this is budget neutral, but this will push build times from 10-15 minutes into 5-10 minutes.

vrongmeal:
> on top of #2949

I think you tagged a wrong issue/PR

***

Why this change though? Is it that Flight SQL is still in development so we can afford to skip some checks for now that take very long?

tychoish:
Did a bit of buddy-system investigation with @scsmithr:
- each run of the sqlserver tests takes 4m30s, 
- the large_table tablescan is the slow thing. we do it four times: each run has a count and a full scan and we do the full run against different protocols. 
- Pushing down the count will probably be the biggest win here, and then we'd be able to (if we want), run flight+rpc in parallel, which would make sqlserver tests run in about as much time as the longer of the other datasource tests.
- the fixtures for setting up the test (the shell script) is slower than it should be, there's some speed gains there.
- there is an errant print statement in th sql server implementation.

scsmithr (approved):

Issue #2527, opened 2024-01-30
Klickhouse fails to convert negative timestamps


vrongmeal:
## Description

There are issues with the Klickhouse lib that we use:

https://docs.rs/klickhouse/latest/klickhouse/struct.DynDateTime64.html

This data type stores any time as u64 whereas clickhouse does store negatives (i64). So times like 1900-01-01 00:00:00 cannot be converted (since signed integers are converted to unsigned int).

vrongmeal:
Plan of action:

- [ ] Create a fork in @GlareDB organization
- [ ] Make changes in the fork
  - [ ] Use changes in the fork
  - [ ] Raise a PR upstream (hopefully it's accepted)

PR #2528, opened 2024-01-30
fix: Use actual precision for clickhouse data types

vrongmeal:
Related: #2527


crates/datasources/src/clickhouse/convert.rs
```diff
@@ -5,20 +5,9 @@ use std::task::{Context, Poll};
...
-    Int64Array,
-    Int8Array,
-    StringBuilder,
-    TimestampNanosecondBuilder,
-    UInt16Array,
-    UInt32Array,
-    UInt64Array,
+    Array, BooleanBuilder, Date32Builder, Float32Array, Float64Array, Int16Array, Int32Array,
+    Int64Array, Int8Array, StringBuilder, TimestampMicrosecondBuilder, TimestampMillisecondBuilder,
+    TimestampNanosecondBuilder, TimestampSecondBuilder, UInt16Array, UInt32Array, UInt64Array,
```
tychoish, line 10: I think your text editor may have the old formatter configured.

crates/pgrepr/src/scalar.rs
```diff
@@ -235,17 +296,29 @@ impl Scalar {
...
+                let secs = v.num_seconds_from_midnight() as i32;
+                let sub_millis = (v.nanosecond() / 1_000_000) as i32;
+                let millis = (secs * 1_000) + sub_millis;
+                DfScalar::Time32Millisecond(Some(millis))
+            }
             (Self::Time(v), ArrowType::Time64(TimeUnit::Microsecond)) => {
-                let nanos = get_nanos_from_time(&v);
-                let micros = nanos_to_micros(nanos);
+                let secs = v.num_seconds_from_midnight() as i64;
+                let sub_micros = (v.nanosecond() / 1_000) as i64;
```
tychoish, line 310: is this just truncation or does it round?

crates/datasources/src/clickhouse/convert.rs
```diff
@@ -228,6 +215,78 @@ fn column_to_array(
...
+                }
+            }
+            Arc::new(vals.finish().with_timezone_opt(tz))
+        }
+        DataType::Timestamp(TimeUnit::Microsecond, tz) => {
+            let mut vals = TimestampMicrosecondBuilder::with_capacity(column.len());
+            for val in column {
+                match val {
+                    Value::DateTime(v) => vals.append_value(
+                        DateTime::<Tz>::try_from(v)
```
tychoish, line 271: do these conversions work across resolutions too? maybe good to have a comment about their rounding/translation strategy.

tychoish (commented):
seems good!

crates/datasources/src/clickhouse/convert.rs
vrongmeal: What do you mean by resolutions?

crates/pgrepr/src/scalar.rs
vrongmeal: Just truncation. Here, we expect to get data in micros so there shouldn't be any need for rounding. The library doesn't provide a function to get sub micro seconds, so this is the only way...

crates/datasources/src/clickhouse/convert.rs
tychoish: I think what I'm asking is "what are the cases where the `try_from` fails? Is it out-of-bounds issues only?

tychoish (approved):
just have a minor curiosity

crates/datasources/src/clickhouse/convert.rs
vrongmeal: Yup, it's only out-of-bound cases.

PR #2529, opened 2024-01-30
chore: ci reorganize dag

tychoish:
Right we essentially compile four or five times (build+2 bindings+unit-tests+clippy). and they can't usefully share artifacts, but we insert a dependency between the build step and everything else so that if compile breaks we don't run anything else. A lot of things (including, I believe, the unit-tests) depend on the glaredb binary, so that makes sense but the bindings can build in parallel with the main cli binary. 

The challenge/impediment, now, is the unit-test->glaredb dependency (integration tests in the glaredb crate). If that can be built and run separately, we can run it with the pytests or a new task and wall clock times would be in the 5-7 minute range for the entire pipeline. 

The big downside is that if you push a broken compile that doesn't also break the formater (so many typos and syntax issues will still be caught,) then it has to fail four times, but this is classically a good trade in terms of increasing developer responsiveness. The other downside, is the first build after a Cargo.lock change (so no cache hits) will take more build time.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,7 +16,6 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest-8-cores
-    needs: ["fmt"]
     steps:
       - name: checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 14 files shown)


tychoish:
The last build in this branch exhibits some more worst-case behavior: the main build takes closer to 5 minutes (cache is less relevant because other branches are building concurrently.) 

Seeing these results makes me inclined to do per-branch caches, but need to experiment in practice.

scsmithr (approved):

PR #2538, opened 2024-01-30
feat: add openai_embed function

universalmind303:
todo: 

- [ ] integration tests & slts


crates/sqlbuiltins/src/functions/scalars/openai.rs
```diff
@@ -0,0 +1,245 @@
...
+                    ),
+                    _ => return Err(DataFusionError::Plan("Invalid argument".to_string())),
+                },
+                ColumnarValue::Scalar(ScalarValue::Utf8(v) | ScalarValue::LargeUtf8(v)) => {
+                    EmbeddingInput::StringArray(vec![v.clone().unwrap_or_default()])
+                }
+                _ => return Err(DataFusionError::Plan("Invalid argument".to_string())),
+            };
+
+            let client = Client::with_config(OpenAIConfig::new().with_api_key(creds.clone()));
```
tychoish, line 204: is there a way to specify a timeout?

crates/sqlbuiltins/src/functions/scalars/openai.rs
```diff
@@ -0,0 +1,245 @@
...
+            let embed = client.embeddings();
+            let res = embed.create(CreateEmbeddingRequest {
+                model: model.to_string(),
+                input: input,
+                encoding_format: Some(EncodingFormat::Float),
+                user: None,
+                dimensions: None,
+            });
+
+            // no way around blocking here. Expressions are not async
```
tychoish, line 214: seems like a user could easily use this function to take down a node... might want/need to feature flag it into only the bindings?

Is there *any* way to avoid this even if it's gross or a lot of work? not necessarily to do it, but just to understand the decision better.

universalmind303: No, expressions are not async, so we pretty much have to block on this. AFAIK, there is no way around this without rewriting datafusion exprs. 

IMO, we're better off doing some rate limiting, throttling, and aggressive timeouts than to try and avoid blocking here. I agree that it's not ideal, but there isn't much we can do.

crates/sqlbuiltins/src/functions/scalars/openai.rs
universalmind303: probably, this just uses reqwest clients under the hood, so timeouts should be doable. 

We can add timeouts in a followup.

crates/sqlbuiltins/src/functions/scalars/openai.rs
tychoish: I definitely think if we do this implementation that it needs to have _some_ kind limitation on it so that it won't bring down cloud or allow one user to DoS a cloud node, or we should feature flag it so it only exists in local mode. 

Can't execution plans and table/funcs be implemented as streams which are async? Is there a way to have a scalar function call a table function, or implement this as a table function? (mostly just spitballing).

crates/sqlbuiltins/src/functions/scalars/openai.rs
tychoish: https://docs.rs/async-openai/latest/async_openai/struct.Client.html#method.with_http_client <- yeah seems straightforward.

crates/sqlbuiltins/src/functions/scalars/openai.rs
```diff
@@ -0,0 +1,252 @@
...
+}
+impl FromStr for EmbeddingModel {
+    type Err = datafusion::error::DataFusionError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "text-embedding-3-small" => Ok(EmbeddingModel::TextEmbedding3Small),
+            "text-embedding-ada-002" => Ok(EmbeddingModel::TextEmbeddingAda002),
+            "text-embedding-3-large" => Ok(EmbeddingModel::TextEmbedding3Large),
+            _ => Err(DataFusionError::Execution("Invalid model name".to_string())),
```
scsmithr, line 93: Would be nice if we could return the available models in this error to keep the user from needing to guess.

crates/sqlbuiltins/src/functions/scalars/openai.rs
scsmithr: Datafusion async UDF issue: https://github.com/apache/arrow-datafusion/issues/6518

I think given the current limitation, a table func would be better here. It would also provide the option of returning a "column" instead of an array which might make it easier to run math functions on (I have no evidence on this last part).

universalmind303: I'm still not sure we should flat out forbid using this technique in functions. After all, when we eventually support UDF's we'll have no control over what those exprs are doing. 

This is much more controlled. OpenAI already has pretty aggressive [rate limits](https://platform.openai.com/docs/guides/rate-limits?context=tier-free), and the need for a user provided credential would likely result in rate limits from OpenAI well before anything negative would happen to our cloud infra. 

While I do think this is a valid concern, I don't think the risk surface area is that large either. 

I'd much prefer keeping this as an expression instead of trying to force it into something it's not _(table function)_.

universalmind303: I also think the idea of feature flagging is somewhat appealing here. We could feature flag this out of the shared cloud instances & reserve it for embedded, self-deployed, or a dedicated cloud instance.

scsmithr: I'd prefer this not being feature flagged. Having feature flags combined with our current configuration will make it easier to miss things during testing.

scsmithr (approved):

tychoish (approved):

PR #2554, opened 2024-02-01
chore: remove per-pr snowflake cleanup

tychoish:
per conversation with @vrongmeal


tychoish:
actually I read through the code and this might actually work, but I think we need to take off the "merged only" if because it's definitely leaking if you close a PR without merging it.

PR #2556, opened 2024-02-01
ci(python, node): update macos runners

greyscaled:


.github/workflows/nodejs-release.yml
```diff
@@ -38,7 +38,7 @@ jobs:
       fail-fast: true
       matrix:
         settings:
-          - host: macos-12-xl
+          - host: macos-12-large
```
universalmind303, line 41: could we just use the new mac-14's across the board instead of using 12 for some, 13 for others?

greyscaled: This is what we're doing in all the other actions, and so I'd just want to make it uniform and better understand why we have a difference

PR #2557, opened 2024-02-01
ci: fix pypi release

universalmind303:


.github/workflows/python-release.yaml
```diff
@@ -126,7 +126,7 @@ jobs:
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
-          target: ${{ matrix.target }}
+          target: ${{ matrix.target }}-apple-darwin
```
universalmind303, line 129: AFAIK, `x86_64` and `aarch64` arent valid targets, they should be they fully qualified ones `aarch64-apple-darwin`

https://www.maturin.rs/config#target-specific-maturin-options

greyscaled: Yea, I noticed this in #2556 as well.

.github/workflows/python-release.yaml
```diff
@@ -108,7 +108,7 @@ jobs:
 
 
   macos:
-    runs-on: macos-12-xl
+    runs-on: macos-14-xlarge
```
greyscaled, line 111: cc @tychoish & @scsmithr you have context from #2543

greyscaled (approved):

.github/workflows/python-release.yaml
tychoish: we should definitely just mirror whatever we're doing for the CLI.

my logic for choosing the 13 was just "use the oldest and smallest thing that works.

.github/workflows/python-release.yaml
tychoish: is this a recent change?

tychoish (approved):

.github/workflows/python-release.yaml
```diff
@@ -113,13 +111,17 @@ jobs:
...
       matrix:
-        target: [x86_64, aarch64]
+        config:
+        - target: x86_64-apple-darwin
+          runner: macos-13-large
+        - target: aarch64-apple-darwin
+          runner: macos-13-xlarge
 
-    name: Build Python Bindings (macos ${{ matrix.target }})
+    name: Build Python Bindings (macos ${{ matrix.config.target }})
```
greyscaled, line 124: I think we need to make this change again below (line 154 if I'm reading the diff right) where we post to slack

PR #2558, opened 2024-02-01
ci test: testing out mac 14 runners

universalmind303:


.github/workflows/python-release.yaml
```diff
@@ -43,14 +39,42 @@ jobs:
...
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+          architecture: x64
+      - name: Build wheels - universal
+        uses: PyO3/maturin-action@v1
+        with:
+          args: --target universal2-apple-darwin --out dist
+          working-directory: bindings/python
+          sccache: "true"
```
tychoish, line 75: I think we don't want this. We want caches to be off for release builds (to avoid having something unexpected from a previous build to leak in,) and even so it we're pretty unlikely to be able to use the cache, and it would take some amount of time to upload.

.github/workflows/python-release.yaml
```diff
@@ -43,14 +39,42 @@ jobs:
...
           container: "off"
+          maturin-version: ${{ env.MATURIN_VERSION }}
+      - name: Upload wheels
+        uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: bindings/python/dist
+
+  macos-universal:
+    runs-on: macos-14
```
tychoish: these are the small/default runners?

.github/workflows/python-release.yaml
```diff
@@ -16,21 +18,17 @@ permissions:
...
 
 jobs:
-  linux:
-    runs-on: ubuntu-2004-8-cores
-    strategy:
-      fail-fast: true
-      matrix:
-        target: [x86_64]
+  macos-x86_64:
+    runs-on: macos-13
```
tychoish, line 31: I think this host type might be too small to complete the build.

tychoish:
this should be a draft PR right?

PR #2564, opened 2024-02-02
chore(release): atomic build and push jobs

tychoish:


.github/workflows/release.yaml
```diff
@@ -13,8 +13,33 @@ env:
...
+    # can depend on all build tasks.
+    name: Atomic Release Release
+    runs-on: ubuntu-latest
+    if: github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/v')
+    needs:
+      - python-linux
+      - python-macos
+      - python-windows
+      - kernel-dist
+      - node-build
```
greyscaled, line 32: question: haven't looked below yet, just sort of came across my mind - why does `python` need 3 separate jobs, but `kernel-dist` and `node-build` don't?

.github/workflows/release.yaml
```diff
@@ -65,10 +84,34 @@ jobs:
               "text": "Build release binary result (${{ matrix.target }}): ${{ job.status }}\n${{ github.event.pull_request.html_url || github.event.head_commit.url }}"
             }
 
-  publish:
-    name: Publish
+
+  docker-push:
+    name: Build and Push Docker Image
+    runs-on: ubuntu-latest-8-cores
+    needs: atomic-build
```
greyscaled, line 91: This here is called `atomic-build`, but the job at the top is called `atomic-release`.

.github/workflows/release.yaml
tychoish: just how it's factored. the kernel and node are also 3 jobs, but they are matrixes and so we can use the group reference.

I think you could factor this as a single task, and get rid of the empty `atomic-release` task, but it'd be much harder to follow, and this was a simpler migration. These are pretty hard configurations to test (and are pretty expensive,) so I went for minimal impact.

.github/workflows/release.yaml
tychoish:
```suggestion
    needs: atomic-release
```

.github/workflows/release.yaml
```diff
@@ -1,4 +1,4 @@
-name: Build Release
+name: Release
```
greyscaled, line 1: How does this sound?

```suggestion
name: Atomic Build and Release
```

.github/workflows/release.yaml
```diff
@@ -8,13 +8,40 @@ on:
   push:
     tags:
       - "*"
+    branches:
+      - tycho/release-atomically
```
greyscaled, line 12: TODO: remove debug

greyscaled (changes requested):

.github/workflows/release.yaml
```diff
@@ -250,27 +246,23 @@ jobs:
...
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
       - name: Build wheels
         uses: PyO3/maturin-action@v1
         with:
+          target: ${{ matrix.config.target }}
           args: --release ${{ matrix.config.target }} --out dist
           working-directory: bindings/python
           container: "off"
```
tychoish, line 268: here we pass release the target and out dist, also target is its own argument to the action.

.github/workflows/release.yaml
```diff
@@ -232,15 +229,14 @@ jobs:
           name: wheels
```
tychoish, line 229: if you expand above this line, we pass no argument to release here, target is specified on its own line.

tychoish (commented):
based on the rendering of the command https://github.com/GlareDB/glaredb/actions/runs/7760755338/job/21167722566#step:6:1536

I think we are passing to the macos builds the target in the wrong place.

.github/workflows/release.yaml
greyscaled: interestingly enough it was like that befoe, and so I made this swtich to see if it would fix things.

I'll revert back tho.

tychoish: sorry, we should, I think leave the `target:` key and _not_ pass anything to the `--release` flag, which doesn't appear to take any options.

greyscaled: Yep sorry, to clarify:

- I see/understand the problem
- I (possibly wrongfully) believe I already tried doing that somewhere in this branch
- https://github.com/GlareDB/glaredb/commit/bb88f76e87486e09b17bc8a74a92a003181b2c1b introduced it which was a "fix" and so I got spooked

Regardless, I added a new commit to test this out: https://github.com/GlareDB/glaredb/pull/2564/commits/d54d21ee32264cea3150770be878116f1239462b

(fingers crossed).

If that commit fails, I'll make sure to comment out the other jobs (which I forgot to), so as to reduce noise and isolate the macos run

greyscaled: I looked through the commits, and what you're suggesting is what I had intended in https://github.com/GlareDB/glaredb/pull/2564/commits/70221471cf2a6a023f398f0c829ea805926ab0c7 but forgot to delete the part after `--release`. My apologies!

.github/workflows/release.yaml
```diff
@@ -8,13 +8,45 @@ on:
   push:
     tags:
       - "*"
+    branches:
+      - tycho/release-atomically
```
greyscaled, line 12:
```suggestion
```

greyscaled (approved):

Issue #2561, opened 2024-02-01
Does not work with a Parquet file on an HTTPs server


alexey-milovidov:
## Description

```
Successfully downloaded the GlareDB binary.
You can run it as:
    > ./glaredb
ubuntu@ip-172-31-38-111:~$ ./glaredb
GlareDB (v0.8.3)
Type \help for help.
> \help
\help           Show this help text
\mode MODE      Set the output mode [table, json, ndjson, csv]
\max-rows NUM   Max number of rows to display
\max-width NUM  Maximum width of the output table to display. Defaults to terminal size.
\open PATH      Open a database at the given path
\timing         Toggle query execution runtime display
\quit           Quit this session
> SELECT COUNT(*) FROM 'https://clickhouse-public-datasets.s3.eu-central-1.amazonaws.com/hits_compatible/athena_partitioned/hits_{0..99}.parquet';
Error: External error: Error during planning: Unexpected status code '404 Not Found' for url: 'https://clickhouse-public-datasets.s3.eu-central-1.amazonaws.com/hits_compatible/athena_partitioned/hits_%7B0..99%7D.parquet'
> SELECT COUNT(*) FROM 'https://clickhouse-public-datasets.s3.eu-central-1.amazonaws.com/hits_compatible/athena/hits.parquet';
Error: External error: Execution error: file size of 0 is less than footer
> 
ubuntu@ip-172-31-38-111:~$ curl -I https://clickhouse-public-datasets.s3.eu-central-1.amazonaws.com/hits_compatible/athena/hits.parquet
HTTP/1.1 200 OK
x-amz-id-2: rQTq/S1trym26TzeNCTii4QU80RuvPQcST4lEVnu4gmTet9aLbGUunMSRAsxeug/kSCPR9HUDOc=
x-amz-request-id: PPB9PJMPEFEK8H20
Date: Thu, 01 Feb 2024 23:46:16 GMT
Last-Modified: Sun, 03 Jul 2022 03:31:13 GMT
ETag: "359976a01ddb98f2363d076a276acb24-1762"
Accept-Ranges: bytes
Content-Type: binary/octet-stream
Server: AmazonS3
Content-Length: 14779976446
```

PR #2565, opened 2024-02-02
fix: parquet over http 

universalmind303:
reqwest's `content_length` method apparantly doesnt check for the casing of "Content-Length". This seems like a bug in reqwest to me. 

https://github.com/seanmonstar/reqwest/issues/843

```rust

#[tokio::main]
async fn main() {
    let client = reqwest::Client::new();
    let head_res = client.head("https://clickhouse-public-datasets.s3.eu-central-1.amazonaws.com/hits_compatible/athena/hits.parquet").send().await.unwrap();
    let content_length = head_res.content_length();
    assert_eq!(content_length, Some(14779976446))
}

```

closes https://github.com/GlareDB/glaredb/issues/2561

### DIFF
--- a/crates/datasources/src/object_store/http.rs
+++ b/crates/datasources/src/object_store/http.rs
@@ -91,9 +91,11 @@ impl ObjStoreAccess for HttpStoreAccess {
                 status, self.url
             )));
         }
+        // reqwest doesn't check the content length header, instead looks at the contents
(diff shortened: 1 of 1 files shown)


tychoish (approved):
regression testable?

universalmind303:
> regression testable?

maybe?? Without relying on other apis like the one mentioned in the issue, We'd have to spin up some infra for hosting some public files.

PR #2567, opened 2024-02-02
chore(cloudauth): remove unused fields

greyscaled:
These are harmless to have around in the tests, but don't add value and may mislead someone in the future into expecting they exist on the actual response

### DIFF
--- a/crates/proxyutil/src/cloudauth.rs
+++ b/crates/proxyutil/src/cloudauth.rs
@@ -173,7 +173,7 @@ mod tests {
               "ip": "1.2.3.4",
               "port": "5432",
               "memory_limit_bytes": 268435456,
-              "gcs_storage_bucket": "",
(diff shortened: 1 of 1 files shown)


tychoish (approved):
these are both in tests, so might be fine to leave them in, just to make sure that the proxy doesn't break if cloud gives it information it isn't expecting, but also we could delete them or add a comment.

greyscaled:
> these are both in tests, so might be fine to leave them in, just to make sure that the proxy doesn't break if cloud gives it information it isn't expecting, but also we could delete them or add a comment.

I'd be down to keep them with a comment for the reason of not requiring all fields in the response being mapped

tychoish:
> I'd be down to keep them with a comment for the reason of not requiring all fields in the response being mapped

Whatever you like, I think it's probably fine either way.

PR #2568, opened 2024-02-02
feat: SQLite data source

vrongmeal:
### TODO:

- [x] `CREATE EXTERNAL TABLE`
- [x] `CREATE EXTERNAL DATABASE`
- [x] Virtual listing
- [x] Tests
  - [x] Fix filter pushdown for Sqlite
- [x] Support dates/times/timestamps

## Followups:
* #2643


crates/datasources/src/sqlite/async_sqlite.rs
```diff
@@ -0,0 +1,232 @@
...
+use tokio::task::JoinHandle;
+
+use crate::sqlite::errors::{Result, SqliteError};
+
+#[derive(Clone, Debug)]
+pub struct SqliteAsyncClient {
+    path: PathBuf,
+}
+
+impl SqliteAsyncClient {
```
tychoish, line 20: This client is synchronous, though?

crates/datasources/src/sqlite/convert.rs
```diff
@@ -0,0 +1,217 @@
...
+                                buf.clear();
+                                encode::encode_binary(&mut buf, b)?;
+                                builder.append_value(&buf);
+                            }
+                        }
+                    }
+                    Arc::new(builder.finish())
+                }
+                DataType::Binary => {
+                    let mut builder = BinaryBuilder::with_capacity(data.len(), 10 * data.len());
```
tychoish, line 182: why 10x?

crates/datasources/src/sqlite/mod.rs
```diff
@@ -0,0 +1,426 @@
...
+
+            while let Some(batch) = stream.next().await {
+                let data = match batch {
+                    Ok(b) => b.data,
+                    Err(e) => {
+                        yield Err(DataFusionError::Execution(e.to_string()));
+                        return;
+                    }
+                };
+                match conv.create_record_batch(data) {
```
tychoish, line 298: how big are the batches we're getting? does it matter?

crates/sqlbuiltins/src/functions/table/sqlite.rs
```diff
@@ -0,0 +1,67 @@
...
+                let access = SqliteAccess {
+                    db: location.into(),
+                };
+                let state = access.connect();
+
+                let provider =
+                    SqliteTableProvider::try_new(state, table).map_err(ExtensionError::access)?;
+                Ok(Arc::new(provider))
+            }
+            _ => Err(ExtensionError::InvalidNumArgs),
```
tychoish, line 64: I thought that we might use object store and be able to pull down entire files of less than... say a gig...

crates/datasources/src/sqlite/async_sqlite.rs
vrongmeal: More like "async friendly".

crates/datasources/src/sqlite/convert.rs
vrongmeal: Just an estimate we've used. Assuming average length of 1 blob to be 10 bytes.

crates/datasources/src/sqlite/mod.rs
vrongmeal: each batch is at-most 1000 rows

crates/sqlbuiltins/src/functions/table/sqlite.rs
vrongmeal: Yeah, we can do that later on. Will add it to the todos.

crates/datasources/src/sqlite/async_sqlite.rs
tychoish: seems like we could make it actually async, right?

vrongmeal: What do you mean by "actually" async? Sqlite has no concept of async, I don't think we can do much about it apart from coloring the function as "async"

tychoish: oh just that the block-in-place could maybe go away, but that only really matters if the consumers of this are async. I think this is fine as is.

crates/datasources/src/sqlite/mod.rs
tychoish: I'd add a comment just to make that clear to would-be-users.

crates/datasources/src/sqlite/async_sqlite.rs
```diff
@@ -0,0 +1,232 @@
...
+
+    pub fn execute(&self, s: impl Into<String>, p: impl Params) -> Result<usize> {
+        let s = s.into();
+        let path = self.path.clone();
+
+        tokio::task::block_in_place(move || {
+            let conn = Connection::open(path)?;
+            let n = conn.execute(&s, p)?;
+            Ok(n)
+        })
```
scsmithr, line 33: Block in place doesn't seem needed.

crates/datasources/src/sqlite/async_sqlite.rs
```diff
@@ -0,0 +1,232 @@
...
+        s: impl Into<String>,
+        p: impl Params + Send + 'static,
+    ) -> SqliteQueryStream {
+        let s = s.into();
+
+        let (req_tx, mut req_rx) = mpsc::channel(1);
+        let (resp_tx, resp_rx) = mpsc::channel(1);
+
+        let path = self.path.clone();
+        let handle: JoinHandle<Result<()>> = tokio::task::spawn_blocking(move || {
```
scsmithr, line 47: This should probably just be `spawn` to allow for non-blocking recvs on the channel. Otherwise this will literally block a thread.

crates/datasources/src/sqlite/async_sqlite.rs
vrongmeal: Why not?

crates/datasources/src/sqlite/async_sqlite.rs
```diff
@@ -0,0 +1,232 @@
...
+                (0..num_cols)
+                    .map(|idx| {
+                        let v = r.get_ref(idx)?;
+                        Ok(Value::from(v))
+                    })
+                    .collect::<Result<Vec<_>, rusqlite::Error>>()
+            });
+
+            while let Some(()) = req_rx.blocking_recv() {
+                eprintln!("processing a batch");
```
vrongmeal, line 70: TODO: Remove this.

crates/datasources/src/sqlite/mod.rs
vrongmeal: We have:

```
const BATCH_SIZE: usize = 1000;
```

in `async_sqlite.rs`

crates/datasources/src/sqlite/wrapper.rs
```diff
@@ -0,0 +1,184 @@
...
+    }
+}
+
+#[derive(Debug, Clone)]
+pub struct Column {
+    pub name: String,
+}
+
+#[derive(Debug, Clone)]
+pub struct SqliteBatch {
```
tychoish, line 152: I'd be inclined to avoid writing to this structure and just construct/build record batches directly.

tychoish (dismissed):

tychoish (commented):
sorry, didn't mean to hit approve yet. it is good though.

.github/workflows/ci.yaml
```diff
@@ -440,6 +440,7 @@ jobs:
             --exclude 'sqllogictests_snowflake/*' \
             --exclude 'sqllogictests_cassandra/*' \
             --exclude 'sqllogictests_clickhouse/*' \
+            --exclude 'sqllogictests_sqlite/*' \
```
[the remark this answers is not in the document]
vrongmeal: > I'm inclined to think of sqlite less like "a database" and more like "a file format."

I have to disagree with this. SQLite is a database.

scripts/create-test-sqlite-db.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Check and delete the old sqlite db (if it exists)
+# This is just to ensure cleanup locally.
+test -f "${SQLITE_DB_LOCATION}" && rm "${SQLITE_DB_LOCATION}"
```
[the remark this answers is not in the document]
vrongmeal: I think we should have a reproducible script for as many things as possible (to make any changes required in the future). It's similar to whether we should have a `Dockerfile` or simply an image.

crates/datasources/src/sqlite/wrapper.rs
```diff
@@ -0,0 +1,184 @@
...
+        Ok(Self { path, inner })
+    }
+
+    pub fn query(&self, s: impl Into<String>) -> SqliteQueryStream {
+        let s = s.into();
+
+        let (tx, rx) = mpsc::channel(1);
+
+        let client = self.inner.clone();
+        let handle = tokio::spawn(async move {
```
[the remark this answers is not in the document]
vrongmeal: Can't really do this because the future is never polled (`.await`ed), i.e., the set of instructions inside `async {}` block are never executed. `spawn`ing here is necessary since the future inside spawn is polled at-least once.

We could go around it by polling the future once but it just makes the code way more complicated and brings nothing to value here, also, we never poll it again and since rust's futures are lazy, we can have a dangling future in memory. Moreover, spawning here makes sense.

crates/datasources/src/sqlite/wrapper.rs
```diff
@@ -0,0 +1,184 @@
...
+                                break;
+                            }
+                            Ok(data) => Ok(SqliteBatch {
+                                cols: cols.clone(),
+                                data,
+                            }),
+                            Err(e) => Err(SqliteError::from(e)),
+                        };
+
+                        if tx.blocking_send(batch).is_err() {
```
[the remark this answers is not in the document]
vrongmeal: `async-sqlite` spawns a dedicated OS thread to execute sqlite, so we're essentially blocking on an OS thread. What do you mean by "back pressure" here? It's simply a `futures::block_on` on `tx.send`. Can't think of any consequences.

crates/datasources/src/sqlite/wrapper.rs
```diff
@@ -0,0 +1,184 @@
...
+
+    pub fn query(&self, s: impl Into<String>) -> SqliteQueryStream {
+        let s = s.into();
+
+        let (tx, rx) = mpsc::channel(1);
+
+        let client = self.inner.clone();
+        let handle = tokio::spawn(async move {
+            client
+                .conn(move |conn| {
```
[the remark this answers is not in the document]
vrongmeal: what `query_all` code?

crates/datasources/src/sqlite/wrapper.rs
vrongmeal: Sure, we can do that! Not very keen on doing it for `query_all` function though. Makes it hard to read values from record batch (for eg, when trying to query schema)

vrongmeal: On a second thought, this does make the code very hard to convert.

We will need an intermediate vector of rows and this serves as that. We'll have to maintain all fields as a vector of array builders (which do not implement any common trait).

```
// No such trait as an ArrayBuilder...
let fields: Vec<_> = schema.fields().map(|f| Arc::new(ArrayBuilder::from(f.data_type())).collect();
```

Even if they do (we could `Any` here for the reference):

```
let fields: Vec<Arc<dyn Any>> = schema.fields().map(|f| Arc::new(f.data_type().array_builder())).collect();
```

We'd have to downcast for each column. Not sure if it's worth it.

I'll push the changes in a commit, let's see if it feels OK.

tychoish: we did this (roughly) with the bson converter, maintained a vector of arrays for each column, and while you need an array for the _current_ column, you don't need _all_ of the rows. 

This avoids ever (basically) having more than (a record batch size)-number of rows in memory at the time.

crates/datasources/src/sqlite/wrapper.rs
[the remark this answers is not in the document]
tychoish: it seemed like there was some duplication between the methods

crates/datasources/src/sqlite/wrapper.rs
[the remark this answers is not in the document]
tychoish: My thought was that this method returns a stream, which means that we don't need to work until that stream is polled, so it's fine that it's lazy, and in fact maybe preferable if we don't do any work until the stream is polled. (though this gets weird for the error handling. 

If you have a `futures::stream::Once` that returns a `Stream` object, you can call `flatten()` on that stream object (via `StreamExt`) and then you have it... no background threads. (I think)

you do need the background thread if you want to collect _all_ of the results without the stream, but you could also just iterate the stream and save the code.

.github/workflows/ci.yaml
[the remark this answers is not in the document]
tychoish: we discussed this offline, and my intention here was to draw a distinction between things that required docker containers and external service dependencies, and things that can be done with just modifying files, and avoiding less complexity in the test configuration and setup. 

this is probably fine to commit as is but we should definitely reconfigure in the future.

crates/datasources/src/sqlite/wrapper.rs
vrongmeal: ohh okay! so i never looked it up but there is such a thing as `ArrayBuilder`! Should be easy then

crates/datasources/src/sqlite/wrapper.rs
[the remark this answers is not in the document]
vrongmeal: Not really, they feel similar but aren't. Abstracting more will look more messy

crates/datasources/src/sqlite/wrapper.rs
vrongmeal: Made this change.

crates/datasources/src/sqlite/wrapper.rs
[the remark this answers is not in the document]
vrongmeal: Completely different code path now...

crates/datasources/src/sqlite/wrapper.rs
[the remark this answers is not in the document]
vrongmeal: Did it without spawning. Did not use stream but should do it!

scripts/create-test-sqlite-db.sh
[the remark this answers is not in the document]
vrongmeal: Added a hook that creates the test file and creates if it doesn't exist.

tychoish (approved):

PR #2593, opened 2024-02-02
fix: bson timestamp type casting

### LINKED ISSUE #2585: panic when copying to bson
## Description
```sh
mkdir testdata/bson

glaredb -q "COPY (select * from 'testdata/parquet/userdata1.parquet') to 'testdata/bson/userdata1.bson';"
thread 'local-thread-2' panicked at /Users/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/arrow-array-49.0.0/src/cast.rs:751:33:
primitive array
Error: Join Error
caused by
External error: task 53 panicked

Caused by:
    0: External error: task 53 panicked
    1: task 53 panicked
```

### LINKED ISSUE #2590: bson roundtrip doesn't work with timestamp fields
## Description
```sql
> create table timestamps (t timestamp);
Table created
> insert into timestamps values (arrow_cast(946684860000000, 'Timestamp(Microsecond, None)'));
Inserted 1 row
> select * from timestamps;
┌─────────────────────┐
│ t                   │
│ ──                  │
│ Timestamp<µs, UTC>  │
╞═════════════════════╡
│ 2000-01-01T00:01:00 │
└─────────────────────┘
> copy timestamps to  '/Users/me/Development/glaredb/out.bson';
Copy success
> select * from '/Users/me/Development/glaredb/out.bson';
Error: External error: Execution error: Unexpected datatype for builder Date64
```

tychoish:
Closes #2590
Closes #2585

### DIFF
--- a/testdata/sqllogictests_object_store/local/bson.slt
+++ b/testdata/sqllogictests_object_store/local/bson.slt
@@ -1,0 +1,33 @@
+statement ok
+COPY (select * from '${PWD}/testdata/parquet/userdata1.parquet') to '${TMP}/userdata1.bson';
+
+query I
+SELECT count(*) FROM '${TMP}/userdata1.bson';
+----
+1000
+
+statement ok
+create table timestamps (t timestamp);
+
+statement ok
+insert into timestamps values (arrow_cast(946684860000000, 'Timestamp(Microsecond, None)'));
+
+query
+select * from timestamps;
+----
+2000-01-01 00:01:00
+
+statement ok
+copy timestamps to '${TMP}/timestamp_test_out.bson';
+
+# bson's date format is an int64, of milliseconds since unix epoch;
+# which most closely maps to arrow's Date64, and since that's the only
+# date format, (the bson "timestamp" type is really a mongodb
+# implementation detail, generally regarded to be a mistake.) all
+# arrow timetstamp and date types are upcast or truncated into this
+# type. (Time64 and Time32 just go to integers)
+
+query
+select * from '${TMP}/timestamp_test_out.bson';
+----
+946684860000
--- a/crates/datasources/src/bson/builder.rs
+++ b/crates/datasources/src/bson/builder.rs
@@ -23,6 +23,7 @@ use datafusion::arrow::array::{
     StructArray,
     TimestampMicrosecondBuilder,
     TimestampMillisecondBuilder,
+    TimestampNanosecondBuilder,
     TimestampSecondBuilder,
 };
 use datafusion::arrow::datatypes::{DataType, Field, Fields, TimeUnit};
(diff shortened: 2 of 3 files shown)


testdata/sqllogictests_object_store/local/bson.slt
```diff
@@ -0,0 +1,33 @@
...
+
+query
+select * from timestamps;
+----
+2000-01-01 00:01:00
+
+statement ok
+copy timestamps to '${TMP}/timestamp_test_out.bson';
+
+# bson's date format is an int64, of milliseconds since unix epoch;
```
universalmind303, line 23: isn't there the 
> "UTC datetime - The int64 is UTC milliseconds since the Unix epoch." 

which isn't part of the internal mongodb spec


https://bsonspec.org/spec.html

tychoish: yes, exactly, that's what we're using. We just convert all arrow `Date` and `Timestamp` values into these int64 (millis since epoch), but when we read bson data back into GlareDB all of those values become `Date64`; not back to timestamps. 

```sql
> create table timestamps (t timestamp);
Table created
> insert into timestamps values (arrow_cast(946684860000000, 'Timestamp(Microsecond, None)'));
Inserted 1 row
>
>
> select * from timestamps;
┌─────────────────────┐
│ t                   │
│ ──                  │
│ Timestamp<µs, UTC>  │
╞═════════════════════╡
│ 2000-01-01T00:01:00 │
└─────────────────────┘
> copy timestamps to 'timestamps.bson';
Copy success
> select * from timestamps;
┌─────────────────────┐
│ t                   │
│ ──                  │
│ Timestamp<µs, UTC>  │
╞═════════════════════╡
│ 2000-01-01T00:01:00 │
└─────────────────────┘
> select * from 'timestamps.bson';
┌─────────────────────┐
│ t                   │
│ ──                  │
│ Date64              │
╞═════════════════════╡
│ 2000-01-01T00:01:00 │
└─────────────────────┘
```

It works as intended but `sqllogictest` renders `Date64` as `int64`s.

When we add support for declaring a schema rather than inferring it, we'll be able to cast these bson dates to whatever we need (with only the expected loss of fidelity and resolution), and the code _mostly_ supports this today.

universalmind303 (approved):
Id like to keep https://github.com/GlareDB/glaredb/issues/2590 open as we still technically can't roundtrip those types. Since we aren't notifying the user, or have it documented anywhere that you can't roundtrip these types, I think the issue is still valid.

tychoish:
> Id like to keep https://github.com/GlareDB/glaredb/issues/2590 open as we still technically can't roundtrip those types. Since we aren't notifying the user, or have it documented anywhere that you can't roundtrip these types, I think the issue is still valid. 

I understand the desire for precision, but the same issue impacts JSON and CSV export (only in even more cases, as they have fewer types,) and more don't round trip. Parquet, even, has a less sophisticated type system than arrow (and therefore glaredb) and would probably lose some type fidelity in the round trip. We can keep a general issue or epic open for "round trip types" as a placeholder for these kinds of issues, but the solutions are all a bit imperfect, so I doubt we'd ever be able to truly resolve it. Having a placeholder might be useful, or it might just confuse things even more.

There's also nothing actionable in the glaredb code base to do for this, about BSON right now: we can't create new types in the upstream format. We could document this, but we should track that in a documentation issue. This would be a good page to have at some point. 

We could consider implementing or providing a "strict" mode (or more likely various strict-modes), with regards to type conversions where we could propagate errors for any non-roundtrip-able types, but that's a different issue, and would impact most data sources. I think that's a different issue.

Being able to declare the schema rather than inferring it from the external type-system, provides us the opportunity to up-cast values, and perhaps change some of the down-casting semantics to support something that more closely approximates round tripping. This is something that we'd like to do and can work on, in the near term, and I think has the greatest effort-to-impact ratio, but it's also a different issue.

PR #2594, opened 2024-02-03
fix: local file path resolution for json and bson 

### LINKED ISSUE #2589: bson doesn't work with relative paths
## Description
```sql
 select * from 'example.bson';
Error: External error: Error parsing Path "./example.bson": Encountered illegal character sequence "." whilst parsing path segment "."
```

```sql
select * from '/Users/me/Development/glaredb/example.bson';
┌───────┬─────────┬───────┐
│    id │ name    │   age │
│    ── │ ──      │    ── │
│ Int32 │ Utf8    │ Int32 │
╞═══════╪═════════╪═══════╡
│     1 │ Alice   │    30 │
│     2 │ Bob     │    25 │
│     3 │ Charlie │    35 │
└───────┴─────────┴───────┘
```

tychoish:
On top of #2593

Resolves #2589

This does not, I think, fix the globbing issue as I had hoped, but it does fix local path canonicalization and tilde expansion.


vrongmeal (approved):
Looks good!

PR #2599, opened 2024-02-05
chore: df 35 upgrade

universalmind303:


tychoish (commented):
Generally this is fine, have two questions: 
- is there anything that you're worried about that the tests wouldn't cover?
- should we tag the release before this is merged or do you think this is going to be safe to release without further testing?

universalmind303:
> Generally this is fine, have two questions:
> 
>     * is there anything that you're worried about that the tests wouldn't cover?
> 
>     * should we tag the release before this is merged or do you think this is going to be safe to release without further testing?

no i think this is fine, the impact surface was much smaller than our last upgrade, and i'd be suprised if our slts missed anything major

tychoish (approved):

PR #2600, opened 2024-02-05
feat: add object store table support for bson format data

tychoish:
We maybe could also support lance in a similar sort of way, (which is
where I started when I was looking at this code, but I got this.)

Need to coordinate credentials to get some data uploaded to a gcs
bucet to test this.


vrongmeal (approved):
This looks good to me.

***

> I looked into the FileType enum and it being a dependent on FileFormat trait. I think we should implement FileFormat for these formats. We can keep the file_type method as unreachable since that code-path is never used by us.

From slack ^

crates/sqlexec/src/dispatch/external.rs
```diff
@@ -592,27 +592,47 @@ impl<'a> ExternalDispatcher<'a> {
...
-        let accessor = ObjStoreAccessor::new(access)?;
-        let objects = accessor.list_globbed(path).await?;
-
-        let state = self.df_ctx.state();
-        let provider = accessor.into_table_provider(&state, ft, objects).await?;
+        let accessor = ObjStoreAccessor::new(access.clone())?;
 
-        Ok(provider)
+        match file_type {
+            "csv" => Ok(accessor
```
scsmithr, line 598: Apparently the file type is getting stored with with quotes, so matching on `"\"csv\""` works...

tychoish: I believe it's not getting stored as a quote but just printed into a quoted form when it's read which is easy enough clean up.

PR #2607, opened 2024-02-07
fix: fixedsizelist in delta

universalmind303:
fixes the issue of delta converting fsl -> list when writing delta tables. 

This PR adds some metadata containing the original field if it was downcast when writing to delta, then compares those fields. If they are different, it applies a projection from the delta field to the original field. 

If we need further performance, we'll likely have to read and write directly instead of using delta's arrow/datafusion logic as they have [no intention](https://github.com/delta-io/delta-rs/issues/2162) on supporting FixedSizeList.


testdata/sqllogictests/datatypes.slt
```diff
@@ -0,0 +1,10 @@
+# fixed size list round trip
+
+statement ok
+create table test as select arrow_cast([1,2], 'FixedSizeList(2, Int64)') as f;
```
tychoish, line 4: can you test:

- inserting into the table
- doing a `select *` query and getting the results
- doing some operation on the data (e.g. summing a few rows).

crates/datasources/src/native/access.rs
```diff
@@ -85,6 +88,48 @@ impl LogStoreFactory for FakeStoreFactory {
...
+
+// Some datatypes get downgraded to a different type when they are stored in delta-lake.
+// So we add some metadata to the field to indicate that it needs to be converted back to the original type.
+fn arrow_to_delta_safe(arrow_type: &DataType) -> DeltaResult<DeltaField> {
+    match arrow_type {
+        dtype @ DataType::Timestamp(_, tz) => {
+            let delta_type =
+                (&DataType::Timestamp(TimeUnit::Microsecond, tz.clone())).try_into()?;
+            let mut metadata = HashMap::new();
+            metadata.insert("arrow_type".to_string(), json!(dtype));
```
tychoish, line 104: - what if the metadata isn't present (presumably our catalog and delta have different ideas of the types?) is this an error? should we backfill or (in theory) migrate user data?
- I don't know delta well enough, but does this meta data impact anything with delta, or impact our compatibility with delta?

crates/datasources/Cargo.toml
```diff
@@ -45,7 +45,7 @@ reqwest = { workspace = true }
 rust_decimal = { version = "1.34.2", features = ["db-tokio-postgres"] }
 serde = { workspace = true }
 serde_with = "3.6.0"
-serde_json = { workspace = true }
+serde_json.workspace = true
```
tychoish, line 48: the other syntax seems to be prevalent in this file.

tychoish (changes requested):

crates/datasources/Cargo.toml
universalmind303: seems like we have a mix of both. The latter is what's used by cargo. 

`cargo install -p datasources serde_json`

tychoish (approved):

PR #2612, opened 2024-02-08
fix: add pg_rewrite and pg_depend tables

talagluck:
These are to make another step toward fixing #2436. 

I built the definitions off of https://www.postgresql.org/docs/16/catalog-pg-depend.html and https://www.postgresql.org/docs/16/catalog-pg-rewrite.html, but please double-check that I did this correctly.

### DIFF
--- a/crates/sqlbuiltins/src/builtins.rs
+++ b/crates/sqlbuiltins/src/builtins.rs
@@ -704,6 +704,35 @@ pub static PG_MATVIEWS: Lazy<BuiltinView> = Lazy::new(|| BuiltinView {
     ",
 });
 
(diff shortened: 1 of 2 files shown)


crates/sqlbuiltins/src/builtins.rs
```diff
@@ -721,6 +750,8 @@ impl BuiltinView {
             &PG_VIEWS,
             &PG_TYPE,
             &PG_MATVIEWS,
+            &PG_REWRITE,
+            &PG_DEPEND
```
greyscaled, line 754:
```suggestion
            &PG_DEPEND,
```

universalmind303 (approved):
Looks good. Thanks @talagluck

PR #2613, opened 2024-02-08
feat: lance insert_into support

tychoish:
on top of #2605


crates/datasources/src/lance/mod.rs
```diff
@@ -1,11 +1,107 @@
...
+        input: Arc<dyn ExecutionPlan>,
+        _overwrite: bool,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let mut stream = execute_stream(input, state.task_ctx())?.chunks(32);
+        let mut ds = self.dataset.clone();
+        let schema = self.schema().clone();
+
+        Ok(Arc::new(StreamExecPlan::new(
+            schema.clone(),
+            Mutex::new(Some(Box::pin(RecordBatchStreamAdapter::new(
```
universalmind303, line 82: we should make an issue to fix this `mutex<option<>>` thing. This keeps catching my eye as a huge red flag.

universalmind303 (changes requested):
Overall looks good, just a few small changes, and some tests.

crates/datasources/src/lance/mod.rs
```diff
@@ -1,11 +1,107 @@
...
+            .await?)
+    }
+
+    async fn insert_into(
+        &self,
+        state: &SessionState,
+        input: Arc<dyn ExecutionPlan>,
+        _overwrite: bool,
+    ) -> Result<Arc<dyn ExecutionPlan>> {
+        let mut stream = execute_stream(input, state.task_ctx())?.chunks(32);
```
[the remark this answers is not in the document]
tychoish: we get a stream of record batches, lance works on (for now) an iterator of record batches, this let's us deal with groups for lance to be able to do it's thing, without need to pass iterators of one. I'm open to anything else though.

crates/datasources/src/lance/mod.rs
universalmind303: I really don't think we should be using a generic ExecutionPlan here at all. I think creating custom `LanceInsertExec` and `MongoInsertExec`s would be much better. It gives us more flexibility later on with the physical planning, and it completely avoids this whole `Mutex<Option<>>` hack. 


```rust
#[derive(Debug)]
struct LanceInsertExec {
    input: Arc<dyn ExecutionPlan>,
    dataset: Dataset,
}

impl DisplayAs for LanceInsertExec {
    fn fmt_as(
        &self,
        _t: datafusion::physical_plan::DisplayFormatType,
        f: &mut std::fmt::Formatter,
    ) -> std::fmt::Result {
        todo!()
    }
}

impl ExecutionPlan for LanceInsertExec {
    fn as_any(&self) -> &dyn Any {
        self
    }

    fn schema(&self) -> SchemaRef {
        Arc::new(Schema::new(vec![Field::new(
            "count",
            DataType::UInt64,
            false,
        )]))
    }
    fn maintains_input_order(&self) -> Vec<bool> {
        vec![false]
    }
    fn output_partitioning(&self) -> datafusion::physical_plan::Partitioning {
        Partitioning::UnknownPartitioning(1)
    }

    fn output_ordering(&self) -> Option<&[datafusion::physical_expr::PhysicalSortExpr]> {
        None
    }

    fn children(&self) -> Vec<Arc<dyn ExecutionPlan>> {
        vec![self.input.clone()]
    }

    fn with_new_children(
        self: Arc<Self>,
        children: Vec<Arc<dyn ExecutionPlan>>,
    ) -> Result<Arc<dyn ExecutionPlan>> {
        Ok(Arc::new(Self {
            input: children[0].clone(),
            dataset: self.dataset.clone(),
        }))
    }
    fn execute(
        &self,
        partition: usize,
        context: Arc<datafusion::execution::TaskContext>,
    ) -> Result<datafusion::physical_plan::SendableRecordBatchStream> {
        let mut stream = execute_stream(self.input.clone(), context)?.chunks(32);
        let mut ds = self.dataset.clone();
        let schema = self.schema().clone();
        let write_opts = WriteParams {
            mode: WriteMode::Append,
            ..Default::default()
        };

        let mut count: u64 = 0;
        let output = futures::stream::once(async move {
            while let Some(batches) = stream.next().await {
                let start = ds.count_rows().await?;
                let rbi = RecordBatchIterator::new(
                    batches
                        .into_iter()
                        .map(|v| v.map_err(|dfe| ArrowError::ExternalError(Box::new(dfe)))),
                    schema.clone(),
                );
                ds.append(rbi, Some(write_opts.clone())).await?;
                count += (ds.count_rows().await? - start) as u64;
            }
            Ok::<RecordBatch, DataFusionError>(create_count_record_batch(count))
        });
        Ok(Box::pin(RecordBatchStreamAdapter::new(
            self.schema(),
            output,
        )))
    }
}


//...
    async fn insert_into(
        &self,
        state: &SessionState,
        input: Arc<dyn ExecutionPlan>,
        _overwrite: bool,
    ) -> Result<Arc<dyn ExecutionPlan>> {
        Ok(Arc::new(LanceInsertExec {
            input,
            dataset: self.dataset.clone(),
        }))
    }
```

universalmind303: This is also what we are already doing for delta.

tychoish: We don't support inserts into delta unless it's in native/managed storage. We also construct a lot of bespoke plans 

There aren't really optimizations to be done on straight insert paths (maybe for updates, but not here), once you have the stream of documents to insert. We definitely have a bunch of poorly written inserts, particularly where we process the insert data before handing it off to the exec (which we do a lot in cases where we do have one-off-exec plans for inserts,) so I'm disinclined to think that the way that we structure the code in this way really means anything for the optimization potential. 

If there's a functional benefit here or I've missed something, though, I'm definitely open to it.

universalmind303: I'm not necessarily referring to optimizations, but localization or other restrictions we may eventually impose on plans. I don't think there's going to be a lot of optimizations, but I could very well see a scenario where we'd want to run a `MongoInsertExec` remotely, and a `LanceInsertExec` locally. 

I'd also forsee a future where we'd need to run the `MongoInsertExec` on a specific instance due to regulations or other restrictions. Putting everything into an opaque `Exec` really limits us in this regard. 

I personally don't think there's anything wrong with creating custom exec nodes for each datasource. It's a bit more verbose, but it's ultimately more flexible later on when we inevitably need to refactor.

tychoish: I think the hypothetical refactorings you suggest carry a lot of assumptions about the future operations of the system that are not particularly clear or certain to me at this point. 

If we're doing the kind of refactoring that you suggest, I think writing out per-implementation `ExecutionPlan` implementation would not be onerous, and given that none of the requirements for these hypothetical plans are known at this time, I don't know we gain anything by doing this work right now, other than writing more code that's isomorphic with the current implementation that we'll just have to change later.

universalmind303: The refactoring is just one scenario that it would be beneficial. If we ignore that, It seems that individual execs still provide the most flexibility, makes the intent very clear, and are pretty easy to implement. 

Most importantly, they don't require the hacky `Mutex<Option<>>` because they dont pass around a stream.


This seems to me like a very low effort change _(10-20 minutes tops)_. 

It seems like there's a strong attachment to this current implementation that I'm not understanding?

tychoish: the current implementation works fine, is easier (in my estimation) to read, and reduces the amount of boilerplate... and on the other hand we can have more boilerplate in service of a future refactoring in service of speculative use-cases? That's just a bad deal. 

We should, use shared implementations for most exec plans like this, and when we have business requirements that mean that we need to do special handling for a data source, it's easy enough to refactor them out. Boilerplate is a liability.

universalmind303 (approved):

PR #2615, opened 2024-02-08
WIP: sentence transformers & similarity functions

universalmind303:
While adding in some sentence transformers functionality, I realized that there is quite a bit of startup cost for each transformer. This is because the model needs to be downloaded from huggingface. This is a usually a onetime penalty as it's cached locally afterwards. But nonetheless... 

I didn't really want everyone to have to pay this penalty if they have no intention on using this functionality. We could lazy load it, but then it's going to add a massive delay the first time the function is called. 

The current sentence_transformer model is relatively small, but others are quite massive (10gb+). So it made sense to have these as an opt-in feature. 

So this puts all of this functionality behind an opt in extension `sentence_transformers` that is loaded similarly to how one would via duckdb

```sql
load sentence_transformers;
```

This is mainly broken up into two parts.

1. Add new syntax and functionality for `load` operation
2. Add scalar functions (`sentence_transformers.*` and `cosine_similarity`)

I want to merge the `FUNCTION_REGISTRY` functionality into the `SessionCatalog` as we currently have two sources of truth for functions & it's quite difficult to keep them in sync. Especially when trying to mutate them over rpc.

---

Incidentally, this also fixes the issue of calling udfs over rpc.


Cargo.toml
```diff
@@ -47,3 +47,7 @@ url = "2.5.0"
 git = "https://github.com/delta-io/delta-rs.git"
 rev = "993e2c202936719855f8831513bcbab1b9930b94"
 features = ["s3", "gcs", "azure", "datafusion"]
+
+
+[patch.crates-io]
```
universalmind303, line 52: todo: revert this patch

crates/catalog/src/session_catalog.rs
```diff
@@ -233,6 +232,30 @@ impl SessionCatalog {
...
+        match ent {
+            CatalogEntry::Function(function)
+                if function.meta.builtin && function.func_type == FunctionType::Scalar =>
+            {
+                Some(function.clone())
+            }
+            _ => None,
+        }
+    }
+
```
universalmind303, line 258: todo: remove this function.

crates/sqlexec/src/planner/context_builder.rs
```diff
@@ -299,8 +299,12 @@ impl<'a> AsyncContextProvider for PartialContextProvider<'a> {
         name: &str,
         args: &[Expr],
     ) -> DataFusionResult<Option<Expr>> {
+        println!("get_function_meta: name: {}", name);
+        // this ensures that it is in both the session catalog and the function registry
+        // We should eventually remove the FUNCTION_REGISTRY and use the session catalog directly
```
universalmind303, line 304: todo: remove

universalmind303:
This does lay down some foundation for a lot of future opportunities via lazily loaded extensions. 

Ideally we'd eventually be able to load external extensions over FFI, but right now this is just scoped to internal extensions that act as a way to selectively load specific features with high startup cost.

crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -34,9 +42,57 @@ use table::{BuiltinTableFuncs, TableFunc};
 
 use self::alias_map::AliasMap;
 use crate::functions::scalars::openai::OpenAIEmbed;
+use crate::functions::scalars::sentence_transformers::BertUdf;
+use crate::functions::scalars::similarity::CosineSimilarity;
 
 /// All builtin functions available for all sessions.
-pub static FUNCTION_REGISTRY: Lazy<FunctionRegistry> = Lazy::new(FunctionRegistry::new);
+pub static FUNCTION_REGISTRY: Lazy<Mutex<FunctionRegistry>> =
+    Lazy::new(|| Mutex::new(FunctionRegistry::new()));
```
tychoish, line 50: Does datafusion require that the function registry be global and mutable at runtime? Of users can load their own functions into a global registry, then users can leak functionality to other users in a multi-user deployment (serverless/etc.) and that doesn't seem particularly tenable. b

I'm not sure what makes sense implementation wise, but I think either the function registry needs to be tied to a users' session, or it should be immutable. 

This might quickly become moot, but is there anyway that the registry itself can manage it's own concurrency control rather than needing to be wrapped in a lock?

crates/sqlbuiltins/src/functions/scalars/sentence_transformers.rs
```diff
@@ -0,0 +1,253 @@
...
+    const EXAMPLE: &'static str = "sentence_transformers.minilm_l6_v2('This is a sentence')";
+
+    const FUNCTION_TYPE: FunctionType = FunctionType::Scalar;
+
+    fn signature(&self) -> Option<Signature> {
+        Some(self.signature.clone())
+    }
+}
+
+impl ScalarUDFImpl for BertUdf {
```
tychoish, line 105: [curiosity] what's the reason for chosing to implement this trait rather than BuiltinUDF (as the other scalar functions are implemented?)

crates/sqlbuiltins/src/functions/scalars/sentence_transformers.rs
```diff
@@ -0,0 +1,253 @@
...
+impl std::fmt::Debug for BertUdf {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("BertUdf").finish()
+    }
+}
+
+impl BertUdf {
+    pub fn new() -> Self {
+        let device = Device::Cpu;
+        let model_id = "sentence-transformers/all-MiniLM-L6-v2".to_string();
```
tychoish, line 46: does this make sense to put in a constant somewhere?

crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -337,8 +396,50 @@ impl FunctionRegistry {
         }
         None
     }
+
+    fn register_udf(&mut self, f: Arc<dyn BuiltinScalarUDF>) {
+        let keys = match f.namespace() {
+            // we register the function under both the namespaced entry and the normal entry
+            // e.g. select foo.my_function() or select my_function()
```
tychoish, line 403: does it make sense to only register the legacy/existing functions in the global namespace, and put everything else in a namespace?

crates/sqlbuiltins/Cargo.toml
```diff
@@ -33,3 +33,12 @@ memoize = { version = "0.4.2", features = ["full"] }
 async-openai = "0.18.2"
 tokio.workspace = true
 reqwest.workspace = true
+candle-nn = "0.3.3"
+candle-core = "0.3.3"
+candle-transformers = "0.3.3"
+hf-hub = "0.3.2"
+tokenizers = "0.15.1"
+serde_json.workspace = true
+lance-linalg =  { git = "https://github.com/lancedb/lance", rev = "310d79eccf93f3c6a48c162c575918cdba13faec" }
```
tychoish, line 42: does it make sense to (eventually) put the extension stuff in a different crate?

crates/sqlbuiltins/src/functions/mod.rs
universalmind303: > Does datafusion require that the function registry be global and mutable at runtime? Of users can load their own functions into a global registry, then users can leak functionality to other users in a multi-user deployment (serverless/etc.) and that doesn't seem particularly tenable. b


yeah that was my comment about moving the function registry into the SessionCatalog. Having it attached to a session is better long term & prevents any accidental leaks between sessions.

crates/sqlbuiltins/src/functions/scalars/sentence_transformers.rs
universalmind303: No sense in adding more tech debt. We're already going to have to migrate to `ScalarUDFImpl` at some point anyways.

crates/sqlbuiltins/src/functions/scalars/sentence_transformers.rs
universalmind303: There are many Bert models, The idea was to have that as a kind of base struct that each sentence_transformer udf would have their own static implementation of. So the config vars like this will eventually be a part of that.

crates/sqlbuiltins/Cargo.toml
universalmind303: Yeah i think our end goal should be to load most of them as dylibs and only have a select few _(if any)_ that are actually embedded into the binary but disabled.

universalmind303:
i'm not very happy with this implementation. Closing for now. Will reopen a PR for the similarity functions.

crates/sqlbuiltins/src/functions/mod.rs
tychoish: got it. 

I think having it attached to the session is blocking releasing those code.

crates/sqlbuiltins/src/functions/scalars/sentence_transformers.rs
tychoish: Having different ways of implementing these functions is also adding technical debt if the transition plan is long.

Issue #2614, opened 2024-02-08
Feat: `read_uploads` table func


scsmithr:
## Description

Motivation: Allow users to query files they've uploaded while making use of our existing credentials setup.

What this would look like for the end user:

```sql
select * from stage.'my_file.csv';
```

- `stage` as the schema name is tbd
- 'my_file.csv' is the file the user uploaded in cloud

Implementation:
- New builtin schema (stage).
  - Users should not be able to create objects in this schema.
- Extend external dispatching rules to check the `stage` schema, dispatch to user files as appropriate.
  - Likely extension point: https://github.com/glaredb/glaredb/blob/7c4270c1334cbcee052fd264ac76b1a496d74243/crates/sqlexec/src/dispatch/external.rs#L95-L131
  - This should use the object store that we create on the `NativeTableStorage` since that will have the credentials already configured for the user bucket.
  - We'll need to thread stuff down related to the bucket/location of user files.

Future:
- Make `stage` an implicit schema such that `select * from 'my_file.csv'` would resolve to the user file.

greyscaled:
Question, would we eventually support `DROP TABLE stage.<my_file>` ?

greyscaled:
We:

- decided on not adding the `glare_uploads` schema
- will add a table func `read_uploads(filename, filetype)`

tychoish:
I like the tablefunc idea a lot and it's such a good starting point. 

I the the only thing I'd change is that it probably wants to be `read_upload` rather than "uploads". 

I keep playing in my mind with some alternate text (because this isn't quite like the other table-funcs so maybe it's fine that it's not `read_`) that takes advantage of being in a `FROM` clause, as in `SELECT * FROM glare_upload(<path>)`

Let me know if you want to pair on this grey.

greyscaled:
> I the the only thing I'd change is that it probably wants to be `read_upload`

Yea, I think the 's' was accidental, +1 for `upload`


> I keep playing in my mind with some alternate text (because this isn't quite like the other table-funcs so maybe it's fine that it's not `read_`) that takes advantage of being in a `FROM` clause, as in `SELECT * FROM glaredb_upload(<path>)`

Worth playing around with and finding what feels best. I'm generally happy with things that have a `glare_` prefix (instead of `glaredb_`), though maybe it's beneficial to keep this one `read_` for symmetry with our other table funcs - I'm unsure.

[...]

PR #2617, opened 2024-02-09
fix(metastore): make builtin schemas readonly

greyscaled:
Summary
---

This adds a restriction in metastore for creating entries on builtin schemas, other than the default schema (`public`).

<details><summary>Details</summary>
<p>

Initially, this added a builtin schema `glare_uploads` (#2614), but we later decided to remove it and instead use a table funcition.
At the time, `glare_uploads` was chosen because prefixing internal schemas with `glare_` reduces collision chance, and is what pg does.

<details><summary>Read more on pg_</summary>
<p>

From https://www.postgresql.org/docs/current/ddl-schemas.html

> Since system table names begin with pg_, it is best to avoid such names to ensure that you won't suffer a conflict if some future version defines a system table named the same as your table. (With the default search path, an unqualified reference to your table name would then be resolved as the system table instead.) System tables will continue to follow the convention of having names beginning with pg_, so that they will not conflict with unqualified user-table names so long as users avoid the pg_ prefix.

</p>
</details> 

<details><summary>Progress</summary>
<p>

- [x] Add `glare_uploads` schema
- [x] Make schema read-only
- [x] Add test(s) 

</p>
</details> 
</p>
</details>

### DIFF
--- a/crates/metastore/src/database.rs
+++ b/crates/metastore/src/database.rs
@@ -73,7 +73,7 @@ static BUILTIN_CATALOG: Lazy<BuiltinCatalog> = Lazy::new(|| BuiltinCatalog::new(
 pub struct DatabaseCatalog {
     db_id: Uuid,
 
-    /// Reference to underlying persistant storage.
+    /// Reference to underlying persistent storage.
     storage: Arc<Storage>,
 
     /// A cached catalog state for a single database.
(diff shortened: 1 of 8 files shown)


tychoish (approved):
I'm ok with this but @scsmithr or @vrongmeal should give it the once over.

scsmithr (approved):

PR #2631, opened 2024-02-09
fix(deps): remove dynamic linking to 'lzma'

universalmind303:


tests/tests/test_smoke.py
universalmind303, line 1: formatting changes were auto applied from ruff.

tychoish (approved):

Issue #2604, opened 2024-02-06
feature: query directly from google sheets with glaredb


talagluck:
## Description
As per https://github.com/GlareDB/meta/issues/118#issuecomment-1900199042:
It should be possible to query directly from a google sheets url like:
```python
sheet_key = "1NI0I2s2Hw-KPK32d7vMqLGetI9FBzYAfTcRm1j7L11Q"
sheet_name = "Sheet1"
url = f"https://docs.google.com/spreadsheet/ccc?key={sheet_key}&sheet={sheet_name}&output=csv"
con.sql(
    f"SELECT * FROM '{url}'"
)
```

Currently, this returns an error: 
```
ExecutionException: Error during planning: missing file extension: https://docs.google.com/spreadsheet/ccc?key=1NI0I2s2Hw-KPK32d7vMqLGetI9FBzYAfTcRm1j7L11Q&sheet=Sheet1&output=csv
```

I also tried `con.sql(f"SELECT * FROM read_csv('{url}')"), and got 
```
ExecutionException: External error: Exec error: Remote session error: unable to dispatch table access: status: Unknown, message: "transport error", details: [], metadata: MetadataMap { headers: {} }
```

PR #2633, opened 2024-02-09
feat: sniff google sheets csv paths

tychoish:
:shrug:

#2604

### DIFF
--- a/crates/datafusion_ext/src/planner/relation/mod.rs
+++ b/crates/datafusion_ext/src/planner/relation/mod.rs
@@ -235,8 +235,8 @@ impl<'a, S: AsyncContextProvider> SqlQueryPlanner<'a, S> {
 /// Returns a reference to table func by inferring which function to use from a
 /// given path.
 fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
-    let ext = if path.contains("docs.google.com/spreadsheet") && path.trim().ends_with("output=csv")
-    {
+    let is_google_sheet = path.contains("docs.google.com/spreadsheet");
+    let ext = if is_google_sheet {
         "csv".to_string()
     } else {
         Path::new(path)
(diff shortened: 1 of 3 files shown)


crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -235,12 +235,18 @@ impl<'a, S: AsyncContextProvider> SqlQueryPlanner<'a, S> {
 /// Returns a reference to table func by inferring which function to use from a
 /// given path.
 fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
-    let ext = Path::new(path)
-        .extension()
-        .ok_or_else(|| DataFusionError::Plan(format!("missing file extension: {path}")))?
-        .to_str()
-        .ok_or_else(|| DataFusionError::Plan(format!("strange file extension: {path}")))?
-        .to_lowercase();
+    let ext = if path.contains("docs.google.com/spreadsheet") && path.trim().ends_with("output=csv")
```
vrongmeal, line 238: Should we instead pass the `path` as `&Url` and check if there is a query parameter output with value `csv`? `path` ending with `output=csv` feels wrong...

tychoish: It does feel wrong. 

I also wonder if we can change the content (mime?) type header without specifying an argument?

I guess also we can do some amount of heroics if we get a URL that _doesn't_ have the argument, we could add it?

tychoish:
I poked at this a bit and the issue is that `ObjectStoreUrl::parse` returns this error in [these functions](https://github.com/glaredb/glaredb/blob/tycho/google-sheets-hack/crates/datasources/src/common/url.rs#L164-L184) which we're hitting, and I think this means we can't use read_csv, on the other hand I think we can probably just write a dedicated function and call reqwest directly. 

This is definitely not going to be a hack and will take a middle/moderate amount of work

tychoish:
closing, but will reuse your test spreadsheet @talagluck

Issue #2462, opened 2024-01-19
Hive partitioning when using `COPY TO`


universalmind303:
## Description

something like
`COPY transactions TO './transactions' FORMAT parquet PARTITION BY (year,quarter)'

should result in a directory as such: 

transactions/
├── year=2020/
│   ├── quarter=Q1/
│   ├── quarter=Q2/
│   ├── quarter=Q3/
│   └── quarter=Q4/
├── year=2021/
│   ├── quarter=Q1/
│   ├── quarter=Q2/
│   ├── quarter=Q3/
│   └── quarter=Q4/

melbourne2991:
I've picked this up - wip branch (https://github.com/GlareDB/glaredb/pull/2634)

PR #2634, opened 2024-02-11
feat(`COPY TO`): hive partitioning support

melbourne2991:
Addresses (https://github.com/GlareDB/glaredb/issues/2462)

Provides hive partitioning support for Parquet & Json. 

Missing from this PR:

- Remaining formats (lance, csv, bson)
- Reading hive partitioned files


crates/datasources/Cargo.toml
```diff
@@ -13,6 +13,8 @@ apache-avro = "0.16"
 async-channel = "2.1.1"
 async-stream = "0.3.5"
 async-trait = { workspace = true }
+arrow-array = { version = "50.0.0", default-features = false }
+arrow-schema = { version = "50.0.0", default-features = false }
```
tychoish, line 17: use the version of arrow that's re-exported via the datafusion package `datafusion::arrow`

crates/datasources/src/common/sink/parquet.rs
```diff
@@ -14,20 +17,21 @@ use datafusion::physical_plan::{DisplayAs, DisplayFormatType, SendableRecordBatc
...
+
+use crate::common::sink::write::parquet::{output_single_parquet_file_parallelized, ParallelParquetWriterOptions};
+use crate::common::sink::write::{AbortableWrite, MultiPart};
+
+use super::write::demux::start_demuxer_task;
 
 const BUFFER_SIZE: usize = 8 * 1024 * 1024;
 
 #[derive(Debug, Clone)]
 pub struct ParquetSinkOpts {
```
tychoish, line 31: my inclination was that we'd be able to do this _above_ the level of the individual formats. so that you could have hive partitioning for any supported output (file?) format.

melbourne2991: Gotcha.

I'll have a think about what that might look like. I suspect with a generic approach there would be some performance trade offs for parquet but let me get back to you.

tychoish: Interesting. I would imagine that generic approaches to reading wouldn't work super well with _reading_ parquet but I'd imagine that writing data might be roughly the same. What were you thinking the impact would be?

crates/datasources/Cargo.toml
melbourne2991: Yeah I'm having some trouble using the `downcast_dictionary_array!` macro in write/demux without adding that crate explicitly 🤔 . I'm far from an expert on Rust macros so maybe I'm missing something obvious?

crates/datasources/src/common/sink/parquet.rs
melbourne2991: Yes I think you're right. There's some code in the DF repo that parallelizes single-file  parquet encoding that I thought may interact with the hive stuff but it looks like it's independent - which does make sense now I think about it.

I've updated the PR to introduce an abstraction that accepts a data sink "producer" that will be called on each new partitioned file with the partition path. So each format just needs to implement the SinkProducer trait and this will unlock hive partitioning. Please see `hive_partitioning.rs`.

At the moment I've just added Parquet (see copy_to.rs) but it should be easy to add for other formats. Please let me know what you think of this approach. 

I've also removed the single-file parallelization code for parquet that I'd brought over from DF as it's not needed for hive partitioning (though may be worth looking at adding that later as an optimization).

crates/datasources/Cargo.toml
```diff
@@ -13,6 +13,7 @@ apache-avro = "0.16"
 async-channel = "2.1.1"
 async-stream = "0.3.5"
 async-trait = { workspace = true }
+arrow-schema = { version = "50.0.0", default-features = false }
```
tychoish, line 16: in general we should use the version of arrow that's reexported from `datafusion` at `datafusion::arrow`

crates/datasources/src/common/sink/write/demux.rs
```diff
@@ -0,0 +1,433 @@
...
+        //     row_count_demuxer(
+        //         tx,
+        //         input,
+        //         context,
+        //         base_output_path,
+        //         file_extension,
+        //         single_file_output,
+        //     )
+        //     .await
+        // }),
```
tychoish, line 123: delete?

tychoish (commented):
this all looks really good to me.
- I'd love to see some tests on "what happens when you partition by something other than a date" 
- we should definitely open another ticket for "reading from hive-partitioned files." right now you can use our glob, function, and that helps, but there is a push down projection that this might not be able to do. Clearly out of scope for this ticket, but it'd be killer feature either way. 
- I'd love to see a test with another file format (json or bson?) just just to make sure that it's generic enough and doesn't rely on something parquet specific.
- I think it'd be good to be explicit about the expectation that the partitioned field remains in the output data or is elided because it's in the partition, so a test there would be good.

melbourne2991:
> this all looks really good to me.
> 
> * I'd love to see some tests on "what happens when you partition by something other than a date"
> * we should definitely open another ticket for "reading from hive-partitioned files." right now you can use our glob, function, and that helps, but there is a push down projection that this might not be able to do. Clearly out of scope for this ticket, but it'd be killer feature either way.
> * I'd love to see a test with another file format (json or bson?) just just to make sure that it's generic enough and doesn't rely on something parquet specific.
> * I think it'd be good to be explicit about the expectation that the partitioned field remains in the output data or is elided because it's in the partition, so a test there would be good.

Agree on all these points - thanks for the feedback. (Just a note: the PR isn't in its final form yet. The current test was primarily for development ease - more comprehensive tests are on the way!).

crates/datasources/Cargo.toml
melbourne2991: Yeah I was having some difficulties using the downcast_dictionary_array! macro in write/demux.rs using the re-export.  I'm not sure how to get around this issue yet... will try to find a solution.

```
error[E0433]: failed to resolve: use of undeclared crate or module `arrow_schema`
   --> crates/datasources/src/common/sink/write/demux.rs:344:17
    |
344 | / ...   downcast_dictionary_array!(
345 | | ...       col_array =>  {
346 | | ...           let array = col_array.downcast_dict::<StringArray>()
347 | | ...               .ok_or(DataFusionError::Execution(format!("it is not yet supported to write to hive partitions with datatype {}",
...   |
356 | | ...       _ => unreachable!(),
357 | | ...   )
    | |_______^ use of undeclared crate or module `arrow_schema`
    |
    = note: this error originates in the macro `downcast_dictionary_array` (in Nightly builds, run with -Z macro-backtrace for more info)
help: consider importing one of these items
    |
29  + use calamine::DataType;
    |
29  + use deltalake::kernel::DataType;
    |

```

melbourne2991: I think it may be because it expands to code that includes the `arrow_schema` qualifier?

```
#[macro_export]
macro_rules! downcast_dictionary_array {
    ($values:ident => $e:expr, $($p:pat => $fallback:expr $(,)*)*) => {
        downcast_dictionary_array!($values => {$e} $($p => $fallback)*)
    };

    ($values:ident => $e:block $($p:pat => $fallback:expr $(,)*)*) => {
        match $values.data_type() {
            arrow_schema::DataType::Dictionary(k, _) => {
                $crate::downcast_integer! {
                    k.as_ref() => ($crate::downcast_dictionary_array_helper, $values, $e),
                    k => unreachable!("unsupported dictionary key type: {}", k)
                }
            }
            $($p => $fallback,)*
        }
    }
}
```

crates/datasources/Cargo.toml
melbourne2991: Have removed since this isn't needed unless we want to support dictionaries.

crates/protogen/src/sqlexec/copy_to.rs
```diff
@@ -90,12 +90,18 @@ pub struct CopyToFormatOptionsCsv {
 pub struct CopyToFormatOptionsJson {
     #[prost(bool, tag = "1")]
     pub array: bool,
+
+    #[prost(string, repeated, tag = "2")]
+    pub partition_columns: Vec<String>,
```
tychoish, line 95: I think we need to do this for CSV, Lance, and BSON. 

I realized that we missed having one of these options structs for bson which I added in #2704

crates/sqlbuiltins/src/validation.rs
```diff
@@ -22,6 +22,9 @@ pub enum ValidationError {
 
     #[error("Format '{format}' not supported by datasource '{datasource}'")]
     FormatNotSupportedByDatasource { format: String, datasource: String },
+
+    #[error("Partitioning is not supported by format '{format}'")]
+    PartitionByNotSupportedByFormat { format: String },
```
tychoish, line 27: I'd think that we could have this be an enum that we already have for the format?

crates/sqlexec/src/parser.rs
```diff
@@ -1373,6 +1400,20 @@ mod tests {
...
+        let test_cases = ["COPY table TO './dest' FORMAT parquet PARTITION BY (year, quarter)"];
+
+        for test_case in test_cases {
+            let stmt = CustomParser::parse_sql(test_case)
+                .unwrap()
+                .pop_front()
+                .unwrap();
+            assert_eq!(test_case, stmt.to_string().as_str());
+        }
+    }
```
tychoish, line 1414: can we test the error conditions:
- `PARTITION` without `BY`
- an unsupported format?
- an unknown format?

crates/sqlbuiltins/src/validation.rs
```diff
@@ -165,3 +168,13 @@ pub fn validate_copyto_dest_format_support(dest: &str, format: &str) -> Result<(
         })
     }
 }
+
+pub fn validate_copyto_format_partition_support(format: &str) -> Result<()> {
+    if matches!(format, |"parquet"| "json") {
```
tychoish, line 173: I feel like we could have a method on the enum or something that could do this validation rather than on a string. Is this extension? I worry about this falling out of sync with other variations of the enum?

crates/sqlexec/src/planner/physical_plan/copy_to.rs
```diff
@@ -237,13 +263,24 @@ fn get_sink_for_obj(
...
-            store,
-            path,
-            JsonSinkOpts {
-                array: json_opts.array,
-            },
-        )),
+        CopyToFormatOptions::Json(json_opts) => {
+            let schema = source.schema();
+
+            println!("location: {}", location);
```
tychoish, line 269: should probably omit this

crates/sqlexec/src/planner/physical_plan/copy_to.rs
```diff
@@ -237,13 +263,24 @@ fn get_sink_for_obj(
...
+            Box::new(HivePartitionedSinkAdapter::new(
+                JsonSinkProducer::from_obj_store(
+                    store,
+                    JsonSinkOpts {
+                        array: json_opts.array,
+                    },
+                ),
+                json_opts.partition_columns,
+                location.into(),
+                ".json".to_string(),
```
tychoish, line 280: why does this need to be a string and not using the enum or another property?

crates/protogen/src/sqlexec/copy_to.rs
tychoish: I merged in #2704 which should make it easier to add partitioning support there too.

universalmind303:
marking as draft as it's not actively waiting on review.

@melbourne2991 please feel free to ping us when it is ready.

tychoish:
@melbourne2991 wanted to check in on this. Is there anything I can do to help you on this?

melbourne2991:
hey @tychoish, apologies, I've been swamped lately - I'm not sure if I'll have time to get around to this in any reasonable time frame - happy for someone else to pick it up, there shouldn't be too much effort left on it I hope

PR #2639, opened 2024-02-12
feat(read_ file funcs): infer from compressed formats

### LINKED ISSUE #2602: `select * from 'some_file.json.gz`
## Description

you can select from 'json.gz' files via `select * from read_ndjson('some_file.json.gz')`, but you can't use the path directly `select * from 'some_file.json.gz'`

hemanth94:
Fixes #2602 
Inferring functions from compressed file formats.
@universalmind303 Please suggest changes if needed.

Thanks.


crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,10 +268,45 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
...
                 "unable to infer how to handle file extension: {ext}"
             )))
         }
     })
 }
+
+fn infer_func_from_compressed_file(filename: &str) -> Result<OwnedTableReference> {
+    if filename.contains(".json.gz")
+        | filename.contains(".json1.gz")
+        | filename.contains(".ndjson.gz")
```
tychoish, line 286: given that we now have this `read_json` function for non-line delimited json does it make sense to use ndjson for `.json.gz` file endings? I think it's easier to add `json.gz` in than it is to take it out, so maybe we should go with that. 

also I'm not sure why `json1.gz` is covered here, I'd remove it unless there's a reason.

crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,10 +268,45 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
...
+    {
+        return Ok(OwnedTableReference::Partial {
+            schema: "public".into(),
+            table: "ndjson_scan".into(),
```
tychoish, line 290: can we use the `read_` functions here?

crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,10 +268,45 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
...
+    } else if filename.contains(".csv.gz") {
+        return Ok(OwnedTableReference::Partial {
+            schema: "public".into(),
+            table: "csv_scan".into(),
+        });
+    } else if filename.contains(".bson.gz") {
+        return Ok(OwnedTableReference::Partial {
+            schema: "public".into(),
+            table: "read_bson".into(),
+        });
```
tychoish, line 306: I'm not 100% sure that this will actually work.

crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,10 +268,45 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
...
+    {
+        return Ok(OwnedTableReference::Partial {
+            schema: "public".into(),
+            table: "ndjson_scan".into(),
+        });
+    } else if filename.contains(".parquet.gz") {
+        return Ok(OwnedTableReference::Partial {
+            schema: "public".into(),
+            table: "parquet_scan".into(),
+        });
```
tychoish, line 296: does it make sense to gzip a parquet file given its internal compression, and the fact that gzip will make it difficult/impossible to partition anything from the file.

crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -242,6 +242,13 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
         .ok_or_else(|| DataFusionError::Plan(format!("strange file extension: {path}")))?
         .to_lowercase();
 
+    let binding = PathBuf::from(path);
+    let filename = binding
+        .file_name()
+        .ok_or_else(|| DataFusionError::Plan(format!("NO file name provided: {path}")))?
+        .to_str()
+        .ok_or_else(|| DataFusionError::Plan(format!("Improper file name: {path}")))?;
```
tychoish, line 250: I'd say "invalid" rather than improper.

tychoish (commented):
This needs tests, and there are a few questions.

universalmind303 (changes requested):
Thanks for picking this one up @hemanth94! 

While not mentioned in the original issue, I think it'd be pretty easy to include support for all supported compression types via datafusion's `FileCompressionType`. 

We should also include a few simple slts. For this, i'd suggest compressing the `userdata1` dataset that's used for parquet, ndjson & csv and save it to the same respective directories. 

ex: `testdata/csv/userdata1.csv` -> `testdata/csv/userdata1.csv.gz`
`testdata/parquet/userdata1.parquet` -> `testdata/parquet/userdata1.parquet.gz`

crates/datafusion_ext/src/planner/relation/mod.rs
universalmind303: I think we should still pass this in to the planner even if it's invalid. I don't think this stage of the planning should be concerned with plan validity. It's only concerned with creating the logical plan from a valid query, even if that plan itself is invalid.

crates/datafusion_ext/src/planner/relation/mod.rs
tychoish: I mean, I don't think this is a very strong opinion, but I think this is likely to lead people to having very poor experiences, and given that, maybe it'd be best to disallow it? 

There's no reason we need to recapitulate arrow or datafuison's mistakes.

crates/datafusion_ext/src/planner/relation/mod.rs
tychoish: I'm not sure that makes a lot of sense, or at least that's not consistent with the implementation. If we want to implement what you describe than we split the right side of the string on `.` and match the penultimate phrase against the formats, and the ultimate segment on the compression formats. Really, If it's not _this_ code's responsibility to decide if the reading function can handle the compression formats, then we should _actually_ ignore them here.

But I don't think that's what we actually want, and I don't think "pass invalid arguments through and hope that the downstream code handles it correctly," is a reasonable design and also (more importantly) it makes it harder to return a clear error message to users in invalid  cases.

crates/datafusion_ext/src/planner/relation/mod.rs
universalmind303: but then if people have gzipped parquet files already created from <insert tool>, it's unreadable. I don't think we should put this limitation on the readers. I could see this being a valid argument for writing though.

crates/datafusion_ext/src/planner/relation/mod.rs
universalmind303: It's not "hoping downstream code handles it correctly". It's a matter of separation of logic and keeping things DRY. The table function already need to handle this. There's no reason to check it here also. Otherwise we have 2 places to check identical logic. 

```sql
select * from read_bson('some_file.bson.gz');
select * from 'some_file.bson.gz';
```

crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,10 +268,45 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
...
+        }
         ext => {
             return Err(DataFusionError::Plan(format!(
                 "unable to infer how to handle file extension: {ext}"
             )))
         }
     })
 }
+
+fn infer_func_from_compressed_file(filename: &str) -> Result<OwnedTableReference> {
```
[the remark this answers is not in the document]
hemanth94: > I think we want these conditions to be "ends with" rather than "contains"?
Yes, "ends with" is better way to ascertain the file's format. I will modify this.

>I'm pretty sure we also support bzip (.bz) and zstd (`.xz) compression methods.
I don't have adequate knowledge about different types we should support, But can surely add as required.

Thanks.

hemanth94:
> While not mentioned in the original issue, I think it'd be pretty easy to include support for all supported compression types via datafusion's `FileCompressionType`.

Sure I will try this.

> We should also include a few simple slts. For this, i'd suggest compressing the `userdata1` dataset that's used for parquet, ndjson & csv and save it to the same respective directories.
> 
> ex: `testdata/csv/userdata1.csv` -> `testdata/csv/userdata1.csv.gz` `testdata/parquet/userdata1.parquet` -> `testdata/parquet/userdata1.parquet.gz`

Will do this as well.

crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,10 +268,45 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
             schema: "public".into(),
             table: "read_bson".into(),
         },
+        "gz" => {
```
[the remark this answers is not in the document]
hemanth94: I'll use this to include all compression types.

I overlooked this code logic you suggested while making commits, Will use this as well properly.

crates/datafusion_ext/src/planner/relation/mod.rs
tychoish: > It's a matter of separation of logic and keeping things DRY.

My concern is primarily for centralizing the encoding of the validation logic for all of the formats and table functions (and DDL-created dispatches to these table providers), so that we can have clear error messages, and avoid having duplicate validation logic.

> The table function already need to handle this scenario.

Sort of? The table functions are going to error if they can't read the data or if it's in the wrong format. It's maybe _not_ necessary for them to validate that the path itself is well-formed.

> It's redundant to check it here also.

It is redundant, but so is enumerating every possible compression algorithm and extension variant. We shouldn't need to care about `.bson.gz` vs `.bson.gzip` (and so forth). I think the reasonable solution to getting something here that's helpful and useful is:

- split the string by `.` if the last or second-to-last element in the resulting sequence is `bson`, `ndjson`, `jsonl`, or `csv` then dispatch to the appropriate function. 
- let's leave `.json` out of this as much as possible because of ambiguity between text sequences and line-separated json. 
- write tests to validate that the table functions behave correctly with compression and propagate reasonable error messages otherwise.

crates/datafusion_ext/src/planner/relation/mod.rs
tychoish: > but then if people have gzipped parquet files already created from <insert tool>, it's unreadable. I don't think we should put this limitation on the readers.

I don't think we have a write path that compresses any objects, anyway. 

I'm pretty sure that there aren't tools that support this as much as people using `gzip` from the shell, but that doesn't make it reasonable or correct. 
 
Should we support use-cases which are fundamentally misunderstandings or mistakes, especially when likely to produce bad experiences just because someone might stumble into them? 

It's much easier to add something like this later if there's a valid, non-pathological use-case, than it is to spend [the lifetime of the product] doing damage control. Sometimes errors in these cases are more kind than supporting something that's broken but possible.

crates/datafusion_ext/src/planner/relation/mod.rs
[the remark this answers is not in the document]
hemanth94: @universalmind303 `FileCompressionType` Only supports [".gz", ".bz2", ".xz", ".zst"]. 

Compression format `BZIP` ".bz" seems to be missing here. Is it worth to use `FileCompressionType`?
Can we not manually check compression formats? and then call inferring function recursively. 
What if, we need to add a different compression format in future?

Kindly suggest.

testdata/sqllogictests/infer.slt
```diff
@@ -51,3 +51,33 @@ select count(*) from './testdata/parquet/*.parquet'
...
+  limit 1
+----
+1  Amanda
+
+#For read_parquet table function with compressed files 
+
+query
+select count(*) from './testdata/parquet/userdata1.parquet.gz'
+----
+1000
```
tychoish, line 75: can we also test other combinations 

I think that parquet shouldn't work and we should add a test that ensures we don't change that behavior.

tychoish:
> SLTs are failing with this error, Seems parquet."gz" format is not supported.

this seems correct and like the desired behavior.

tychoish:
I just wanted to check in and see where this was and if there was anything you needed from us on this.

hemanth94:
Hello, @tychoish    

I think this can be merged now. 

Last changes i made, **_as you requested_**.
- I added bz2, xz, zst formats of csv, Json, Parquet. Added respective SLTs for them, which passed.
- Changed SLTs for parquet to catch Error, since `Parquet files` are supposed to fail with compressed formats. 

That's all.
Thanks,

tychoish (approved):

greyscaled (approved):

universalmind303 (commented):
Thanks for the contribution @hemanth94!

PR #2641, opened 2024-02-12
feat: flightsql datasource

tychoish:


crates/datasources/src/lance/mod.rs
```diff
@@ -117,7 +117,7 @@ impl ExecutionPlan for LanceInsertExecPlan {
     }
 
     fn schema(&self) -> SchemaRef {
-        TableProvider::schema(&self.dataset)
+        COUNT_SCHEMA.clone()
```
universalmind303, line 120: Is this unrelated to flightsql?

crates/datasources/src/common/query.rs
```diff
@@ -0,0 +1,116 @@
+use std::fmt::Write;
+
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::error::DataFusionError;
+use datafusion::logical_expr::Expr;
+
+use super::errors::Result;
+use super::util::{self, Datasource};
+
+pub fn from_exprs(
```
universalmind303, line 10: we shouldn't use the module to provide context for a function name. 

`from_exprs` has no context without the `query::` part. `query::from_exprs`

```suggestion
pub fn exprs_to_query_string(
```

universalmind303: I'd also proceed with caution as not all sql dialects are the same, as we add more expr pushdowns, we'll need to specialize for different dialects.

crates/datasources/src/flightsql/insert.rs
```diff
@@ -0,0 +1,165 @@
...
+    SendableRecordBatchStream,
+    Statistics,
+};
+use futures::StreamExt;
+use tonic::transport::Channel;
+
+use super::FlightSqlSourceConnectionOptions;
+use crate::common::util::create_count_record_batch;
+
+pub struct ExecPlan {
```
universalmind303, line 29:
```suggestion
pub struct FlightSqlExecPlan {
```

crates/sqlbuiltins/src/functions/table/flightsql.rs
```diff
@@ -0,0 +1,69 @@
...
+use crate::functions::ConstBuiltinFunction;
+
+#[derive(Debug, Clone, Copy)]
+pub struct ReadFlightSql;
+
+impl ConstBuiltinFunction for ReadFlightSql {
+    const NAME: &'static str = "read_flightsql";
+    const DESCRIPTION: &'static str = "Reads a FlightSQL table";
+    const EXAMPLE: &'static str =
+        "SELECT * FROM read_flightsql('https://localhost:37019', 'database', 'token')";
```
universalmind303, line 23: I don't think `database` and `token` should be specified. We should instead allow users to pass in headers as each flightsql implementation likely has it's own requirements.

```sql
read_flightsql('https://localhost:37019', authorization => '<token>', database => 'database', some_other_header => 'some_value')"
```

crates/datasources/src/flightsql/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+
+#[derive(Debug, Clone)]
+pub struct FlightSqlSourceConnectionOptions {
+    pub uri: String,
+    pub database: String,
+    pub table: String,
+    pub token: String,
+}
+
+pub struct FlightSqlSourceProvider {
```
universalmind303, line 37:
```suggestion
pub struct FlightSqlTableProvider {
```

crates/datasources/src/lance/mod.rs
tychoish: yeah, I can pull this into a separate PR, mostly didn't want to lose track of this. 

I think our exec plans in general say that they have the schema of their _inputs_ and not their _outputs_ so the problem is not localized to this case.

crates/datasources/src/flightsql/mod.rs
```diff
@@ -0,0 +1,152 @@
...
+        let catalogs = client.get_catalogs().await?;
+        let schema = Arc::new(
+            client
+                .get_db_schemas(CommandGetDbSchemas {
+                    catalog: Some(catalogs.to_string()),
+                    db_schema_filter_pattern: Some(db),
+                })
+                .await?
+                .try_decode_schema()?,
+        );
```
universalmind303, line 64: Does this return the correct schema? This doesn't seem right to me.

crates/datasources/src/common/query.rs
tychoish: > I'd also proceed with caution as not all sql dialects are the same, as we add more expr pushdowns, we'll need to specialize for different dialects.

the code in question already has specialization for this. 

> we shouldn't use the module to provide context for a function name.

meh, I disagree

- https://github.com/rust-lang/rfcs/blob/master/text/0356-no-module-prefixes.md
- https://github.com/rust-lang/api-guidelines/discussions/66

anyway, this isn't the right place for this conversation.

crates/datasources/src/flightsql/mod.rs
tychoish: yeah, probably should be a table thing.

crates/sqlbuiltins/src/functions/table/flightsql.rs
tychoish: to be honest, this was just a prototype; my inclination is that we'll actually make this function be `read_influxdb` and avoid providing generic flightsql cases, for exactly the reasons you have (though I think we can entertain the potential fantasy that flightsql is actually generic.)

crates/datasources/src/common/query.rs
universalmind303: i think this is slightly different than the examples you provided. 

an error inside `io::Error` is still an error, but more specifically it is an IO error, so any of `IOError`, `io::Error`, `Error` still indicate that it's an error. `io` just provides additional context. same with the `log::LogLevel` example. 

`from_exprs` is completely meaningless without `query::from_exprs`.

PR #2642, opened 2024-02-13
fix: make insert plan schema correct and clear

tychoish:


crates/datasources/src/postgres/query_exec.rs
```diff
@@ -105,7 +105,7 @@ impl ExecutionPlan for PostgresQueryExec {
     }
 }
 
-impl DisplayAs for PostgresQueryExec {
+impl DisplayAs for PostgresInsertExec {
     fn fmt_as(&self, _t: DisplayFormatType, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "PostgresQueryExec(query = {})", self.query)
```
scsmithr, line 110: Needs update

scsmithr (approved):

crates/datasources/src/postgres/query_exec.rs
tychoish:
```suggestion
        write!(f, "PostgresInsertExec(query = {})", self.query)
```

greyscaled:
This seems reasonable, will make sure it lands in 0.8.5

PR #2646, opened 2024-02-14
release: v0.8.5

greyscaled:


tychoish:
https://github.com/GlareDB/glaredb/pull/2642 <- can we wait to get this in? it's minor, and was definitely wrong before and while I think it's a harmless, the fact that I'm not 100% sure makes me want to slip it in...

greyscaled:
> #2642 <- can we wait to get this in? it's minor, and was definitely wrong before and while I think it's a harmless, the fact that I'm not 100% sure makes me want to slip it in...

Yep, will wait on it to merge before adding this commit on top

greyscaled:
@tychoish this now includes #2642 🎉

tychoish (approved):

PR #2647, opened 2024-02-14
chore(ci): move postgres tests and simplify tunnels testing

tychoish:
I started with the intention of just having the postgres tasks in
their test, but took the oppertunity to:

- clarify some capitalization that was bugging me (since I'll need to
  update the required builders, which are keyed off of this string).

- while I was editing this I realized that `tunnes/ssh` tests which we
  go through a lot of trouble to test with/without, are only present
  for mysql and postgres, and the only reason to skip running the
  tunnel tests, given the execution of the SLT runner and the hooks,
  is to omit running the `basic.slti` an extra time (but only for
  mysql/postgres), and given the new test architecture I think its
  safe to dispense with (and the invocations were redundnat in the
  main case.)

### DIFF
--- a/crates/slt/src/test.rs
+++ b/crates/slt/src/test.rs
@@ -47,8 +47,9 @@ pub trait Hook: Send + Sync {
         _config: &Config,
         _client: TestClient,
         _vars: &mut HashMap<String, String>,
-    ) -> Result<()> {
-        Ok(())
+    ) -> Result<bool> {
+        // TODO: make enum for skip/continue rather than booleans
+        Ok(true)
     }
 
     async fn post(
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -625,7 +625,7 @@ jobs:
 
           [ "${{matrix.settings.name}}" = "MongoDB" ] && ./scripts/create-fixture-mongodb.sh
 
-          just slt-bin --protocol=rpc ${{matrix.settings.path}}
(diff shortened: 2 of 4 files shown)


tychoish:
Ok this works now. whew.

crates/slt/src/test.rs
```diff
@@ -47,8 +47,9 @@ pub trait Hook: Send + Sync {
         _config: &Config,
         _client: TestClient,
         _vars: &mut HashMap<String, String>,
-    ) -> Result<()> {
-        Ok(())
+    ) -> Result<bool> {
+        // TODO: make enum for skip/continue rather than booleans
```
tychoish, line 51: I don't feel strongly that we need to do this right now, but I'm not opposed.

tychoish (commented):
I was reading through some of the code here again, and I realized that we could implement fixtures using the current hook mechanism, which would mean that the `slt` command would be able to figure out how to run scripts (etc.) before/after each test/file/etc. which could take care of setting up and/or running SLTs, which could untangle a lot of the CI configuration and scripts. 

**Proposal 1** have a set of names/scripts that get executed by the hook in any given SLT directory, `setup`, `teardown`, `setup_file`, `teardown_file`. 

**Proposal 2** same as 1, but we only do it for directories. (the slt package seems to have a `system` statement that we can use to execute inter-statement fixtures.

**Proposal 3** no magic names but we have a yaml/kdl/etc file with more specification of which commands, which might be mor explicit and easier to change. 

... (something else)

PR #2648, opened 2024-02-14
fix(virtual_listing): signatures

greyscaled:
A very minor fix to df signature implementations for `list_schemas` (missing) and `list_tables` (wrong len) that I found while working on something else.


crates/sqlbuiltins/src/functions/table/virtual_listing.rs
```diff
@@ -46,6 +46,13 @@ impl ConstBuiltinFunction for ListSchemas {
     const DESCRIPTION: &'static str = "Lists schemas in a database";
     const EXAMPLE: &'static str = "SELECT * FROM list_schemas('database')";
     const FUNCTION_TYPE: FunctionType = FunctionType::TableReturning;
+    fn signature(&self) -> Option<Signature> {
```
tychoish, line 49:
```suggestion

    fn signature(&self) -> Option<Signature> {
```

tychoish (approved):

PR #2653, opened 2024-02-15
chore(ci): build images from 'main'

greyscaled:
We are no longer getting updates to the `main` tag

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -644,7 +644,8 @@ jobs:
   docker-push:
     name: Build and Push Docker Image
     if: github.event_name == 'push' && github.repository == 'GlareDB/glaredb' && github.ref == 'refs/heads/main'
-    runs-on: ubuntu-latest-8-cores
+    needs: ["sql-logic-tests"]
+    runs-on: ubuntu-latest
     permissions:
       id-token: write # required to use OIDC auth
       contents: write # required to checkout code
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -42,8 +42,8 @@ jobs:
             !target/**/pgprototest
           key: ${{ runner.os }}-cargo-glaredb-build-${{ github.ref_name }}
           restore-keys: |
-             ${{ runner.os }}-cargo-glaredb-build-${{ hashFiles('**/Cargo.lock') }}
-             ${{ runner.os }}-cargo-glaredb-build-
+            ${{ runner.os }}-cargo-glaredb-build-${{ hashFiles('**/Cargo.lock') }}
+            ${{ runner.os }}-cargo-glaredb-build-
       - run: just build
       - run: cargo build --bin pgprototest
       - uses: actions/cache/save@v4
@@ -56,7 +56,6 @@ jobs:
         with:
           path: target/debug/pgprototest
           key: ${{ runner.os }}-pgprototest-bin-${{ github.run_id }}
-
 
   fmt:
     name: Format (rustfmt +nightly)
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -588,7 +576,6 @@ jobs:
           just slt-bin 'sqllogictests_snowflake/*'
           just slt-bin --protocol=rpc 'sqllogictests_snowflake/*'
 
-
```
greyscaled, line 591: All changes before this point were just automagic lint/fmt 🤷🏻

scsmithr (approved):

.github/workflows/ci.yaml
```diff
@@ -653,3 +640,25 @@ jobs:
...
+        uses: actions/checkout@v4
+      - name: Authenticate to Google Cloud
+        uses: google-github-actions/auth@v2
+        with:
+          credentials_json: ${{ secrets.GCP_SERVICE_ACCOUNT_JSON }}
+      - name: Setup gcloud
+        uses: google-github-actions/setup-gcloud@v2
+      - run: gcloud auth configure-docker -q
+      - run: ./scripts/build-and-push-images.sh
+        # TODO: this should pull the binary that was built in 'build'
```
tychoish, line 664: actually this might not be true because we want to use a release build and we build a debug build in test. 

@scsmithr we _could_ make the build we test a release build but then the SLTs wouldn't get overflow checking and similar. 

there might also be less cache reuse.

.github/workflows/ci.yaml
```diff
@@ -653,3 +640,25 @@ jobs:
           [ "${{matrix.settings.name}}" = "Sqlserver" ] && exit 0
 
           just slt-bin --protocol=flightsql --exclude '*/tunnels/ssh' ${{matrix.settings.path}}
+
+  docker-push:
+    name: Build and Push Docker Image
+    if: github.event_name == 'push' && github.repository == 'GlareDB/glaredb' && github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest-8-cores
```
tychoish, line 647: depending on if we decide to reuse the previous artifacts or not we can downside this.

.github/workflows/ci.yaml
```diff
@@ -653,3 +640,25 @@ jobs:
           [ "${{matrix.settings.name}}" = "Sqlserver" ] && exit 0
 
           just slt-bin --protocol=flightsql --exclude '*/tunnels/ssh' ${{matrix.settings.path}}
+
+  docker-push:
+    name: Build and Push Docker Image
+    if: github.event_name == 'push' && github.repository == 'GlareDB/glaredb' && github.ref == 'refs/heads/main'
```
tychoish, line 646: the repository check probably doesn't matter, but it's fine.

tychoish: we should, however, make this task depend on... some stuff.

greyscaled: > the repository check probably doesn't matter, but it's fine.

True it would fail in forks, just thought it didn't hurt.

.github/workflows/ci.yaml
greyscaled: I'd be happy to accomplish both. The code as-is was just copy-pasta'd from the release variant, but since this runs on every push to `main` and is generally OK to take longer, I'm happy to downsize.

.github/workflows/ci.yaml
greyscaled: > actually this might not be true because we want to use a release build and we build a debug build in test

fair point, let's remove the comment

greyscaled, line 664:
```suggestion
```

PR #2654, opened 2024-02-15
feat: Initial globbing ranges support

vrongmeal:
We support two kinds of glob ranges:

* `{N..M}`: Here both `N` and `M` are valid integers and (`N <= M`).
* `{str1,str2,str3}`: Three different strings: `"str1", "str2", "str3"`

**Example:**

```
> select * from 'file_{2..3}.csv';
┌─────────┐
│ column1 │
│      ── │
│   Int64 │
╞═════════╡
│       2 │
│       3 │
└─────────┘
> select * from 'file{_1,_3}.csv';
┌─────────┐
│ column1 │
│      ── │
│   Int64 │
╞═════════╡
│       1 │
│       3 │
└─────────┘
```


crates/datasources/src/object_store/glob_util.rs
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    let end = pattern[start..pattern.len()]
+        .find('}')
+        .ok_or_else(|| ReplacerFindError::new("no matching } found"))?;
+    let end = start + end; // Since we sliced to find the end
+
+    let pattern = &pattern[start + 1..end];
+
+    let vec = if pattern.contains(',') {
```
tychoish, line 51: nit, seems like `vec` is a confusing name.

crates/datasources/src/object_store/glob_util.rs
```diff
@@ -0,0 +1,155 @@
...
+#[derive(Debug, Clone)]
+struct Replacer {
+    start: usize,
+    end: usize,
+    iter: IntoIter<String>,
+}
+
+#[derive(Debug, thiserror::Error)]
+#[error("Replacer find error: {0}")]
+struct ReplacerFindError(String);
```
tychoish, line 18: should we use an enum/this error?

vrongmeal: Any benefits of doing that?

tychoish: just that we do enums in lots of places. it's fine if that makes sense.

vrongmeal: We just want the error for debug log. All errors here are ignored, so didn't seem useful to have an enum. It's simply an error message.

universalmind303 (approved):

crates/datasources/src/excel/mod.rs
```diff
@@ -52,9 +53,15 @@ impl ExcelTable {
...
-                    .await?;
+
+                let path = source_url.path().into_owned();
+                let paths = get_resolved_patterns(path);
+
+                let mut list = Vec::new();
+                for path in paths {
+                    let sub_list = store_access.list_globbed(&store, path).await?;
+                    list.extend(sub_list);
+                }
```
tychoish, line 65: does it make sense to have a helper function to wrap all this up into a single operation? we seem to do it three times...

crates/datasources/src/object_store/mod.rs
```diff
@@ -198,10 +201,14 @@ pub trait ObjStoreAccess: Debug + Display + Send + Sync {
...
-            let list = self
-                .list_globbed(&store, &loc.path())
-                .await
-                .map_err(|e| DataFusionError::External(Box::new(e)))?;
+            let mut list = Vec::new();
+            for path in get_resolved_patterns(loc.path().into_owned()) {
+                let sub_list = self
+                    .list_globbed(&store, path)
+                    .await
+                    .map_err(|e| DataFusionError::External(Box::new(e)))?;
```
tychoish, line 209: It looks like `ObjectStoreSourceError` should have a `From<>` implemented for DataFusionError, so I would have thought that we wouldn't have needed the map here anymore?

tychoish (approved):
some nits.

I'd also sort of like to see the tests also do BSON because the object store integration is a bit different, but I'm fine if we don't, I think.

crates/datasources/src/excel/mod.rs
vrongmeal: Yeah, we should have a helper function do this.

crates/datasources/src/object_store/mod.rs
vrongmeal: I usually avoid implementing from for different enum errors since a DF error might be `External` or `Execution` etc. Mapping feels fine and explicit to me.

crates/datasources/src/object_store/glob_util.rs
```diff
@@ -0,0 +1,155 @@
...
+            ),
+        ];
+
+        for (pattern, expected) in test_cases {
+            let patterns = get_resolved_patterns(pattern.to_string());
+            let actual = patterns.iter().map(AsRef::as_ref).collect::<Vec<_>>();
+            assert_eq!(&actual, expected);
+        }
+
+        // TODO: Add test cases for failed patterns...
```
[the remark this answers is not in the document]
vrongmeal: Added more cases to handle the cases you mentioned.

I am not very keen on throwing an error since the "pattern" formation can be coincidental and we shouldn't throw an error until user opts-in or opts-out.

We definitely should have an "opt-out" from glob pattern so we can throw errors with confidence.

tychoish (approved):

Issue #2651, opened 2024-02-15
Feature: Inference for Excel file extensions


talagluck:
## Description
Currently, for CSV files, both of these will work
```sql
SELECT * FROM read_csv('my_csv.csv');

SELECT * FROM 'my_csv.csv';
```

For Excel files, only this will work
```sql
SELECT * FROM read_excel('my_excel.xlsx')
```

If you instead run
```sql
SELECT * FROM 'my_excel.xlsx'
``` 
you get

```
ExecutionException: Error during planning: unable to infer how to handle file extension: xlsx
```

greyscaled:
Partially related: #2639 via #2602

PR #2655, opened 2024-02-15
fix: infer excel files by extension

tychoish:
addresses #2651


crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -261,6 +261,10 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
             schema: "public".into(),
             table: "read_bson".into(),
         },
+        "xslx" => OwnedTableReference::Partial {
```
greyscaled, line 264:
```suggestion
        "xlsx" => OwnedTableReference::Partial {
```

greyscaled (approved):

PR #2668, opened 2024-02-16
feat: read_google_sheets table function

tychoish:
I got pretty far yesterday, but would like to get more eyes on how to do more of this. 

I think some the challenge is that the object store reads (part? all?)
of the CSV file twice (once for schema inference and once for the
actual load, and I _think_ if we read the file into memory entirely
then we won't end up downloading it twice, but then it'll all be in
memory, and that also seems bad. 

Not sure what the right thing to do here is, probably pulling the data
twice is ok.

Would be nice to have some folks spot this and see if there's
something I've missed.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1455,7 +1455,6 @@ version = "0.8.5"
 dependencies = [
  "datafusion",
  "logutil",
- "metastore",
  "object_store",
  "parking_lot",
  "protogen",
@@ -2438,6 +2437,7 @@ dependencies = [
  "env_logger",
  "futures",
  "ioutil",
+ "object_store",
  "once_cell",
  "parking_lot",
  "paste",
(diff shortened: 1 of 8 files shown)


universalmind303 (commented):
If we use the google sheets api directly, it gives us much more control over treating it as a csv file. The google apis allow for projection and slice pushdowns as well. 

[ROAPI](https://github.com/roapi/roapi) does this, so that may be a good place to take a look. At my previous employer we did this too on a closed source federation engine. We also used the google apis directly instead of treating it as a CSV.

tychoish:
That's a good reminder. Somewhere along this journey I tried playing with the API but it was a bigger lift than I was expecting this to be, but the object store is not great for this use case. 

It definitely would be good to get _some way_ of being able to support this natively and in the shorter rather than longer term.

PR #2670, opened 2024-02-16
feat: `cloud_upload(<filename>)` table func

greyscaled:
## Summary

When a file is uploaded via [GlareDB Cloud](https://console.glaredb.com), it can be read with `cloud_upload(<filename>)`.

This function can only be called in remote and hybrid contexts; it is not meaningful locally.

The `filename` must contain a valid extension (`foo.csv`).

### Example

After uploading a file called `test.csv` to [GlareDB Cloud](https://console.glaredb.com), it can be accessed like:

```sql
SELECT * FROM cloud_upload('test.csv');
```

and joined with other data sources, etc.

## Details

<details><summary>MVP File Format Limitations</summary>
<p>

Presently only:

- CSV as `filename.csv`
- Parquet as `filename.parquet
- Newline-delimited JSOn as `filename.json`

are accepted.

The file name must currently include the extension.

The file can only use alphanumeric characters.

</p>
</details> 

<details><summary>Testing</summary>
<p>

I tested this by building the image and running with a GlareDB Cloud environment.
I think it's tractable to write an integration test for this function, in the same vein
as other integration tests, but I'd prefer to leave this as a follow-up refinement
item.

</p>
</details> 

Resolves: #2614


.github/workflows/ci.yaml
```diff
@@ -629,10 +629,9 @@ jobs:
 
           just slt-bin --protocol=flightsql ${{matrix.settings.path}}
 
-
   docker-push:
     name: Build and Push Docker Image
-    if: github.event_name == 'push' && github.repository == 'GlareDB/glaredb' && github.ref == 'refs/heads/main'
+    # if: github.event_name == 'push' && github.repository == 'GlareDB/glaredb' && github.ref == 'refs/heads/main'
```
greyscaled, line 634: TODO(before merge): remove diff in this file

crates/sqlbuiltins/src/functions/table/object_store.rs
```diff
@@ -582,3 +585,122 @@ fn create_azure_store_access(
...
+        let base_url = ObjectStoreUrl::parse(base_url)?;
+
+        // glaredb_upload currently does not support globbing, and therefore we
+        // do not need to iterate: there should only be one.
+        let mut objects = store.list(Some(&prefix));
+        let meta = objects
+            .next()
+            .await
+            .ok_or_else(|| ExtensionError::String(format!("file not found: {}", file_name)))?
+            .or_else(|_| return Err(ExtensionError::ObjectStore("read file".to_string())))?;
```
greyscaled, line 690: this will fail clippy, I believe. Need to figure out canonical way to handle an `Option<Result>`

crates/sqlbuiltins/src/functions/table/object_store.rs
```diff
@@ -582,3 +585,125 @@ fn create_azure_store_access(
...
+            .next()
+            .await
+            .ok_or_else(|| ExtensionError::String(format!("file not found: {}", file_name)))?
+        {
+            Ok(meta) => meta,
+            Err(e) => return Err(ExtensionError::ObjectStore(e.to_string())),
+        };
+        let objects: Vec<ObjectMeta> = vec![meta];
+
+        // Infer schema
```
greyscaled, line 696:
```suggestion
```

crates/sqlbuiltins/src/functions/table/object_store.rs
```diff
@@ -582,3 +585,126 @@ fn create_azure_store_access(
...
+
+impl ConstBuiltinFunction for GlareDBUpload {
+    const NAME: &'static str = "glaredb_upload";
+    const DESCRIPTION: &'static str = "Reads a file that was uploaded to GlareDB Cloud.";
+    const EXAMPLE: &'static str = "SELECT * FROM glaredb_upload('my_upload.csv')";
+    const FUNCTION_TYPE: FunctionType = FunctionType::TableReturning;
+
+    // signature for GlareUpload is a single filename. The filename may
+    // optionally contain an extension, though it is not required. Filename
+    // should not be a path.
```
greyscaled, line 600:
```suggestion
    // signature for GlareUpload is a single filename. The filename must
    // contain an extension. Filename should not be a path.
```

crates/datasources/src/native/access.rs
```diff
@@ -56,12 +58,15 @@ pub struct NativeTableStorage {
     /// what this type is for.
     ///
     /// Arcs all the way down...
-    store: SharedObjectStore,
+    pub store: SharedObjectStore,
```
greyscaled, line 61: q: should I provide a getter that performs a clone? (does that make sense from mem model)?

There's a comment here that describes that this object was created to be cloned:

https://github.com/GlareDB/glaredb/blob/a1787da8c70311782305ba7b5f6ce0997f370c58/crates/object_store_util/src/shared.rs#L28-L33

tychoish: I think the compiler does the thing that you would want it to do, so no need to get clever.

crates/datasources/src/native/access.rs
```diff
@@ -41,11 +41,13 @@ use crate::native::insert::NativeTableInsertExec;
...
+
+    /// URL pointing to the bucket and/or directory which is the root of the
+    /// native storage, for example `gs://<bucket-name>`.
     ///
-    /// In other words this is the location to which the the table prefix is applied to get
-    /// a full table URL.
-    root_url: Url,
+    /// In other words this is the location to which the the table prefix is
+    /// applied to get a full table URL.
+    pub root_url: Url,
```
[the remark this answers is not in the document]
tychoish: I wouldn't unless you need/want to modify it for callers/users of this API.

tychoish:
this all seems good, and I'd be fine merging it.

- we could probably write tests for this using mino or the mock gcs fixture, but it'd be gnarly. Shouldn't do that today. 
- before we merge we should definitely go around and make sure the name is good.
- want to make sure that it errors correctly, when running in pure local or embedded.

greyscaled:
> want to make sure that it errors correctly, when running in pure local

It does and we can easily add that test, I see no harm in it.

tychoish:
> > 
> 
> It does and we can easily add that test, I see no harm in it.

I was moving too quick mostly meant: "a really good error message" and not, say ("credentials not found") or something else.

.github/workflows/ci.yaml
greyscaled, line 634: I reverted this just to run another test with recent code changes. Will remove after image builds.

tychoish (approved):

Issue #2650, opened 2024-02-15
Bug: Support for headers in Excel


talagluck:
## Description
Currently, if you run
```
select * from read_excel('my_excel_file.xlsx')
```
the first row, which is a header row, is not treated as a header, and the column names are instead just auto-generated.

If you [download this file](https://docs.google.com/spreadsheets/d/1f5epAPxP_Yd3g1TunEMdtianpVAhKS0RG6BKRDSLtrk/edit#gid=0) as an .xlsx, you will see that this is the case. If you instead download it as a csv and run
```
select * from read_csv('my_csv_file.csv')
```
then the headers will be appropriately inferred.

tychoish:
there is a `has_header` option in the `read_excel` function. 

The CSV code is largely in datafusion so hard to duplicate. and I'm not sure how they decide if the first row is a header or not. 

We could invert the default of this value if you think that's best/expected.

greyscaled:
One issue here is that the `has_header` is not documented in https://docs.glaredb.com/glaredb/sql-functions/read_excel/

I recall holding off on documenting it because there was a big initially: https://github.com/GlareDB/glaredb/issues/2372

greyscaled:
@talagluck in the meantime, we actually have working SLTs that use the header option, see here:

https://github.com/GlareDB/glaredb/commit/62f8bb8c463a5f4b62fc92ffd75c0e0963380d3e#diff-4e3456fcfa82f06b69190a3583ffac9a46e794da0cd45b2796d9a7d1be4aae56

tychoish:
The excel tests are dodgy, but I added a few tests which confirm (more or less) that it works.

talagluck:
Thanks all! I'll keep this open until we document (which we can roll into the current docs push)

talagluck:
I do think we should reverse the default here - I have a hard time imagining it being useful in the current circumstances, and it's always weird that it's different from `read_csv`.

greyscaled:
> I do think we should reverse the default here

+1

PR #2672, opened 2024-02-16
fix: excel files should default to having a header like csv files

tychoish:
Addresses #2650


tychoish:
> We should add a test case for `has_header = false`

https://github.com/GlareDB/glaredb/pull/2672/files#diff-4e3456fcfa82f06b69190a3583ffac9a46e794da0cd45b2796d9a7d1be4aae56R21

That exists.

Issue #2676, opened 2024-02-21
Bug: Enable altering views through the ALTER TABLE function


talagluck:
## Description
I'm hitting an issue with dbt where it works fine for tables but panics when it hits views.

I just did a round of debugging with @scsmithr and it seems like the issue is that it's running an ALTER TABLE function to rename a view. Looking for two things here:
1. ALTER TABLE should work on views for renaming views
2. Don't panic

scsmithr:
Specifically panicking here: https://github.com/glaredb/glaredb/blob/db252ce59d066e073b40be0d30eecc30f61a130e/crates/metastore/src/database.rs#L938-L941

greyscaled:
I can look into this one today after #2670 merges if no one else is picking it up

PR #2678, opened 2024-02-21
fix: Alter table renames view successfully

vrongmeal:
Fixes: #2676

### DIFF
--- a/crates/datafusion_ext/src/planner/relation/mod.rs
+++ b/crates/datafusion_ext/src/planner/relation/mod.rs
@@ -262,6 +262,10 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
             schema: "public".into(),
             table: "read_bson".into(),
         },
+        "xlsx" => OwnedTableReference::Partial {
+            schema: "public".into(),
+            table: "read_excel".into(),
+        },
         ext => {
             if let Ok(compression_type) = ext.parse::<FileCompressionType>() {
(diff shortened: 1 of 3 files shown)


crates/datafusion_ext/src/planner/relation/mod.rs
```diff
@@ -250,10 +250,6 @@ fn infer_func_for_file(path: &str) -> Result<OwnedTableReference> {
             schema: "public".into(),
             table: "read_parquet".into(),
         },
-        "xlsx" => OwnedTableReference::Partial {
-            schema: "public".into(),
-            table: "read_excel".into(),
-        },
```
vrongmeal, line 256: Redundant case, handled again below: Just a fix for compiler warning.

greyscaled: this was merged to main here: https://github.com/GlareDB/glaredb/commit/770f4e4a0122c562332525db38e03122556d8b14

scsmithr (approved):

crates/datafusion_ext/src/planner/relation/mod.rs
greyscaled: @vrongmeal we need to undo this diff now that you've merged main, as it will be deleting both arms

vrongmeal: Yup! I'll delete this.

tychoish (approved):

PR #2680, opened 2024-02-21
feat: `cosine_similarity` function

universalmind303:
adds new function `cosine_similarity` for comparing similarity across floating point vectors. `cosine_similarity` is commonly used in vector search workloads.


crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -217,6 +218,8 @@ impl FunctionRegistry {
             Arc::new(PartitionResults),
             // OpenAI
             Arc::new(OpenAIEmbed),
+            // Similary
```
greyscaled, line 221:
```suggestion
            // Similarity
```

crates/sqlbuiltins/src/functions/scalars/similarity.rs
```diff
@@ -0,0 +1,309 @@
...
+                query_vec.as_ref(),
+                &target_vec,
+            )
+            .map_err(|e| DataFusionError::Execution(e.to_string()))?;
+
+            Ok(ColumnarValue::Array(result))
+        });
+
+        let return_type_fn: ReturnTypeFunction = Arc::new(move |_| {
+            let dtype = DataType::Float32;
```
tychoish, line 193: is there a compelling reason to use single precision floats?

crates/sqlbuiltins/src/functions/scalars/similarity.rs
```diff
@@ -0,0 +1,309 @@
...
+            args,
+        )))
+    }
+}
+
+// modified/copied from arrow_cast
+fn cast_list_to_fixed_size_list<OffsetSize>(
+    array: &GenericListArray<OffsetSize>,
+    field: &FieldRef,
+    size: i32,
```
tychoish, line 216: why not usize?

crates/sqlbuiltins/src/functions/scalars/similarity.rs
```diff
@@ -0,0 +1,309 @@
...
+    let array = FixedSizeListArray::new(field.clone(), size, values, nulls);
+    Ok(array)
+}
+
+
+// modified copy from arrow_cast
+fn cast_fsl_inner(
+    array: &FixedSizeListArray,
+    field: &FieldRef,
+    size: i32,
```
tychoish, line 288: should probably be usize as well.

testdata/sqllogictests/functions/cosine_similarity.slt
```diff
@@ -0,0 +1,45 @@
+# tests for cosine similarity function
```
tychoish, line 1: I think we should probably also have some small collection of tests  where the input numbers (to the table, the vector, the target) that are negative, have non-zero decimal point values, and are less than one, and numbers that would overflow a single precision float.

crates/sqlbuiltins/Cargo.toml
```diff
@@ -33,3 +33,7 @@ memoize = { version = "0.4.2", features = ["full"] }
 async-openai = "0.18.3"
 tokio.workspace = true
 reqwest.workspace = true
+# Important to keep this in sync with the datafusion arrow-cast version
+arrow-cast = { version = "50.0.0" }
```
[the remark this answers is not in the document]
universalmind303: datafusion doesnt directly export arrow-cast.

crates/sqlbuiltins/src/functions/scalars/similarity.rs
universalmind303: fixedsizelist datatype expects i32. Using usize would require unnecessary casts inside this function. 
This is just copy/pasted from arrow-cast with the dyn array output replaced with fixedsizearray.

crates/sqlbuiltins/src/functions/scalars/similarity.rs
universalmind303: same comment

crates/sqlbuiltins/src/functions/scalars/similarity.rs
universalmind303: we're just a thin wrapper over lance, we are just aligning with their inputs/outputs.

testdata/sqllogictests/functions/cosine_similarity.slt
universalmind303: I think those would probably be best tested inside lance directly. We are just a thin wrapper over lance's implementation, so if there are issues like the ones you mention, I'd expect that those should likely be opened and handled upstream.

tychoish (changes requested):

testdata/sqllogictests/functions/cosine_similarity.slt
universalmind303: Yeah im going to have to fundamentally disagree here. We shouldn't be testing the function itself, and it's generally a bad practice to essentially unit test external functions. 

We test the **integration**, if the function has bugs such as handling of negative values, then that should be taken upstream. We add a few tests like the ones here for testing the integration with our own system. We want to test **OUR** logic, generally we assume that external libraries code is well tested and if bugs arise we open it up with them, not test it internally.

--- 
Ill add some tests just to get this approved, but I **strongly** disagree with the decision to add them as i think it is not good practice to test external functions like this.

universalmind303:
> I tried it out, and it worked nicely.
> 
> I don't fully understand why two equivalent vectors gives me 5.9604645e-8 instead of 0.0 (as shown in the tests), but I understand that this is upstream behaviour and I lack some prior art/experience.

yeah i think this is just a math/precision thing when performing similarity. `0.000000059604645` is practically 0.0

tychoish (approved):

tychoish (approved):

universalmind303:
@tychoish the CI seems to keep getting stuck on `build cache`. I don't think this is related to this PR at all.


https://github.com/GlareDB/glaredb/actions/runs/8008291239/job/21874514119?pr=2680

Issue #2549, opened 2024-01-31
`\timing` does not display timings for ddl commands like `COPY`


universalmind303:
## Description
```sql
> \timing 
Timing is on

> COPY (SELECT * FROM bq.austin_bikeshare.bikeshare_trips limit 1) TO './bikeshare_trips.csv';
Copy success
```

## Expected

Timings are shown for `COPY` statements.

PR #2681, opened 2024-02-21
fix: report time display for DDL commands

hemanth94:
Resolves: #2549 

This shows time elapsed where it matters, For commands like COPY, INSERT, DELETE  and UPDATE.

Please suggest changes if any.
Thanks


tychoish (approved):

universalmind303 (approved):
thanks again @hemanth94.

Issue #2644, opened 2024-02-14
`CREATE EXTERNAL TABLE` support for excel


tychoish:
Currently Excel file reading is only available as a table function, but we could support `CREATE EXTERNAL TABLE` from excel files as we do for other formats. 

The code to do this is similar to the code for the other formats and there's nothing that is special about excel files, it's just a matter of hooking it all up and writing a few tests.

greyscaled:
Do I have it correct that we would start with `create external table` which would specify a `sheet`, and that we could later add `create external table` that maps all sheets to tables?

hemanth94:
@tychoish Can i try contributing to this? 
Will ask for suggestions if needed.

Thanks,

tychoish:
@hemanth94 this would be great for you! I cannot imagine that there'll be anything unusual. the `bson` example, which I did recently, is pretty clean and might be a good thing to look at for a model.

Issue #2645, opened 2024-02-14
support reading excel files from object storage.


tychoish:
Could use the GenericStoreAccess as [here](https://github.com/glaredb/glaredb/blob/main/crates/sqlbuiltins/src/functions/table/bson.rs#L52-L79).

hemanth94:
Hello @tychoish 
I think, adding Object_Store support to excel may not be as simple as you assumed.

Since the  [code here](https://github.com/GlareDB/glaredb/blob/9b8483fb5a1cd63476efc5132468378e30323587/crates/datasources/src/excel/mod.rs#L156) written to read excel is only applicable for local files. 

I didn't find any code similar to [Bson Object_Store implementation](https://github.com/GlareDB/glaredb/blob/9b8483fb5a1cd63476efc5132468378e30323587/crates/datasources/src/bson/table.rs#L50-L70) written which i could replicate to excel.

Please correct me if I'm missing something. Can i try writing code for this?

Thanks,

tychoish:
This was done in #2684

PR #2684, opened 2024-02-22
feat: support for `create external table excel <options>`

hailelagi:
closes https://github.com/GlareDB/glaredb/issues/2644, summary of changes:
- implements the `create external table excel <options>` sql command by adding a "streaming table" interface that maps excel rows/columns into a `RecordBatch`.

this will likely be superseded with a followup after the `TableProvider` is implemented for (https://github.com/GlareDB/glaredb/issues/2645) to allow querying on tables and it's missing support for streaming multiple sheets from a worksheet, it's also not super optimal quite yet, but happy to iterate on it on this pr or a followup.

### DIFF
--- a/crates/datasources/src/excel/errors.rs
+++ b/crates/datasources/src/excel/errors.rs
@@ -1,0 +1,70 @@
+use std::fmt::Debug;
+
+use calamine::{self, XlsxError};
+use datafusion::arrow::error::ArrowError;
+use datafusion::error::DataFusionError;
+use datafusion_ext::errors::ExtensionError;
+
+use crate::object_store::errors::ObjectStoreSourceError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ExcelError {
+    #[error("Failed to load XLSX: {0}")]
+    Load(String),
+    #[error("Failed to create record batch: {0}")]
+    CreateRecordBatch(#[from] ArrowError),
+    #[error(transparent)]
+    CalamineError(#[from] calamine::XlsxError),
+    #[error(transparent)]
+    ObjectStoreError(#[from] object_store::Error),
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CalamineError {
+    #[error("Failed to open workbook: {0}")]
+    OpenWorkbook(#[from] calamine::XlsxError),
+}
+
+
+impl From<calamine::Error> for ExcelError {
+    fn from(error: calamine::Error) -> Self {
+        ExcelError::Load(error.to_string())
+    }
+}
+
+
+impl From<ExcelError> for DataFusionError {
+    fn from(e: ExcelError) -> Self {
+        DataFusionError::Execution(e.to_string())
+    }
+}
+
+impl From<CalamineError> for ExcelError {
+    fn from(error: CalamineError) -> Self {
+        ExcelError::Load(error.to_string())
+    }
+}
+
+impl From<ObjectStoreSourceError> for ExcelError {
+    fn from(error: ObjectStoreSourceError) -> Self {
+        ExcelError::Load(error.to_string())
+    }
+}
+
+impl From<std::io::Error> for ExcelError {
+    fn from(error: std::io::Error) -> Self {
+        ExcelError::Load(error.to_string())
+    }
+}
+
+impl From<ExcelError> for calamine::XlsxError {
+    fn from(e: ExcelError) -> Self {
+        XlsxError::FileNotFound(e.to_string())
+    }
+}
+
+impl From<ExcelError> for ExtensionError {
+    fn from(e: ExcelError) -> Self {
+        ExtensionError::String(e.to_string())
+    }
+}
--- a/crates/datasources/src/excel/mod.rs
+++ b/crates/datasources/src/excel/mod.rs
@@ -1,89 +1,109 @@
 use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
+use std::io::Cursor;
 use std::path::PathBuf;
 use std::sync::Arc;
 
-use calamine::{open_workbook, DataType as CalamineDataType, Range, Reader, Xlsx};
+use calamine::{open_workbook, DataType as CalamineDataType, Range, Reader, Sheets, Xlsx};
 use datafusion::arrow::array::{ArrayRef, BooleanArray, Date64Array, PrimitiveArray, StringArray};
 use datafusion::arrow::datatypes::{DataType, Field, Float64Type, Int64Type, Schema};
 use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::datasource::MemTable;
+use object_store::ObjectStore;
 
-#[derive(Debug, thiserror::Error)]
-pub enum Error {
-    #[error("Failed to load XLSX: {msg}")]
-    Load { msg: String },
-    #[error("Failed to create record batch: {0}")]
-    CreateRecordBatch(#[from] datafusion::arrow::error::ArrowError),
-    #[error("Failed to open workbook: {0}")]
-    OpenWorkbook(#[from] calamine::XlsxError),
-}
+use crate::common::url::DatasourceUrl;
+use crate::object_store::generic::GenericStoreAccess;
+use crate::object_store::ObjStoreAccess;
 
-fn infer_value_type(v: &calamine::Data) -> Result<DataType, Error> {
-    match v {
-        calamine::Data::Int(_) => Ok(DataType::Int64),
-        calamine::Data::Float(_) => Ok(DataType::Float64),
-        calamine::Data::Bool(_) => Ok(DataType::Boolean),
-        calamine::Data::String(_) => Ok(DataType::Utf8),
-        calamine::Data::Error(e) => Err(Error::Load { msg: e.to_string() }),
-        calamine::Data::DateTime(_) => Ok(DataType::Date64),
-        calamine::Data::Empty => Ok(DataType::Null),
-        _ => Err(Error::Load {
-            msg: "Failed to parse the cell value".to_owned(),
-        }),
-    }
-}
+pub mod errors;
+pub mod stream;
+pub mod table;
 
-fn infer_schema(
-    r: &Range<calamine::Data>,
+use errors::ExcelError;
+
+pub struct ExcelTable {
+    cell_range: Range<calamine::Data>,
     has_header: bool,
-    infer_schema_length: usize,
-) -> Result<Schema, Error> {
-    let mut col_types: HashMap<&str, HashSet<DataType>> = HashMap::new();
-    let mut rows = r.rows();
-    let col_names: Vec<String> = rows
-        .next()
-        .unwrap()
-        .iter()
-        .enumerate()
-        .map(
-            |(i, c)| match (has_header, c.get_string().map(|s| s.to_string())) {
-                (true, Some(s)) => Ok(s),
-                (true, None) => Err(Error::Load {
-                    msg: "failed to parse header".to_string(),
-                }),
-                (false, _) => Ok(format!("col{}", i)),
-            },
-        )
-        .collect::<Result<_, _>>()?;
+}
 
-    for row in rows.take(infer_schema_length) {
-        for (i, col_val) in row.iter().enumerate() {
-            let col_name = col_names.get(i).unwrap();
-            let col_type = infer_value_type(col_val).unwrap();
-            if col_type == DataType::Null {
-                continue;
+impl ExcelTable {
+    pub async fn open(
+        store_access: GenericStoreAccess,
+        source_url: DatasourceUrl,
+        sheet_name: Option<&str>,
+        has_header: bool,
+    ) -> Result<ExcelTable, ExcelError> {
+        match source_url {
+            DatasourceUrl::File(path) => {
+                let path = ioutil::resolve_path(&path)?;
+                let mut sheet = calamine::open_workbook_auto(path)?;
+
+                let first_sheet = sheet_name.map(Cow::Borrowed).unwrap_or_else(|| {
+                    let sheets = sheet.sheet_names();
+                    let first = sheets.first().expect("file has a sheet");
+                    Cow::Owned(first.clone())
+                });
+
+                Ok(ExcelTable {
+                    cell_range: sheet.worksheet_range(&first_sheet)?,
+                    has_header,
+                })
             }
-            let entry = col_types.entry(col_name).or_default();
-            entry.insert(col_type);
+
+            DatasourceUrl::Url(_) => {
+                let store = store_access.create_store()?;
+                let list = store_access
+                    .list_globbed(&store, source_url.path().as_ref())
+                    .await?;
+
+                if list.is_empty() {
+                    return Err(ExcelError::Load(
+                        "could not find .xlsx file at remote".to_string(),
+                    ));
+                };
+
+                let meta = list.first().expect("remote file has a sheet");
+                let bs = store.get(&meta.location).await?.bytes().await?;
+
+                let buffer = Cursor::new(bs);
+                let mut sheets: Sheets<_> = calamine::open_workbook_auto_from_rs(buffer).unwrap();
+
+                let first_sheet = sheet_name.map(Cow::Borrowed).unwrap_or_else(|| {
+                    let sheets = sheets.sheet_names();
+                    let first = sheets.first().unwrap();
+                    Cow::Owned(first.clone())
+                });
+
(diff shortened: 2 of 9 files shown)


crates/datasources/src/excel/errors.rs
```diff
@@ -0,0 +1,26 @@
+use calamine;
+use datafusion::arrow::error::ArrowError;
+use datafusion::error::DataFusionError;
+use datafusion_ext::errors::ExtensionError;
+
+#[derive(Debug, thiserror::Error)]
+pub enum ExcelError {
+    #[error("Failed to load XLSX: {msg}")]
+    Load { msg: String },
```
tychoish, line 9: I'd have this be `Load(String)`

crates/datasources/src/excel/errors.rs
```diff
@@ -0,0 +1,26 @@
...
+    #[error("Failed to create record batch: {0}")]
+    CreateRecordBatch(#[from] ArrowError),
+    #[error("Failed to open workbook: {0}")]
+    OpenWorkbook(#[from] calamine::XlsxError),
```
tychoish, line 13: Have generic `CalamineError` and `ArrowError`, with specifics for these "create record batch and Open Workbook cases"

crates/datasources/src/excel/stream.rs
```diff
@@ -0,0 +1,70 @@
...
+use datafusion::physical_plan::{RecordBatchStream, SendableRecordBatchStream};
+use futures::{Stream, StreamExt};
+
+use crate::excel::errors::ExcelError;
+
+pub type SendableSheetStream = Pin<Box<dyn Stream<Item = Result<RecordBatch, ExcelError>> + Send>>;
+
+pub struct ExcelStream {
+    schema: Arc<Schema>,
+    stream: Pin<Box<dyn Stream<Item = Result<RecordBatch, ExcelError>> + Send>>,
```
tychoish, line 19: does the `SendableRecordBatchStream` alias cover this type?

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+pub async fn excel_streaming_table(
+    _store_access: GenericStoreAccess,
+    source_url: DatasourceUrl,
+    sheet_name: Option<&str>,
+    has_header: bool,
+    infer_schema_length: usize,
+) -> Result<Arc<dyn TableProvider>, ExcelError> {
+    let mut workbook: Xlsx<_> = match source_url {
+        DatasourceUrl::File(path) => open_workbook(path)?,
+        DatasourceUrl::Url(_) => todo!("remote files not supported yet"),
```
tychoish, line 25: we've talked about this in other contexts, but I think it's pretty straight forward at this point.

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    // for now, this only infers table data from only the first sheet in a workbook
+    // when read_excel gets vectorised to real columns this should be updated
+    let sheet = sheet_name.map(Cow::Borrowed).unwrap_or_else(|| {
```
tychoish, line 30: I'm not sure that we're getting much out of the `Cow` wrapping, but I suppose it's not hurting anyone here.

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+        let first = sheets.first().expect("sheet is not empty");
+        Cow::Owned(first.clone())
+    });
+
+    let working_sheet_range = match workbook.worksheet_range(&sheet) {
+        Ok(r) => Ok(r),
+        Err(_) => Err(ExcelError::Load {
+            msg: "Failed to open .xlsx file.".to_owned(),
+        }),
+    };
```
tychoish, line 41: can't we replace the match here, with `.map_err(|e| <...>)?`

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let range = working_sheet_range?;
+    let schema = excel::infer_schema(&range, has_header, infer_schema_length)?;
+
+    // todo: expand this handle multiple sheets
```
tychoish, line 46: capitalize TODO; 

I think there's a lot of possible options here for dealing with multiple sheets, but the one that sort of makes the most sense to me is to treat sheets as tables and workbooks as databases. So we might have a different implementation that would handle that.

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+    let mut streams = Vec::<Arc<(dyn PartitionStream + 'static)>>::with_capacity(1);
+
+    streams.push(Arc::new(ExcelSheetPartitionStream::new(
+        Arc::new(schema.clone()),
+        Box::pin(futures::stream::iter(vec![
+            xlsx_sheet_value_to_record_batch(range, has_header, infer_schema_length),
+        ])),
+    )));
```
tychoish, line 54: why not make this just `vec![Arc::new(ExcelSheetPartitionStream::new()]`?

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+
+    let table_stream = StreamingTable::try_new(Arc::new(schema.clone()), streams);
+
+    match table_stream {
+        Ok(streaming_table) => Ok(Arc::new(streaming_table)),
+        Err(_) => Err(ExcelError::Load {
+            msg: "Failed to map to an excel table".to_owned(),
+        }),
+    }
```
tychoish, line 63: you can wrap the table_stream in an `Arc` on line 56, and do this with a `.map_err()?`

testdata/sqllogictests/xlsx.slt
```diff
@@ -57,3 +57,21 @@ select "HEADING" from read_excel('file://${PWD}/testdata/xlsx/multiple_sheets.xl
...
 statement error
 select * from read_excel('file://${PWD}/testdata/xlsx/multiple_sheets.xlsx', sheet_name => 'other', infer_rows => -1);
+
+# create external table
+
+statement ok
+create schema 'xlsx';
+
+statement ok
+set search_path = 'xlsx';
```
tychoish, line 67: is there a reason to do this in a different schema?

testdata/sqllogictests/xlsx.slt
```diff
@@ -57,3 +57,21 @@ select "HEADING" from read_excel('file://${PWD}/testdata/xlsx/multiple_sheets.xl
...
+
+statement ok
+create external table basic_report from excel options( location='./testdata/xlsx/userdata1.xlsx', file_type = 'xlsx');
+
+# implicitly create
```
tychoish, line 72: this seems explicit to me?

crates/datasources/src/excel/table.rs
```diff
@@ -0,0 +1,64 @@
...
+use datafusion::physical_plan::streaming::PartitionStream;
+
+use super::xlsx_sheet_value_to_record_batch;
+use crate::common::url::DatasourceUrl;
+use crate::excel;
+use crate::excel::errors::ExcelError;
+use crate::excel::stream::ExcelSheetPartitionStream;
+use crate::object_store::generic::GenericStoreAccess;
+
+pub async fn excel_streaming_table(
```
tychoish, line 16: so I really like the streaming table model that DF provides, _but_ they are pretty simple, and offer the greatest amount of benefit when there are multiple files (streams) or very large files that can't be in memory for whatever reason. 

In the case of these excel files, there's only ever going to be one file, and I think all or most of it has to be in memory to execute the query, so we might be able to get better results by:
writing our own query plan manually, take advantage of pushdown operations, and support more optimization in the future.

testdata/sqllogictests/xlsx.slt
```diff
@@ -57,3 +57,21 @@ select "HEADING" from read_excel('file://${PWD}/testdata/xlsx/multiple_sheets.xl
...
+create schema 'xlsx';
+
+statement ok
+set search_path = 'xlsx';
+
+statement ok
+create external table basic_report from excel options( location='./testdata/xlsx/userdata1.xlsx', file_type = 'xlsx');
+
+# implicitly create 
+statement ok
```
tychoish, line 73: seems like it'd be ok to test some of the error cases/ or edge cases just to demonstrate that we get errors (safely) when provided with bad inputs.

tychoish (commented):
comments.

crates/datasources/src/excel/table.rs
hailelagi: I kept going back and forth on the mapping here and couldn't decide, even if there were multiple sheets supported it'd all be read into memory anyway by calamite, so yes agree that extending at the query plan makes sense.

crates/datasources/src/excel/stream.rs
hailelagi: I believe it _should_ given:
```rust
impl From<ExcelError> for DataFusionError {
    fn from(e: ExcelError) -> Self {
        DataFusionError::Execution(e.to_string())
    }
}
```

crates/datasources/src/excel/table.rs
hailelagi: yupI was thinking this too! will likely follow this up with this bit in separate pr

testdata/sqllogictests/xlsx.slt
hailelagi: not really

crates/datasources/src/excel/table.rs
```diff
@@ -4,14 +4,17 @@ use std::sync::Arc;
 use calamine::{open_workbook, Reader, Xlsx};
 use datafusion::datasource::streaming::StreamingTable;
 use datafusion::datasource::TableProvider;
+use datafusion_ext::errors::ExtensionError;
 use datafusion::physical_plan::streaming::PartitionStream;
+// use datafusion::datasource::object_store::ObjectStoreUrl;
```
tychoish, line 9: I would delete this

crates/datasources/src/excel/table.rs
```diff
@@ -22,7 +25,16 @@ pub async fn excel_streaming_table(
 ) -> Result<Arc<dyn TableProvider>, ExcelError> {
     let mut workbook: Xlsx<_> = match source_url {
         DatasourceUrl::File(path) => open_workbook(path)?,
-        DatasourceUrl::Url(_) => todo!("remote files not supported yet"),
+        DatasourceUrl::Url(_) => {
+            /*
+            let store = store_access.create_store().unwrap();
+            let buffer = object_store::buffered::BufReader::with_capacity(store, &meta, infer_schema_length);
+            let sheets = calamine::open_workbook_auto_from_rs(buffer).unwrap();
+             */
```
tychoish, line 33: is this temporary or should we delete it?

crates/datasources/src/excel/table.rs
```diff
@@ -22,7 +25,16 @@ pub async fn excel_streaming_table(
 ) -> Result<Arc<dyn TableProvider>, ExcelError> {
```
tychoish, line 25: It wouldn't be absurd to just convert this type into `ExtensionError` or `DataFusionError` at this point rather than at the call site.

crates/datasources/src/excel/table.rs
```diff
@@ -33,32 +45,26 @@ pub async fn excel_streaming_table(
...
-        Ok(r) => Ok(r),
-        Err(_) => Err(ExcelError::Load {
-            msg: "Failed to open .xlsx file.".to_owned(),
-        }),
-    };
-
-    let range = working_sheet_range?;
+    let range = workbook
+        .worksheet_range(&sheet)
+        .map_err(|e| Into::<ExcelError>::into(e))?;
```
tychoish, line 50: You might be able to get `Ok(workbook.worksheet_range(&sheet)?)` to get the conversion, right? I mgiht be wrong about this case.

crates/datasources/src/excel/table.rs
```diff
@@ -33,32 +45,26 @@ pub async fn excel_streaming_table(
...
-    let mut streams = Vec::<Arc<(dyn PartitionStream + 'static)>>::with_capacity(1);
-
-    streams.push(Arc::new(ExcelSheetPartitionStream::new(
-        Arc::new(schema.clone()),
-        Box::pin(futures::stream::iter(vec![
-            xlsx_sheet_value_to_record_batch(range, has_header, infer_schema_length),
-        ])),
-    )));
+    // TODO: expand this handle multiple sheets
+    let stream: Vec<Arc<(dyn PartitionStream + 'static)>> =
```
tychoish, line 54: I don't think you need to specify the type in the vector declaration `Vec<_>` should work.

crates/datasources/src/excel/table.rs
tychoish: well or just providing an implementation of `TableProvider` and `ExecutionPlan` rather than using the streaming version. I think deep down the code doesn't change just gets moved around a bit,

crates/datasources/src/excel/table.rs
hailelagi: temporary, was an experiment that I intended to come back to, deleting.

hailelagi:
phew! ch-cha-changes! @tychoish like we discussed the "streaming" table idea didn't quite map as nicely so it's currently a partial impl of `TableProvider` and a straightforward `ExecutionPlan`, notably it's still missing lots of features and improvements like `INSERT INTO`, `COPY` & `FILTER` or `WHERE` predicates but those can be added on in a separate issue hopefully. This also includes an `ObjectStore` interface to abstract away local vs remote file handling.

crates/datasources/src/excel/stream.rs
hailelagi: it no longer does.

crates/datasources/src/excel/mod.rs
```diff
@@ -1,89 +1,109 @@
...
+                })
+            }
+
+            DatasourceUrl::Url(_) => {
+                let store = store_access.create_store()?;
+                let list = store_access
+                    .list_globbed(&store, source_url.path().as_ref())
+                    .await?;
+
+                if list.is_empty() {
```
tychoish, line 59: should we error if there's more than one file? do we have a good way of handling multiple files?

crates/datasources/src/excel/mod.rs
```diff
@@ -143,27 +163,69 @@ fn xlsx_sheet_value_to_record_batch(
...
+        .iter()
+        .map(|col_name| {
+            let set = col_types.entry(col_name).or_insert_with(|| {
+                let mut set = HashSet::new();
+                set.insert(DataType::Utf8);
+                set
+            });
+
+            let mut dt_iter = set.iter().cloned();
+            let dt = dt_iter.next().unwrap_or(DataType::Utf8);
```
tychoish, line 210: can't we `into_iter` and avoid the mutable/extra declaration? 

is there a case where the datatype isn't string/utf8 here?

testdata/sqllogictests/xlsx.slt
```diff
@@ -57,3 +57,37 @@ select "HEADING" from read_excel('file://${PWD}/testdata/xlsx/multiple_sheets.xl
 # negatives are not allowed for infer_rows
 statement error
 select * from read_excel('file://${PWD}/testdata/xlsx/multiple_sheets.xlsx', sheet_name => 'other', infer_rows => -1);
+
+
+# create external table
```
tychoish, line 62: can we try tests/external tables with different sheet names and make sure that we get different sheets when specified?)

tychoish (commented):
this is looking really good!

crates/datasources/src/excel/mod.rs
hailelagi: yup an err makes sense for now if there's more than one file, likely this bit would get refactored when adding the machinery for `create database excel`

crates/datasources/src/excel/mod.rs
hailelagi: > is there a case where the datatype isn't string/utf8 here?
added better err handling for this at the call site in `infer_value_type`, the `unwrap_or` here seems mostly to satisfy iter semantics returning an Option

crates/datasources/src/excel/mod.rs
tychoish: my initial expectation is that "create database excel" would still be only one file: each and sheet would be it's own table...

tychoish (approved):

crates/datasources/src/excel/mod.rs
hailelagi: ah... alright then.

universalmind303 (approved):

PR #2699, opened 2024-02-23
refactor: extensible function registry

universalmind303:
I didn't want to create 1 massive undecipherable PR, so I wanted to submit this first. 

This ties a `FunctionRegistry` instance to a session instead of the previously global `FUNCTION_REGISTRY`. 

Right now, the entries are written to both the catalog and the function registry. 

I'd prefer writing to just one or the other, but havent quite figured that out yet as our udfs are not serializable so they cant be written directly to the catalog. 

Some other the notable changes so far
- some catalog changes to allow for adding new user defined functions to the catalog.
- internally to expose a new method `register_function` on the session. 
- rework `cosine_similarity` to the new datafusion udf impl pattern for easier usage with the new `register_function` method.

I manually tested this by removing cosine_similarity from the builtin functions and registering it manually

```rs
let mut sess = engine
    .new_local_session_context(SessionVars::default(), SessionStorageConfig::default())
    .await?;

let udf = Arc::new(CosineSimilarity::new());
sess.register_function(udf).await?;
```

This works as expected in **local** sessions only. We'll have to modify the planner to handle  udfs in a hybrid context at a later date. 

----
### Next steps. 

I'm going to try to integrate the work I did with this datafusion <--> polars POC in a subsequent PR.

https://github.com/universalmind303/polar-fusion

### DIFF
--- a/crates/metastore/src/database.rs
+++ b/crates/metastore/src/database.rs
@@ -1035,8 +1035,13 @@ impl State {
             }
             Mutation::CreateFunction(f) => {
                 let schema_id = self.get_schema_id(DEFAULT_SCHEMA)?;
+                let oid = self
+                    .schema_objects
+                    .get(&schema_id)
+                    .and_then(|objs| objs.functions.get(&f.name))
+                    .copied()
+                    .unwrap_or_else(|| self.next_oid());
 
-                let oid = self.get_or_next_oid(schema_id, &f.name);
                 let ent = FunctionEntry {
                     meta: EntryMeta {
                         entry_type: EntryType::Function,
@@ -1051,6 +1056,9 @@ impl State {
                     signature: Some(f.signature),
                     user_defined: true,
                 };
+                if let Some(ent) = self.entries.get(&oid)? {
+                    return Err(MetastoreError::DuplicateName(ent.get_meta().name.clone()));
+                }
                 self.entries.insert(oid, CatalogEntry::Function(ent))?;
                 self.schema_objects
                     .entry(schema_id)
--- a/crates/metastore/src/database.rs
+++ b/crates/metastore/src/database.rs
@@ -1033,6 +1033,31 @@ impl State {
                 // Update the new storage size
                 self.deployment.storage_size = update_deployment_storage.new_storage_size;
             }
+            Mutation::CreateFunction(f) => {
+                let schema_id = self.get_schema_id(DEFAULT_SCHEMA)?;
+
+                let oid = self.get_or_next_oid(schema_id, &f.name);
+                let ent = FunctionEntry {
+                    meta: EntryMeta {
+                        entry_type: EntryType::Function,
+                        id: oid,
+                        parent: schema_id,
+                        name: f.name.clone(),
+                        builtin: false,
+                        external: true,
+                        is_temp: false,
+                    },
+                    func_type: f.function_type,
+                    signature: Some(f.signature),
+                    user_defined: true,
+                };
+                self.entries.insert(oid, CatalogEntry::Function(ent))?;
+                self.schema_objects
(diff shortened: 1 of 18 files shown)


tychoish (commented):
seems reasonable. 

- this needs automated tests.
- should our other functions migrate over as well?

crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -131,6 +139,11 @@ pub trait BuiltinScalarUDF: BuiltinFunction {
     fn namespace(&self) -> FunctionNamespace {
         FunctionNamespace::None
     }
+
+    fn try_into_scalar_udf(self: Arc<Self>) -> datafusion::error::Result<ScalarUDF> {
```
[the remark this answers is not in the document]
universalmind303: because you can't implement tryfrom on foreign types

`impl TryFrom<ScalarUDF> for Arc<dyn BuiltinScalarUDF>` isnt valid.

crates/protogen/src/metastore/types/service.rs
```diff
@@ -691,20 +735,3 @@ impl From<UpdateDeploymentStorage> for service::UpdateDeploymentStorage {
         }
     }
 }
-
-#[cfg(test)]
-mod tests {
```
tychoish, line 696: can you add this back

crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -349,6 +362,16 @@ impl FunctionRegistry {
         }
         None
     }
+
+    pub fn register_udf(&mut self, udf: Arc<dyn BuiltinScalarUDF>) {
```
tychoish, line 366: this should probably error/panic if you attempt to register a function that has already been registered

crates/protogen/src/metastore/types/service.rs
universalmind303: no, we can't proptest anymore as `Signature` isn't compatible with `Arbitrary`.

crates/sqlbuiltins/src/functions/mod.rs
universalmind303: the catalog mutator already will error if it already exists

`catalog_mutator.mutate(catalog_version, mutations).await?;`

scsmithr (commented):
Could we add an SLT for the cosine similarity function just to make sure it's all wired up correctly (even if it needs to be skipped for rpc/flight).

universalmind303:
> Could we add an SLT for the cosine similarity function just to make sure it's all wired up correctly (even if it needs to be skipped for rpc/flight).

@scsmithr we already have a bunch of tests for it. `testdata/sqllogictests/functions/cosine_similarity.slt`

crates/metastore/src/database.rs
```diff
@@ -1033,6 +1033,39 @@ impl State {
...
+                        external: true,
+                        is_temp: false,
+                    },
+                    func_type: f.function_type,
+                    signature: Some(f.signature),
+                    user_defined: true,
+                };
+                if let Some(ent) = self.entries.get(&oid)? {
+                    return Err(MetastoreError::DuplicateName(ent.get_meta().name.clone()));
+                }
```
greyscaled, line 1061: (minir) - would it be possible/preferable to move this logic into lines 1038-1043?

It reads to me that we try to get an oid for the function name and otherwise just set the next oid, but maybe I am misunderstanding from reading the code, and this line is checking against the entries map, so maybe it's not the same thing?

tychoish (approved):

tychoish:
Let's hold off on merging this until we decide if #2715 is worth cutting a quick 0.9.1 over.

PR #2706, opened 2024-02-26
chore: omit remote metastore service

### LINKED ISSUE #560: Chore: Add integration testing for metastore
## Description

Currently all tests in Metastore use either an in-memory object store or a temp file object store. I believe we should also add in a way to test against GCS to avoid issues like the 411 Length Required error we were getting.

For every test in the `metastore` crate, we'll want to have a way of optionally building a GCS object store client. If `METASTORE_BUCKET_NAME` and `METASTORE_SERVICE_ACCOUNT_KEY` are non-empty environment variables, use those to build a GCS client. Otherwise use a memory object store.

We should have a static bucket in GCP to test against and the github actions service account should have storage object admin for that bucket.

## What This Enables

Bug prevention.

tychoish:
Want to run some things through CI. 

We sholdn't merge this as it'll require some changes to cloud
(granting access to the appropriate nodes,) passing new arguments to
the service, which we should do some testing to figure out how to do
that first.

Once this _does_ land we can remove all the client/service abstraction
and clean up a fair amount of the code, which might make some things
more clear (and make the metastore just a storage subsystem, and
interace onto the catalog).

Closes #560


tychoish:
In talking with @greyscaled, it became clear to us that we can drop all of the new configuration except the bucket name from the configuration options, because the access to that bucket will be added to the "main" credentials, which makes a lot of the code more simple. 

I want to hold of merging this until after things calm down a bit and we have time to coordinate landing this with cloud.

greyscaled:
adding cloud label, with love ❤️

crates/glaredb/src/args/server.rs
```diff
@@ -108,4 +101,17 @@ pub struct ServerArgs {
...
     #[arg(long, default_value="false", action = clap::ArgAction::SetTrue)]
     pub disable_postgres_api: bool,
+
+    /// Bucket to use for database catalogs.
+    #[clap(long, value_parser)]
+    pub metastore_bucket: Option<String>,
+
+    /// Path to GCP service account to use when connecting to GCS.
+    #[clap(long, value_parser)]
+    pub metastore_service_account_path: Option<String>,
```
greyscaled, line 111: These are the two main args that power it on cloud 👍🏻

tychoish: I think we need to delete the `service_account_path` because it should be pulled from the environment credentials, we only really need the bucket name, I would think. 

File path is only for "local" operation, and we can/should leave that.

crates/metastore/src/util.rs
```diff
@@ -31,16 +27,16 @@ impl MetastoreClientMode {
...
+            // MetastoreClientMode::Remote { addr } => {
+            //     info!(%addr, "connecting to remote metastore");
+            //     let channel = Endpoint::new(addr)?
+            //         .tcp_keepalive(Some(Duration::from_secs(600)))
+            //         .tcp_nodelay(true)
+            //         .keep_alive_while_idle(true)
+            //         .connect()
+            //         .await?;
+            //     Ok(MetastoreServiceClient::new(channel))
+            // }
```
tychoish, line 39:
```suggestion
```

crates/glaredb/src/args/server.rs
greyscaled: Yep, sounds like a plan

tychoish: I took out the metastore service account so we're just to bucket name (or filepath, if we want that to be configured, though in that case we could just put `__metastore` in the data directory...

crates/glaredb/src/server.rs
```diff
@@ -151,51 +157,62 @@ impl ComputeServerBuilder {
...
+        self.metastore_bucket = Some(name);
+        self
+    }
+
+    pub fn with_metastore_bucket_opt(mut self, bucket: Option<String>) -> Self {
+        self.metastore_bucket = bucket;
+        self
+    }
+
+
```
greyscaled, line 182: there are some extra newlines from 170-182

crates/glaredb/src/server.rs
```diff
@@ -57,14 +58,15 @@ pub struct ComputeServerBuilder {
     disable_rpc_auth: bool,
     enable_simple_query_rpc: bool,
     enable_flight_api: bool,
+    metastore_bucket: Option<String>,
+    metastore_local_file_path: Option<PathBuf>,
```
greyscaled, line 62: so this already took a `service_account_path`, and just to double-check that I have it right, we are adding some capabilities to this SA that was previously on a separate metastore SA?

That tracks to me, just wanted to double check.

We should make that change and deploy this branch to QA to test before merging it into main.

We can also phase out the metastore key until v0.10.0 or later releases

crates/glaredb/src/server.rs
tychoish, line 182:
```suggestion
```

crates/glaredb/src/server.rs
tychoish: the local file path is "how do we run in single node with the metastore in local files" rather than remote in a service (current cloud implementation) or in memory (tests/etc.)

This option is nil/none (I think) on cloud, and I think (re another comment) I can get rid of it to just be a sub-directory of "data-dir"... 

I think I will remove this option entirely and put it in

greyscaled: This comment is not well placed, the line of code it's referring to is:

https://github.com/GlareDB/glaredb/blob/6a03e0ea84353dbd9c1118f085d515f46c9bb393/crates/glaredb/src/server.rs#L52

tychoish: I believe that's correct. 

We'll be able to test this out on QA (or from here, if you want?) and I'm down to find a time when we can pair on this.

greyscaled: Yea, all we need to do is comment out this line:

https://github.com/GlareDB/glaredb/blob/6a03e0ea84353dbd9c1118f085d515f46c9bb393/.github/workflows/ci.yaml#L623

and then I can roll it out to QA, along with dropping metastore.

We should look to see it fail first and then merge the SA perms and verify it works.

We can pair on this tomorrow (Friday) if you'd like, it would be a good task for me after the release stuff.

crates/glaredb/src/args/server.rs
```diff
@@ -108,4 +101,13 @@ pub struct ServerArgs {
...
     pub disable_postgres_api: bool,
+
+    /// Bucket to use for database catalogs.
+    #[clap(long, value_parser)]
+    pub metastore_bucket: Option<String>,
+
+    /// Local file path to store database catalog (for a local persistent
+    /// store).
+    #[clap(long, value_parser)]
+    pub metastore_local_file_path: Option<PathBuf>,
```
vrongmeal, line 112: Where is this used?

tychoish: fixed, cruft

PR #2707, opened 2024-02-26
fix: improve bson iteration performance

tychoish:
Minor: this avoids casting bson into values until we need to. This
means that if a document in a stream has a field which is not in the
schema, we just skip it.

### DIFF
--- a/crates/datasources/src/bson/builder.rs
+++ b/crates/datasources/src/bson/builder.rs
@@ -90,27 +90,26 @@ impl RecordStructBuilder {
 
     pub fn append_record(&mut self, doc: &RawDocument) -> Result<()> {
         let mut cols_set: BitVec<u8, Lsb0> = BitVec::repeat(false, self.fields.len());
+        let doc_buf = doc.to_raw_document_buf();
+        let iter = doc_buf.iter_elements();
 
-        for iter_result in doc {
-            match iter_result {
-                Ok((key, val)) => {
-                    let idx = *self
-                        .field_index
-                        .get(key)
-                        .ok_or_else(|| BsonError::ColumnNotInInferredSchema(key.to_string()))?;
+        for item in iter {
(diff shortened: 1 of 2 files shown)


vrongmeal:
How does this affect the performance. I can't figure out where do we "skip" if the field doesn't exist?

tychoish:
> How does this affect the performance. I can't figure out where do we "skip" if the field doesn't exist?

It depends on the data, but it makes the process of iterating through a document lazier. If you have a schema that was `(a int, b text)` but a document that was `{"a":1,"c":<...>,"b":"foo"}` you could skip reading the c field entirely. Because all fields are length prefixed you just skip ahead... in the previous iterator it would need to parse the value of the field even if we weren't going to use it.

crates/datasources/src/bson/builder.rs
```diff
@@ -90,27 +90,26 @@ impl RecordStructBuilder {
 
     pub fn append_record(&mut self, doc: &RawDocument) -> Result<()> {
         let mut cols_set: BitVec<u8, Lsb0> = BitVec::repeat(false, self.fields.len());
+        let doc_buf = doc.to_raw_document_buf();
```
vrongmeal, line 93: This seems to copy the underlying bytes.

Are we sure we're actually improving an average performance by implementing this (compared to the extra alloc and copy of bytes)? This seems like a guess.

---

I did look into the underlying iterator. `RawIter::new` is not public. We can raise a PR to make it public since it accepts `&RawDocument` and this could be avoided. Until then, I don't think we should move ahead with this without any concrete benchmarks.

tychoish: I pushed a patch that avoids the extra copy (except for nested documents, but will open a couple of ergonomic features on the upstream lib.)

vrongmeal (approved):

PR #2710, opened 2024-02-27
feat: polars exprs over ffi

universalmind303:
depends on https://github.com/GlareDB/glaredb/pull/2699


crates/sqlbuiltins/src/functions/scalars/polars_ffi.rs
universalmind303, line 1: Most of this is copied from

https://github.com/pola-rs/polars/blob/d6c6cef2c1d699c75c3c5d24fc56b32c6a49542f/crates/polars-plan/src/dsl/function_expr/plugin.rs

universalmind303:
closing for now. There seems to be a lot of edge cases with polars extensions that I'm not entirely sure how to handle yet.

PR #2712, opened 2024-02-27
feat: external functions for internal usage

universalmind303:
This adds some new ffi stuff for adding external functions. 
`extensions/distance` contains a working example of creating said external function.

While some functionality is added to potentially support non scalar functions (table_returning, aggregate), the functionality currently limited to scalar udfs only.

Currently it only works by 'installing' them via the lib path directly.
```sh
> cd extenstions/distance
> cargo build
> cd - 
> just run local

>>> `install '/path/to/target/debug/libglaredb_distance.dylib'`;
>>> select cosine([1.0], [1.0]);
```

Eventually I'd like try to get an extension published up to github, or maybe our own private extension registry with "offical" extensions like duckdb. 

that way youd just be able to do 

```sql
install 'distance'; # or whatever else the extension name is.
```


crates/sqlbuiltins/src/functions/mod.rs
```diff
@@ -182,8 +183,8 @@ where
 pub struct FunctionRegistry {
     // TODO: What's the difference between `BuiltinFunction` and
     // `BuiltinScalarUDF`? Still confused.
-    funcs: AliasMap<String, Arc<dyn BuiltinFunction>>,
-    udfs: AliasMap<String, Arc<dyn BuiltinScalarUDF>>,
+    funcs: Arc<Mutex<AliasMap<String, Arc<dyn BuiltinFunction>>>>,
+    udfs: Arc<Mutex<AliasMap<String, Arc<dyn BuiltinScalarUDF>>>>,
```
universalmind303, line 187: we need inner mutability

crates/ffi/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+[dependencies]
+arrow = { version = "50.0.0", features = ["ffi"], default-features = false }
+arrow-schema = { version = "50.0.0", features = ["ffi"] }
+thiserror.workspace = true
+
+datafusion = { workspace = true, optional = true }
+paste = "1.0.14"
+
+[features]
+datafusion = ["dep:datafusion"]
```
universalmind303, line 17: i wasn't sure if we'd need datafusion in any/all of these extensions so i made it opt in. This cuts out ~150 dependencies, reduces build time, build size of `extensions/distance` significantly from `5.6MB` to `3.8MB`.

.DS_Store
greyscaled, line 1: Looks like this needed to be `.gitignore`d

crates/ffi/src/lib.rs
```diff
@@ -238,11 +259,8 @@ macro_rules! generate_lib {
...
-    array: FFI_ArrowArray,
-    schema: &FFI_ArrowSchema,
-) -> datafusion::error::Result<ArrayRef> {
-    let data = datafusion::arrow::ffi::from_ffi(array, schema).unwrap();
-    let arr = datafusion::arrow::array::make_array(data);
+pub unsafe fn import_array(array: FFI_ArrowArray, schema: &FFI_ArrowSchema) -> Result<ArrayRef> {
+    let data = arrow::ffi::from_ffi(array, schema).unwrap();
+    let arr = arrow::array::make_array(data);
     Ok(arr)
 }
```
tychoish: seems like we should either unwrap/panic or return a result, but the current state of "panic and never return an error result" is either cruft or a mistake.

tychoish (commented):
some of this code is copy-pasted from other sources, right? can you put in links to where you got them for future reference?

universalmind303:
> some of this code is copy-pasted from other sources, right? can you put in links to where you got them for future reference?

The stuff that is straight copy/pasted is documented as such. The rest is derivative of other ffi code bases, not copy/pasted.

Issue #2675, opened 2024-02-21
test(integration): `cloud_upload`


greyscaled:
As a follow-up to #2670:

> we could probably write tests for this using mino or the mock gcs fixture

_Originally posted by @tychoish in https://github.com/GlareDB/glaredb/issues/2670#issuecomment-1956605809_

greyscaled:
I think mock gcs fixture would be best - but this code relies on a cloud/remote instance. Do we have prior art for running such tests without reaching out to a test instance of glaredb cloud?

Intuitively, I'd think we can start a glaredb connected to the gcs fixture and then use a client to connect to it via pg or rpc.

If such a test or pieces of it exist, I'd be very thankful to be pointed to the code(s).

tychoish:
I think it doesn't require cloud, as much as a glaredb invocation that has credentials for the service account that can access the buckets where native storage is, and one user, but I might be missing something

greyscaled:
Yes I didn't mean to imply cloud I'd absolutely required the `cloud instance` arises from the name of the function on session vars.

Oops.

But ya I would think we need to run the binary with the right things configured and I'm unsure if we have any existing tests doing that so it may end up being a novel setup.

Will have bandwidth to look into this more in the coming sprints

PR #2715, opened 2024-02-28
Grey/hybrid exec file upload

greyscaled:


tychoish:
This seems fine, particularly if it works. Is there a way to write a test, or a comment (or set of comments that might prevent this kind of confusion in the future?

greyscaled:
> Is there a way to write a test, or a comment (or set of comments that might prevent this kind of confusion in the future?

I'll take a stab at it, as I agree that it feels uncomfortable with the lack of testing for specific hybrid cases (#2675)

universalmind303 (approved):

scsmithr (approved):

PR #2717, opened 2024-03-01
chore: add SQLite to list of supported Databases

judahrand:
It looks to me like updating the README was missed in #2568.


tychoish:
Thanks for the catch!

PR #2718, opened 2024-03-01
fix: Iceberg for format v1

vrongmeal:
Currently this commit simply adds `#[serde(default)]` for fields that are required in v1 (or sets a valid default).

We do need more datasets to test this on. There's some deprecated stuff that might require more attention to support.


tychoish (approved):
I feel like we could write some random generators for iceberg data and (maybe there are some scripts/generators that we could take advantage of?) and get that hooked up to the pytests to generate some datasets on the fly. 

I also sort of want to test this through other object-storage layers, mostly because my trust of the `object_store` crate is pretty low. 

Having said that, I don't think either of these things should block merging, and this patch is pretty great.

PR #2732, opened 2024-03-04
chore: add dbt read_csv test

talagluck:

PR #2734, opened 2024-03-05
chore: pytest path cleanup and port management

### LINKED ISSUE #2730: pytests fail locally if `target/release` is not available
## Description

```sh
> git clone https://github.com/GlareDB/glaredb.git
> cd glaredb
> just build
> just venv
> just pytest
```

this fails on `test_release_exists`

tychoish:
This contains two major changes:

- we'll use either the release or the debug (in that order of
  preference) build if they exist. If neither exists, 🤷, that seems
  like a reasonable failure mode for integration tests. (Inspired by
  #2730)

- randomly select an unused port, and use that (rather than the
  default postgres port.) this means we can run tests in parallel (and
  doing so halves the runtime on my laptop.) This had been on my list
  for a while, but I hadn't yet been bothered by the runtime.

Closes #2730


tests/test_dbt.py
```diff
@@ -80,20 +81,21 @@ def test_dbt_glaredb_external_postgres(
...
+                    "--project-dir",
+                    dbt_project_path,
+                    "--profiles-dir",
+                    dbt_project_path,
+                    "-m",
+                    model_name,
+                ]
+            )
+
+            assert res.success is True
```
talagluck, line 98: The nested `with` block looks kind of gnarly. [Maybe something like this instead?](https://stackoverflow.com/a/46481946)

tests/fixtures/dbt_project/profiles.yml
```diff
@@ -4,9 +4,9 @@ glaredb_dbt_test:
       dbname: default
       host: 127.0.0.1
       pass: ""
-      port: 5432
+      port: "{{ env_var('GLAREDB_PORT') | as_number }}"
```
talagluck, line 7: This is a welcome improvement.

talagluck (approved):
Just one suggestion in a couple of places. I'll wait to merge #2732.

tests/test_dbt.py
tychoish: huh! amazing!

PR #2735, opened 2024-03-05
fix: Allow a process to acquire a lease its already acquired

scsmithr:
Related: https://github.com/GlareDB/cloud/issues/2746

Previously we weren't allowing it just errored with something along the lines of:

```
Catalog error: Lease for database 'a9edc39a-6129-4b19-b8dc-02483df4fc6e' held by other process; other_process_id: d7796f4c-c3e4-4822-b59e-1b4b9f32a48d, current_process_id: d7796f4c-c3e4-4822-b59e-1b4b9f32a48d
```

I believe I originally intended to allow a leaser to acquire a lease its already acquired based on the error message.

This doesn't fix the rate limit issue, but does partially alleviate errors causes by failing to drop leases because dropping the lease failed due to a rate limit.

### DIFF
--- a/crates/metastore/src/storage/lease.rs
+++ b/crates/metastore/src/storage/lease.rs
@@ -242,7 +242,7 @@ impl LeaseRenewer {
             })?;
 
             // Some other process has the lease.
-            if expires_at > now {
+            if expires_at > now && held_by != self.process_id {
                 return Err(StorageError::LeaseHeldByOtherProcess {
                     db_id: self.db_id,
                     other_process_id: held_by,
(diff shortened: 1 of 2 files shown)


crates/metastore/src/storage/lease.rs
```diff
@@ -242,7 +242,7 @@ impl LeaseRenewer {
             })?;
 
             // Some other process has the lease.
-            if expires_at > now {
+            if expires_at > now && held_by != self.process_id {
```
tychoish, line 245: i presume that `self.process_id` is a UUID generated on process start?

tychoish (approved):

universalmind303 (approved):

crates/metastore/src/storage/lease.rs
scsmithr: yep, see https://github.com/glaredb/glaredb/blob/30ff37ca38d32a0084c6c620b6a333bc81274d5b/crates/metastore/src/srv.rs#L40-L49

PR #2739, opened 2024-03-05
wip(auth): argument for client ident cert

greyscaled:
Not much code yet, I'm working across both this cb and cloud, because ultimately the cert we're passing in here needs to come from cloud.

I wanted to get this started and out there early even though it's currently just an unused arg.

### DIFF
--- a/crates/glaredb/src/args/local.rs
+++ b/crates/glaredb/src/args/local.rs
@@ -96,6 +96,23 @@ pub struct LocalClientOpts {
     /// Note: Keep in sync with py-glaredb connect
     #[arg(long, default_value = "https://console.glaredb.com", hide = true)]
     pub cloud_addr: String,
+
+    /// Path to a certificate that identifies this client.
+    ///
+    /// (Internal) - volatile, might be removed or drastically modified.
+    ///
+    /// This option is meaningless if either `--disable_tls`or
+    /// `--ignore_rpc_auth` is set.
+    ///
+    /// TODO/REVIEW: We're probably going to want to store certs in data-dir,
+    ///              or some well-known path on disk (like AppData). We want a
+    ///              place to store them (if available) even if `--cloud-url`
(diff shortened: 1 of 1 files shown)


crates/glaredb/src/args/local.rs
```diff
@@ -96,6 +96,23 @@ pub struct LocalClientOpts {
...
+    ///
+    /// TODO/REVIEW: We're probably going to want to store certs in data-dir,
+    ///              or some well-known path on disk (like AppData). We want a
+    ///              place to store them (if available) even if `--cloud-url`
+    ///              is passed (unlike --data-dir). It probably makes sense to
+    ///              keep an env var or arg around to enable workarounds and
+    ///              debugging, but otherwise the ideal flow is having this be
+    ///              managed by glaredb at a known persistence (non-temp). Must
+    ///              be mindful/aware of where we persist these for rotation.
+    pub identity_path: String,
```
tychoish, line 115: don't we have a `~/.glaredb` directory/location that we have some stuff? I think for CLI state like this it makes sense.

greyscaled: I am under the impression that dir only exists/is created if using CLI without `--cloud-url`, and I also think it's important we consider if we want to store certs always with `--data-dir` or separately (my inclination is separately, but that's not based on anything material)

greyscaled: > I am under the impression that dir only exists/is created if using CLI without `--cloud-url`

Ok, I am wrong about this - we store query history there (even if glaredb is only ever used with `--cloud-url`. This is a good start.

greyscaled: OK, it turns out `.glaredb` is always created for query history, regardless of `--data-dir`, which is what I was hoping for.

I will resolve this comment and update the comment in the code - `.glaredb` is what we'll use!

PR #2745, opened 2024-03-06
feat: insert into sqlite tables

tychoish:
This is close, but the test fails with a common error from SQLite on
the insert statement (e.g. `attempt to write a readonly
database`). I'm pretty sure that this _isn't_ due to the python sqlite
library which creates the db and sets things up, but something with
how we (probably?) create the sqlite client object.

Wanted to put this up so @vrongmeal could take a quick look and make
sure there isn't something specific to the rest of our implementation
that I missed.

### DIFF
--- a/crates/datasources/src/sqlite/mod.rs
+++ b/crates/datasources/src/sqlite/mod.rs
@@ -18,7 +18,9 @@ use datafusion::execution::TaskContext;
 use datafusion::logical_expr::{BinaryExpr, TableProviderFilterPushDown, TableType};
 use datafusion::physical_expr::PhysicalSortExpr;
 use datafusion::physical_plan::metrics::{ExecutionPlanMetricsSet, MetricsSet};
+use datafusion::physical_plan::stream::RecordBatchStreamAdapter;
 use datafusion::physical_plan::{
+    execute_stream,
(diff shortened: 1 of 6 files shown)


vrongmeal:
> This is close, but the test fails with a common error from SQLite on
the insert statement (e.g. attempt to write a readonly database). I'm pretty sure that this isn't due to the python sqlite
library which creates the db and sets things up, but something with
how we (probably?) create the sqlite client object.

Yeah, we do create the client in "read-only" mode: https://github.com/GlareDB/glaredb/blob/9c27dd0289481bcca895b4a3764002682d7d8e16/crates/datasources/src/sqlite/wrapper.rs#L32-L36

vrongmeal (approved):
In general, this looks good!

crates/datasources/src/sqlite/mod.rs
```diff
@@ -286,6 +290,26 @@ impl TableProvider for SqliteTableProvider {
             metrics: ExecutionPlanMetricsSet::new(),
         }))
     }
+
```
scsmithr, line 293:
```suggestion
```

scsmithr (approved):

Issue #2744, opened 2024-03-06
Chore: refactor datasources


universalmind303:
## Description

proposal to do some internal refactoring on table functions and datasources to make adding new ones easier, while also making better usage of incremental compilation. 

Some background context..

It is a pretty big pain point that table syntax is completely separate from table function syntax. We usually don't implement these at the same time due to the added complexity of the table syntax.

```sql
select * from read_excel(...) # table function
create external table t from excel options (...) # table syntax
```

The code also lives in very different parts of the application and is quite difficult to follow. 

Ideally we should be able to create a new datasource in isolation and register it all in one shot. Currently we have to 

1. create the `TableProvider` impl ex: `crates/datasources/src/excel/mod.rs`
2. create the FunctionEntry _(for the catalog)_ ex: `crates/sqlbuiltins/src/functions/table/excel.rs:21`
3. create the `TableFunc` `crates/sqlbuiltins/src/functions/table/excel.rs:21`
4. create the `TableOptions` proto `crates/protogen/proto/metastore/options.proto:208`
5. create the `TableOptions` crates/protogen/src/metastore/types/options.rs:1053`
6. correlate the options variant (this happens in quite a few spots in `options.rs`)
7. finally connect it into the planner: `crates/sqlexec/src/planner/session_planner.rs:903`

That's a lot of different parts of the code that need to be changed. Datasources should instead just program to a trait or traits, we register them in the registry, and then all of the above should just work. Reducing the amount of code to just two. (datasource implementation, and registry). 

Proposal

The `sqlbuiltins::functions` module currently provides 
- `TableFunc` (used to convert a table function into a table provider)
- `BuiltinFunction` _this should be renamed to `FunctionCatalogEntry` or similar_ (used to register a function within the catalog)

and datafusion provides us with `TableProvider` trait that is used for creating record batches. 

Currently we have no traits for the table options syntax so it is manually implemented each time, but we probably need one. `DynTableOptions` _(bikeshedding)_

I suggest that we create a new crate that provides all of this functionality wrapped up in a single trait `Datasource: TableFunc + FunctionCatalogEntry + TableProvider + TryFrom<DynTableOptions>` _(bikeshedding)_. then that can be used to populate the catalog, function registry, and planner all at once. 

So when adding a new datasource, you just implement this set of traits, and plug it in as `Arc<dyn Datasource>` somewhere and we're good to go.

tychoish:
I like this! 

[...]

PR #2747, opened 2024-03-06
refactor: move parser into own crate

universalmind303:
some prereq work for https://github.com/GlareDB/glaredb/issues/2744

### DIFF
--- a/crates/datasources/Cargo.toml
+++ b/crates/datasources/Cargo.toml
@@ -25,6 +25,7 @@ futures = { workspace = true }
 gcp-bigquery-client = "0.18.0"
 klickhouse = { version = "0.11.2", features = ["tls"] }
 protogen = { path = "../protogen" }
+parser = {path = "../parser"}
 datafusion_ext = { path = "../datafusion_ext" }
 mongodb = "2.8.1"
 mysql_async = { version = "0.33.0", default-features = false, features = [
--- a/crates/parser/Cargo.toml
+++ b/crates/parser/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "parser"
+version.workspace = true
+edition.workspace = true
+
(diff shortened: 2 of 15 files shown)


crates/parser/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+version.workspace = true
+edition.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+datafusion.workspace = true
+datafusion_ext = { path = "../datafusion_ext" }
+protogen = { path = "../protogen" }
+prql-compiler = "0.11.3"
```
scsmithr, line 12: Are there other crates that depend on prql? Can we make it just this crate that depends on it?

scsmithr (approved):

crates/parser/Cargo.toml
universalmind303: yeah good catch, I forgot to remove that from `sqlexec` should be good to go now.

crates/datasources/Cargo.toml
```diff
@@ -25,6 +25,7 @@ futures = { workspace = true }
 gcp-bigquery-client = "0.18.0"
 klickhouse = { version = "0.11.2", features = ["tls"] }
 protogen = { path = "../protogen" }
+parser = {path = "../parser"}
```
tychoish, line 28: super-nit: we added spaces in other places for this form. 

idle musing: is there a toml linter we could get?

tychoish (approved):
This seems my only question/concern is "should we worry about the proptest or think about replacing it?"

I don't know if there's anything to be done about it, but the name collision between our `parser` and the upstream parser package (particularly `ParserError`) seems like a thing that will be confusing in the future.

PR #2766, opened 2024-03-08
chore: add explicit schema support for json streaming

tychoish:
This is the same implementation as in the BSON, need to have an
external table connection to wire this up directly.


vrongmeal (approved):
This looks good!

PR #2767, opened 2024-03-08
refactor: table options

universalmind303:


crates/protogen/src/metastore/types/options.rs
universalmind303, line 1: Eventually, the individual table options will be moved into their respective datasources, but I just kept them in there for now to keep things more relevant.

crates/protogen/src/metastore/types/options.rs
```diff
@@ -16,6 +17,48 @@ pub struct InternalColumnDefinition {
     pub arrow_type: DataType,
 }
 
+impl From<InternalColumnDefinition> for OptionValue {
```
universalmind303, line 20: @scsmithr So i think the serde route may actually be a bit cleaner here. 

The `OptionValue` is already functioning the same as `serde_json::Value`, but with a ton more work on our end. Once we have nested/recursive datatypes that don't translate easily to this structure, such as `DataType`, it requires us to do an opaque serialization anyways, so i think just using serde for everything makes a lot of sense. 

It also adds a TON of boilerplate that is essentially the equivalent of `#[derive(Serialize, Deserialize)]` _(like half of this PR is `From` and `TryFrom` impls that serde would replace)._

crates/sqlexec/src/dispatch/external.rs
```diff
@@ -148,160 +145,282 @@ impl<'a> ExternalDispatcher<'a> {
...
-                };
+        // TODO, use the DATASOURCE_REGISTRY to dispatch the database.
 
-                let conn_params = SnowflakeDbConnection {
-                    account_name: account_name.clone(),
-                    login_name: login_name.clone(),
-                    password: password.clone(),
-                    database_name: database_name.clone(),
-                    warehouse: warehouse.clone(),
+        if let Some(ds) = DATASOURCE_REGISTRY.get(&db.options.as_str()) {
```
tychoish, line 150: I'm not sure I understand the goal/path here: having this match statement _here_ still has the same dependency properties as it did before. Is there more work planned here?

crates/datasources/src/lib.rs
```diff
@@ -17,3 +30,49 @@ pub mod postgres;
 pub mod snowflake;
 pub mod sqlite;
 pub mod sqlserver;
+
+
+pub type DatasourceError = Box<dyn std::error::Error + Send + Sync>;
+#[async_trait]
+pub trait Datasource: Send + Sync {
+    fn name(&self) -> &'static str;
```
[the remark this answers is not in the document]
universalmind303: you cant have const attributes on object safe traits.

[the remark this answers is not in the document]
scsmithr: https://doc.rust-lang.org/reference/items/traits.html#object-safety

crates/sqlexec/src/dispatch/external.rs
universalmind303: The `else` statement will eventually go away once everything is migrated to the new table options. The goal here is _decoupling_. Now everything is encapsulated within the `DATASOURCE_REGISTRY` and the external dispatcher is no longer tightly coupled to the individual variants of the datasources. Right now the `DATASOURCE_REGISTRY` is a static global, but the idea long term is to do the same as we did with the `FunctionRegistry` in https://github.com/GlareDB/glaredb/pull/2699 which will allow us to dynamically register new datasources.

crates/sqlbuiltins/src/lib.rs
```diff
@@ -8,7 +8,35 @@
...
+    once_cell::sync::Lazy::new(|| DatasourceRegistry::new());
+
+pub struct DatasourceRegistry {
+    pub(crate) datasources: HashMap<&'static str, Arc<dyn datasources::Datasource>>,
+}
+
+impl DatasourceRegistry {
+    pub fn new() -> Self {
+        let datasources: Vec<Arc<dyn datasources::Datasource>> =
+            vec![Arc::new(DebugDatasource), Arc::new(LanceDatasource)];
```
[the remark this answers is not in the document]
universalmind303: yeah, I didn't want to add another layer of complexity to this PR, so the `DatasourceRegistry` is just a global static, but long term it'll be implemented the same as https://github.com/GlareDB/glaredb/pull/2699 which will allow us to dynamically register datasources.

crates/protogen/src/metastore/types/options.rs
```diff
@@ -1263,6 +2332,93 @@ pub struct TableOptionsSnowflake {
     pub table_name: String,
 }
 
+impl From<TableOptionsSnowflake> for TableOptions {
```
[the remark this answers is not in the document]
universalmind303: this is part of the serde topic. the from/tryfrom are equivalent to serialize/deserialize. the `From<T> for TableOptions` should always be infallible, but converting `TableOptions` to `T` is fallible.

They are used within the `Datasource` implementation

```rs
    fn table_options_from_stmt(
        &self,
        opts: &mut StatementOptions,
        creds: Option<CredentialsOptions>,
        tunnel_opts: Option<TunnelOptions>,
    ) -> Result<TableOptions, DatasourceError>;
    
    async fn create_table_provider(
        &self,
        options: &TableOptions,
        tunnel_opts: Option<&TunnelOptions>,
    ) -> Result<Arc<dyn TableProvider>, DatasourceError>
```
a good example of the implementation and how they are used can be found in the `Debug` datasource.

- `crates/datasources/src/debug/mod.rs` 
- `crates/datasources/src/debug/options.rs`

crates/datasources/src/lib.rs
```diff
@@ -17,3 +30,49 @@ pub mod postgres;
...
+    fn validate_tunnel_connections(
+        &self,
+        tunnel_opts: Option<&TunnelOptions>,
+    ) -> Result<(), DatasourceError>;
+
+    /// Validate the provided credentials.
+    fn validate_credentials(
+        &self,
+        creds: Option<CredentialsOptions>,
+    ) -> Result<(), DatasourceError>;
```
tychoish, line 50: missed this comment in my previous review: 

I feel like there should be default implementations of this so that if there aren't any credentials or options then implementors shouldn't have to implement? 

Also:
- confused why `TunnelOptions` is a ref but credentials isn't.
- not sure that we need to have a reference to the object?

crates/datasources/src/lib.rs
```diff
@@ -17,3 +30,49 @@ pub mod postgres;
...
+
+    /// Create a new datasource from the provided table options and credentials.
+    /// CREATE EXTERNAL TABLE foo FROM <name> OPTIONS (...) [CREDENTIALS] (...) [TUNNEL] (...)
+    // TODO: the datasource should have control over it's own CredentialsOptions and TunnelOptions
+    fn table_options_from_stmt(
+        &self,
+        opts: &mut StatementOptions,
```
tychoish, line 57: I don't think I understand why this needs to be mutable.

crates/datasources/src/lib.rs
```diff
@@ -17,3 +30,49 @@ pub mod postgres;
...
+        creds: Option<CredentialsOptions>,
+        tunnel_opts: Option<TunnelOptions>,
+    ) -> Result<TableOptions, DatasourceError>;
+
+
+    async fn create_table_provider(
+        &self,
+        tbl_options: &TableOptions,
+        _tunnel_opts: Option<&TunnelOptions>,
```
tychoish, line 66: in my mind, tunnel options (like credentials, really, are a component of the table options and depend on the implementation details of the provider rather than something that's pulled out at this level. I understand that it's not for historical reasons, and maybe can't be yet in the process, but I think it'd be good to:
- move toward that if it's possible
- explain why it's not possible if it's not/neer going to be possible
- leave a note/plan here if it's possible but we want to defer it.

crates/datasources/src/lib.rs
```diff
@@ -17,3 +30,49 @@ pub mod postgres;
...
+    ) -> Result<Arc<dyn TableProvider>, DatasourceError>;
+
+    /// Create a new datasource from the provided database options and credentials.
+    /// If the datasource does not support databases, return `Ok(None)`.
+    async fn table_provider_from_db_options(
```
tychoish, line 71: do implementations need to create two factories for tables for both db configs (which many data sources aren't going to have)?

it seems reasonable that `Database Options` could have some kind of converter to table options in some cases and this could be an implementation detail of the registry/dispatcher.

crates/datasources/src/lib.rs
```diff
@@ -17,3 +30,49 @@ pub mod postgres;
...
+        &self,
+        schema: &str,
```
tychoish, line 73: thought: it might be useful if we call this something like "namespace" rather than schema, as there's ambiguity about table schema in this name.

crates/datasources/src/lib.rs
universalmind303: I agree that they are a component of the datasource, but disagree that they should be part of the table options. The table options should be a 1-1 mapping of the sql `options (...)`, similar to `credentials` and `tunnel`. Instead of trying to throw everything into the `TableOptions`, i'd prefer that `CredentialsOptions` and `TunnelOptions` are refactored similarly to `TableOptions`. 

Each datasource should optionally provide their own `CredentialsOptions`, and/or `TunnelOptions` variant. I already added a small note in the `Datasource` trait definition expressing this.

testdata/sqllogictests/external_schema.slt
```diff
@@ -1,53 +1,55 @@
...
-	z int
-);
-
-query II
-select * from dstextra;
-----
-NULL 1 0 
-NULL 2 10
-NULL 3 100
+# SKIP FOR NOW
```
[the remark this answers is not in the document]
tychoish: I guess I don't quite understand what the demarcation between "things that belong in table options" and "things that are an additional option".

crates/datasources/src/lib.rs
universalmind303: we historically have been taking the values out of the statement via `.remove`. This avoids a few unnecessary `.clone()`.

It's important to keep in mind that this current `Datasource` definition is based on a lot of historical decisions such as using `Option<CredentialOptions>` and `Option<&TunnelOptions>`. The goal is just to first encapsulate everything into that trait, then we should be able to make incremental improvements much easier

crates/datasources/src/lib.rs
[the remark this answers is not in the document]
tychoish: do the datasources need to be object-safe? why would they be different from functions in this regard?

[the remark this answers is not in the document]
scsmithr: Yes, being object safe allows for dynamic dispatch. Without it, we can't do `Box<dyn Datasource>`. There's ways around that (like down casting or adding additional type bounds), but I think just having this return a static string is the easiest.

crates/datasources/src/lib.rs
tychoish: I think being explicit makes sense, but I think that the default implementations could/should error so that if you _did_ specify tunnel for a file-based protocol, you'd get an error, (which is correct,) and implementers of datasources wouldn't have to think if tunnels made sense for their datasource?

crates/datasources/src/lib.rs
tychoish: This makes sense: I'd like to have comments or TODOs for the things that call out the parts that are just cruft or legacy accommodations, because it's hard (as a reviewer) to see "what's the design" and "what's the implementation" in all of this.

crates/datasources/src/lib.rs
tychoish: >  The table options should be a 1-1 mapping of the sql options (...), similar to credentials and tunnel. 

I'm not sure I understand this: the indirection between the statements issued by the users and the table options affords us some extra flexibility and provides the ability for both the implementation and the syntax to be easier to use and develop. By forcing our internal representation to mirror the syntax, all of the table options code, becomes inflexible and we don't actually benefit very much from the indirection. 

If the design means that credentials and tunnel options have to be orthogonal to table options, and we want to handle things like explicit schema orthogonally, then every time we need to add a new feature it sort of seems like we'll have a discussion about 'should this be a new orthogonal configuration category or go in the table options' which doesn't feel productive. 

If authentication (credentials/tunnel) can't be part of the table options then it sort of feels like the only thing in the table options are "location/path" and "format" (if that's not on the provider itself,) and so the options just become this catch all for "everything that isn't top level," which feels like it'll both make the code (and the prospect of implementing a new datasource or adding a feature to an existing datasource) quite hard. 

I might be extrapolating too far along some of these lines, and would definitely be interested in hearing your vision for this, but maybe we should talk about this synchronously sometime soon, but I think we need to have some alignment on how we expect to use these in the future.

testdata/sqllogictests/external_schema.slt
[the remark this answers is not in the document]
universalmind303: this has a clear 1-1 mapping to the sql syntax. 

```sql
create external table t 
from <Datasource::NAME> 
[CREDENTIALS] <CredentialOptions> 
[TUNNEL] <TunnelOptions>
[OPTIONS] <TableOptions>
[COLUMNS] <Fields>
```

crates/datasources/src/lib.rs
universalmind303: I could see an argument for the inverse as you suggested. (the tunnel having control over the datasources).

something like

```rs
pub trait Tunnel {
  fn validate_datasource(&dyn Datasource) -> Result<()> 
}
```

Unfortunately, our current architecture tightly couples the datasources to the tunnels, so i think this would be a larger change outside the scope of the table options refactor.

crates/datasources/src/lib.rs
tychoish: I definitely understand the sentiment, but this refactor also includes the datasource interface, and implies/suggests the design of that system, so I think it's important that we agree about where we're going with that design at this stage.

crates/datasources/src/lib.rs
tychoish: I would have thought that this refactor could shim a different interface on top of our current architecture.

Regardless we should leave comments here to describe the desired endstate, and make sure we track getting there.

testdata/sqllogictests/external_schema.slt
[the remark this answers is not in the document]
tychoish: I think that we either need to have: 
- every class of configuration object as a top level object with a clear way to add new kinds of configuration objects, but without an `options` section.
- all configuration classes are components of an `TableOptions` type. 

I do understand that migrating to this might be a multi-step process, but we should agree on what the goal is before getting too much further. 

I also think that a 1-1 mapping with the sql syntax shouldn't be a design goal (and might be a non-goal?)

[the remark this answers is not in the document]
universalmind303: I don't think we should dump everything into `TableOptions`. To me, the `TableOptions` represent the disjunctive union of the create table parameters for that specific discriminant. The intersecting fields are represented in the top level structure.

crates/datasources/src/lib.rs
universalmind303: I just removed these validate methods for now. I think having the tunnel/credentials validate the datasource makes a lot more sense & on further inspection that is what we are already doing. So i just removed those and we can continue to use the current logic.

crates/datasources/src/lib.rs
universalmind303: > If the design means that credentials and tunnel options have to be orthogonal to table options, and we want to handle things like explicit schema orthogonally, then every time we need to add a new feature it sort of seems like we'll have a discussion about 'should this be a new orthogonal configuration category or go in the table options' which doesn't feel productive.

i think there are just a few top level arguments. 

- any datasource can accept a schema (columns). 
- any datasource can require credentials. 
- any datasource can be behind a tunnel.

inversely, 
- not all datasources specify a "format"
- nod all datasources specify a "compression" 

These are all examples of datasource specific arguments. 

The top level ones should be quite limited, and applicable to *any* datasource.

universalmind303:
@tychoish I removed all of the `table_provider_from_db_options` related stuff. I realized I was going against my own words of "keep the changes as minimal as possible". 

We'll probably want some logic for that conversion at some point, but I think it's just noise right now.

testdata/sqllogictests/external_schema.slt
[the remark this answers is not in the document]
tychoish: I don't think that's manageable long term, and I don't think it's reasonable to extend the datasource interface every time there's a new "create" parameter: a sizeable percentage of our roadmap are features that exist for some subset of data sources require some configuration, and for which we could choose to either make top a top level object or configuration _or_ embed in options. (access control, dynamic materialized tables, background vacuuming/partitioning operations, etc.) 

We're also aware that we want to change the syntax and a major goal of this refactoring is to make it easier to implement data sources, this solution proposes: that people developing data-sources need to be aware of a legacy create table syntax, and have to change the interface (and every implementation) anytime a "non-disjoint" configuration is  needed, _and_ we're (apparently) opposed to having default implementations of methods that are unused in most datasource implementations... I think we need a design that is a better fit for the needs of the product.

crates/datasources/src/lib.rs
tychoish: While we can imagine "any datasource can accept <...>" as being true in principle, it's absolutely not true in practice:
https://github.com/glaredb/glaredb/blob/main/crates/sqlbuiltins/src/validation.rs#L63-L124

tychoish:
> 
We'll probably want some logic for that conversion at some point, but I think it's just noise right now.

Definitely, it's also a fringe feature that (I think) is a superset of table providers rather than a property of a table provider. 

Makes sense.

testdata/sqllogictests/external_schema.slt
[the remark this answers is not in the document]
tychoish: > the datasource implementor needs access to all of the possible top level fields (credentials, tunnel, schema, ...)

I think that is not true of all datasource implementers: I don't believe that there is a set of fields or options that are applicable to all data sources. 

- tunnels are only (meaningfully) applicable to data sources that we access over a socket, and currently only implemented for postgres and mysql.
- schemas could be applied anywhere, in theory, though they are mostly applicable to contexts where we do schema inference (json/bson/mongodb/csv): applying them to a relational data source is a thing we have to consider more, and there are some sources (like 
- credentials are only applicable in a set of non-local cases, though credentials are not _always_ used and in many cases sources read credentials out of the `options` block. Also credentials will take a number of different forms depending on the data sources.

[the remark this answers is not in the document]
tychoish: `DatasourceOptions` seems like it's moving in the right direction, though I don't know that it makes sense to make "tunnels" and "credentials" first class citizens in any of this (and also because credential handling and tunnels are pretty-per datasource as a matter of practice.)

crates/datasources/src/lance/mod.rs
```diff
@@ -76,114 +69,69 @@ impl TableProvider for LanceTable {
...
+    fn name(&self) -> &'static str {
+        "lance"
     }
 
-    fn children(&self) -> Vec<Arc<dyn ExecutionPlan>> {
-        Vec::new()
-    }
+    /// Create a new datasource from the provided table options and credentials.
+    /// CREATE EXTERNAL TABLE foo FROM <name> OPTIONS (...) [CREDENTIALS] (...) [TUNNEL] (...)
+    // TODO: the datasource should have control over it's own CredentialsOptions and TunnelOptions
```
scsmithr, line 90: What does this mean?

crates/datasources/src/lib.rs
scsmithr: Why don't we just have this accept complete ownership? Is there a case where we want to use statement options after we create table options?

crates/protogen/proto/metastore/catalog.proto
```diff
@@ -163,7 +163,7 @@ message TableEntry {
   options.TableOptions options = 3;
   optional uint32 tunnel_id = 4;
   SourceAccessMode access_mode = 5;
-  repeated options.InternalColumnDefinition columns = 6;
+  optional common.arrow.Schema schema = 6;
```
scsmithr, line 166: This should remain as internal column definititions (both for backwards compat and we don't want an arrow schema here).

crates/protogen/proto/metastore/service.proto
```diff
@@ -95,7 +96,8 @@ message CreateExternalTable {
   bool if_not_exists = 4;
   optional string tunnel = 5;
   bool or_replace = 6;
-  // next: 7
+  optional common.arrow.Schema table_schema = 7;
```
scsmithr, line 99: Should also be internal column definitions.

crates/datasources/src/lib.rs
scsmithr: After looking at all of this, I'm strongly of the opinion creds/tunnels should be tied to table options.

Passing in creds/tunnels restricts what we're able to do in the future as it relates to the SQL commands themselves (what would happen if I want a multi tunnel hop with aws private link to ssh, or how might I specify backup credentials in the case of rolling creds). These are concerns of the data source, not of the planner. This I think brings up a secondary point of data sources can be hooked up as external tables _or_ external databases, which I think highlights a need for something more along the lines of `DatasourceOptions` and not table specific options.

But at far as avoiding passing in creds/tunnels in directly, I think something like this is reasonable:

```rust
/// Resolve external dependencies for data sources.
pub trait Resolver {
    fn credentials_by_name(&self, name: &str) -> Result<Option<CredentialsOptions>, DatasourceError>;
    fn credentials_by_oid(&self, oid: u32) -> Result<Option<CredentialsOptions>, DatasourceError>;

    fn tunnel_by_name(&self, name: &str) -> Result<Option<TunnelOptions>, DatasourceError>;
    fn tunnel_by_oid(&self, oid: u32) -> Result<Option<TunnelOptions>, DatasourceError>;
}

#[async_trait]
pub trait Datasource: Send + Sync {
    fn name(&self) -> &'static str;

    fn table_options_from_stmt(
        &self,
        opts: &mut StatementOptions,
        resolver: &dyn Resolver,
    ) -> Result<TableOptions, DatasourceError>;


    async fn create_table_provider(
        &self,
        tbl_options: &TableOptions,
        resolver: &dyn Resolver,
    ) -> Result<Arc<dyn TableProvider>, DatasourceError>;
}
```

And elsewhere:

```rust
pub struct CatalogResolver<'a> {
    catalog: &'a SessionCatalog,
}

impl<'a> CatalogResolver<'a> {
    fn credentials_by_name(&self, name: &str) -> Result<Option<CredentialsOptions>, DatasourceError> {...}
    fn credentials_by_oid(&self, oid: u32) -> Result<Option<CredentialsOptions>, DatasourceError> {...}

    fn tunnel_by_name(&self, name: &str) -> Result<Option<TunnelOptions>, DatasourceError> {...}
    fn tunnel_by_oid(&self, oid: u32) -> Result<Option<TunnelOptions>, DatasourceError> {...}
}
```

universalmind303: Sounds good, I was also thinking of passing in a resolver to the datasource trait.

crates/datasources/src/lance/mod.rs
universalmind303: Oh i forgot to remove this. This made more sense in a previous iteration.

crates/datasources/src/lib.rs
tychoish: Is the resolver not just a higher level wrapper around the catalog+session?

scsmithr: It is, but provides two benefits:

- `datasources` won't depend on the catalog directly
- If we have a "secure store" eventually we could just abstract over that too

crates/metastore/Cargo.toml
```diff
@@ -11,6 +11,7 @@ ioutil = { path = "../ioutil" }
 logutil = { path = "../logutil" }
 protogen = { path = "../protogen" }
 sqlbuiltins = { path = "../sqlbuiltins" }
+datasources = { path = "../datasources" }
```
tychoish, line 14: do we need this yet?

crates/protogen/proto/metastore/catalog.proto
```diff
@@ -49,8 +49,13 @@ message CatalogState {
...
   // Metadata for the deployment.
   DeploymentMetadata deployment = 3;
-
   // next: 4
+
+  // The version of the actual catalog implementation.
+  // This is incremented when there are incompatible changes to the physical
+  // catalog layout.
+  // Purposely set to a high number to avoid accidental collisions
+  optional uint32 catalog_version = 2047;
```
tychoish, line 58: might make sense to put this, just for readability at the top of the message definition (I think field number and ordering in the IDL doesn't need to be the same, but I might be wrong).

crates/protogen/proto/metastore/options.proto
```diff
@@ -120,9 +119,14 @@ message StorageOptions {
   map<string, string> inner = 1;
 }
 
+message TableOptionsV1 {
+  string name = 1;
+  bytes options = 2;
+}
```
tychoish, line 125: Seems like it'd be a good idea to add a version number here so that we can iterate on versioning on a per-data source way (without needing to rename data sources.)

crates/protogen/proto/metastore/options.proto
```diff
@@ -204,10 +208,9 @@ message TableOptionsMongo {
   string connection_string = 1;
   string database = 2;
   string collection = 3;
-  repeated InternalColumnDefinition columns = 4;
```
tychoish, line 207: Is this intentional?

crates/protogen/proto/metastore/options.proto
```diff
@@ -242,9 +245,6 @@ message TableOptionsObjectStore {
 
   // Optional: number of records to sample for formats with inferred schema.
   optional int64 schema_sample_size = 5;
-
-  // (optional) explit schema
-  repeated InternalColumnDefinition columns = 6;
```
tychoish, line 247: this seems unintentional.

crates/sqlexec/src/dispatch/external.rs
```diff
@@ -77,6 +77,7 @@ use sqlbuiltins::functions::FunctionRegistry;
 use super::{DispatchError, Result};
 
 /// Dispatch to external tables and databases.
+// TODO: add a `DatasourceRegistry` to the `ExternalDispatcher` to allow for dynamic datasources.
 pub struct ExternalDispatcher<'a> {
```
tychoish, line 81: when do we need to do this todo?

crates/sqlexec/src/planner/session_planner.rs
```diff
@@ -959,17 +917,78 @@ impl<'a> SessionPlanner<'a> {
...
+            .columns
+            .map(|columns| {
+                let fields = columns
+                    .into_iter()
+                    .map(|coll| -> Result<Field, PlanError> {
+                        // check if there is a NOT NULL constraint
+                        let is_nullable = !coll
+                            .options
+                            .into_iter()
+                            .any(|k| matches!(k.option, ColumnOption::NotNull));
```
tychoish, line 972: given the application of user-specified schema for external data sources is that people _probably_ never actually want an _not null_ constraint and I'd be inclined to error early when such schemas are specified and not attempt to enforce the constraint.

tychoish (commented):
few questions and comments. 

I guess this is not a new issue, but the refactoring highlights this issue: 

if users specify "COLUMNS" for a datasource that doesn't support them, then we should error, and I think right now we just continue, and that's probably incorrect but it probably should be follow up to this, unless it's easy/obvious to fix.

crates/protogen/proto/metastore/options.proto
universalmind303: Yes, i thought we agreed that the columns would only be applied to the `TableEntry`, not the `TableOptions`.

crates/sqlexec/src/planner/session_planner.rs
universalmind303: i think that is an assumption we shouldn't make, but I'll make the changes regardless.

crates/sqlexec/src/dispatch/external.rs
universalmind303: will remove.

PR #2770, opened 2024-03-11
refactor(table options): serde 

universalmind303:


tychoish (approved):
talked to @scsmithr about this, and we're both on board.

PR #2773, opened 2024-03-12
fix: Retry writing lease when error code 429 (too many requests).

vrongmeal:
This should fix the issue. We are retrying here as suggested by GCS since writing the same object again and again is rate limited to once per second.

Another thing that might help is writing to a new "temp" object every time (instead of using the same one for the same process).

Fixes: https://github.com/GlareDB/cloud/issues/2746

### DIFF
--- a/crates/metastore/src/storage/lease.rs
+++ b/crates/metastore/src/storage/lease.rs
@@ -27,7 +27,7 @@ use protogen::gen::metastore::storage;
 use protogen::metastore::types::storage::{LeaseInformation, LeaseState};
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinHandle;
-use tracing::{debug_span, error, Instrument};
+use tracing::{debug, debug_span, error, Instrument};
 use uuid::Uuid;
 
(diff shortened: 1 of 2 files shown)


crates/metastore/src/storage/lease.rs
```diff
@@ -339,11 +339,16 @@ impl LeaseRenewer {
...
+    async fn try_write_lease(&self, lease: LeaseInformation) -> Result<()> {
         // Write to storage.
         let proto: storage::LeaseInformation = lease.into();
         let mut bs = BytesMut::new();
         proto.encode(&mut bs)?;
+
+        // TODO: Generate a new temp path everytime? This will definitely avoid
+        // the rate limitation over over-writing the same object here and might
+        // just be enough to delay the "renaming" process so we don't exceed
+        // the limit there as well.
```
vrongmeal, line 351: @scsmithr how do you feel about this?

tychoish (approved):

crates/metastore/src/storage/lease.rs
vrongmeal: I merged this PR before reading this. Will open another to add this as well.

PR #2775, opened 2024-03-12
refactor(table options): backwards compat for catalog

universalmind303:


crates/protogen/proto/metastore/catalog.proto
```diff
@@ -49,8 +49,13 @@ message CatalogState {
...
   // Metadata for the deployment.
   DeploymentMetadata deployment = 3;
-
   // next: 4
+
+  // The version of the actual catalog implementation.
+  // This is incremented when there are incompatible changes to the physical
+  // catalog layout.
+  // Purposely set to a high number to avoid accidental collisions
+  optional uint32 catalog_version = 999;
```
tychoish, line 58: I'm not particularly worried about collisions: any collison that gets past the compiler and code review is buggy in more ways than one, and setting this here only protects against well behaved accidents, and collisions here are definitionally not well behaved. 

Given var-length encoding, and other peculiarities, https://protobuf.dev/programming-guides/proto3/#assigning, I'd be inclined to make this 2047 (the largest two-byte encoded integer)

crates/protogen/src/metastore/types/catalog.rs
```diff
@@ -10,19 +10,28 @@ use super::options::{
...
+pub const CURRENT_CATALOG_VERSION: u32 = 1;
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct CatalogState {
+    /// the version of the catalog state. This increments on every Mutation
     pub version: u64,
     pub entries: HashMap<u32, CatalogEntry>,
     pub deployment: DeploymentMetadata,
+    /// This is the version of the catalog implementation.
+    /// This is incremented when the catalog changes in a way that is not backwards compatible.
```
tychoish, line 33: In these comments it might be useful to clarify that it's the physical representation of the catalog that we're ensuring backwards compatibility, not semantic stuff.

tychoish (approved):
lgtm modulo some nits

PR #2776, opened 2024-03-13
chore: Create a new temp object when writing lease all the time.

vrongmeal:
The temp path is generated with a random UUID as suffix.

### DIFF
--- a/crates/metastore/src/storage/lease.rs
+++ b/crates/metastore/src/storage/lease.rs
@@ -99,12 +99,10 @@ impl RemoteLeaser {
     ///
     /// Errors if some other process holds the lease.
     pub async fn acquire(&self, db_id: Uuid) -> Result<RemoteLease> {
-        let tmp_path = LEASE_INFORMATION_OBJECT.tmp_path(&db_id, &self.process_id);
         let visible_path = LEASE_INFORMATION_OBJECT.visible_path(&db_id);
         let renewer = LeaseRenewer {
             db_id,
             process_id: self.process_id,
-            tmp_path,
             visible_path,
             store: self.store.clone(),
         };
@@ -218,8 +216,6 @@ impl RemoteLease {
 struct LeaseRenewer {
     db_id: Uuid,
     process_id: Uuid,
-    /// Where to store the temporary lease object.
-    tmp_path: ObjectPath,
     /// Path to the visible lease object.
     visible_path: ObjectPath,
     store: Arc<dyn ObjectStore>,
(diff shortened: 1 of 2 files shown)


crates/metastore/src/storage/lease.rs
```diff
@@ -250,6 +246,13 @@ impl LeaseRenewer {
                 });
             }
 
+            if expires_at - LEASE_DURATION + (LEASE_RENEW_INTERVAL / 2) > now {
+                // Don't acquire a new lease. Let the process use the existing
+                // lease. This should help in reducing the number of writes to
+                // the lease object.
+                return Ok(lease.generation);
+            }
+
```
vrongmeal, line 255: @scsmithr @tychoish ^

This should really solve everything...

tychoish: I just want to clarify things:
- is there a way that the same catalog could be modified or accessed by two tasks/threads within the same process? My concern is that this lease could be (unintentionally) shared between two orthogonal processes. 
- what is the case when the lease is (actually) renewed, and in what cases are leases released? Does this change any of those points?

tychoish (commented):
just questions. 

where are the files that would need to be cleaned up in cloud? Is this a change from a previous version.

greyscaled:
> where are the files that would need to be cleaned up in cloud? Is this a change from a previous version.

To me, it seems like they would be in sub-dirs of `databases/<id>/tmp`, based from the code. There is not an existing bg job for this and I appreciate you bringing it up @vrongmeal , creating an issue for it.

crates/metastore/src/storage/lease.rs
vrongmeal: > is there a way that the same catalog could be modified or accessed by two tasks/threads within the same process? My concern is that this lease could be (unintentionally) shared between two orthogonal processes.

So, I just took a better look and `process_id` refers to the service and not session. So, it's most definitely possible that two threads within the same process modify the same catalog! I should undo this change.

> what is the case when the lease is (actually) renewed, and in what cases are leases released? Does this change any of those points?

Lease is renewed after the given interval. A lease is released once the mutation is over (or automatically if it expires).

tychoish: > So, it's most definitely possible that two threads within the same process modify the same catalog! I should undo this change.

Maybe. If this is an existing flaw, and this change doesn't increase the likelihood of a collision, then it's probably safe to commit it. We could also change to something that isn't the `process_id`.

tychoish:
@vrongmeal I think we discussed closing this because the existing solution (retries) works, we have a longer term plan to solve this issue, and this has some larger in-process concerns.

scsmithr:
> @vrongmeal I think we discussed closing this because the existing solution (retries) works, we have a longer term plan to solve this issue, and this has some larger in-process concerns.

What's the concern? The code seems reasonable to me.

vrongmeal:
> What's the concern? The code seems reasonable to me.

@scsmithr does the lease retention seem fine to you? I had some doubts about retaining a lease by the same process as Sam mentioned, a process might try to modify the same catalog from two different threads. But seems like that problem exists from before. We can merge this in as it is and should be OK (after thinking about it for a bit) but want to know your opinion.

tychoish:
> What's the concern? The code seems reasonable to me.

Two concurrent operations on the same instance would both assume that they held the lock.

tychoish (approved):

Issue #2756, opened 2024-03-08
bug: read_json doesn't work with data over http


talagluck:
## Description
I would expect this to work:
```sql
SELECT * FROM read_json('https://opdb.org/api/search/typeahead?q=*');
```

Instead, it raises `External error: Accessing delta tables over http not supported`

tychoish:
This is a known issue: the [object_store](https://docs.rs/object_store/latest/object_store/) crate that we use for accessing data doesn't handle query strings. @vrongmeal has this in his queue.

vrongmeal:
The best I could get is:

```
> select * from read_json('https://opdb.org/api/search/typeahead?q=*');
Error: External error: Missing content-length header
```

I think @universalmind303 has more context about this ^

vrongmeal:
I've made this work:

```
> select * from read_json('http://localhost:3004/?path=test.json');
┌───────┬─────────┐
│   _id │ name    │
│    ── │ ──      │
│ Int64 │ Utf8    │
╞═══════╪═════════╡
│     1 │ Vaibhav │
│     2 │ Tal     │
└───────┴─────────┘
```

But the mentioned URL fails due to lack of "content-length" header. We need some other strategy to handle the case. Maybe download the contents locally into a file (similar to SQLite?)

PS: Query params working?

vrongmeal:
So... have something working!

```
> select * from read_json('https://opdb.org/api/search/typeahead?q=*');
┌─────────────┬──────────────────────────────────┬──────────┬────────────────────────┬──────────────┐
│ id          │ text                             │ name     │ supplementary          │ display      │
│ ──          │ ──                               │ ──       │ ──                     │ ──           │
│ Utf8        │ Utf8                             │ Utf8     │ Utf8                   │ Utf8         │
╞═════════════╪══════════════════════════════════╪══════════╪════════════════════════╪══════════════╡
│ G5W1w-MkPy7 │ "300" (Gottlieb, 1975)           │ "300"    │ Gottlieb, 1975         │ reels        │
│ GrqL6-MLy7l │ 15 Pool (Playmec Flippers, 1976) │ 15 Pool  │ Playmec Flippers, 1976 │ reels        │
│ GRBWj-MQKlO │ 2 in 1 (Bally, 1964)             │ 2 in 1   │ Bally, 1964            │ reels        │
[...]

PR #2784, opened 2024-03-15
fix: Allow HTTP object store for json/bson etc.

vrongmeal:
Fixes: #2756

### DIFF
--- a/crates/datasources/src/object_store/http.rs
+++ b/crates/datasources/src/object_store/http.rs
@@ -1,9 +1,10 @@
 use core::fmt;
 use std::fmt::Display;
+use std::io::SeekFrom;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use bytes::Bytes;
+use bytes::{Bytes, BytesMut};
 use chrono::Utc;
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::datasource::file_format::FileFormat;
@@ -25,7 +26,8 @@ use object_store::{
     PutOptions,
     PutResult,
 };
-use tokio::io::{AsyncWrite, AsyncWriteExt};
+use tempfile::tempdir;
+use tokio::io::{AsyncReadExt, AsyncSeekExt, AsyncWrite, AsyncWriteExt};
 use url::Url;
 use uuid::Uuid;
 
@@ -189,7 +191,9 @@ impl SimpleHttpStore {
             )));
         }
 
-        let download_file_path = std::env::temp_dir().join(format!("{}", Uuid::new_v4()));
+        let download_file_dir = tempdir()?;
+
+        let download_file_path = download_file_dir.path().join(format!("{}", Uuid::new_v4()));
 
         let mut contents = res.bytes_stream();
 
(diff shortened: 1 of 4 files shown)


crates/datasources/src/object_store/http.rs
```diff
@@ -94,17 +98,17 @@ impl ObjStoreAccess for HttpStoreAccess {
...
             .get("Content-Length")
             .and_then(|v| v.to_str().ok())
-            .and_then(|v| v.parse().ok())
-            .unwrap_or_else(|| res.content_length().unwrap_or(0));
-        if len == 0 {
-            return Err(ObjectStoreSourceError::Static(
+            .and_then(|v| v.parse::<u64>().ok())
+            .or_else(|| res.content_length())
+            // TODO: Download the contents of the file and handle using local
+            // store (maybe outside of HTTP store).
```
tychoish, line 108: this is what I did in the `sqlite` from cloud handler, if I understand things correctly. 

Also it seems like, we probably don't want to rely on content-length headers being accurate...

vrongmeal:
> which part of the original issue remains unresolved?

The "content-length" part

crates/datasources/src/object_store/http.rs
```diff
@@ -23,6 +38,36 @@ pub struct HttpStoreAccess {
...
+        let res = reqwest::Client::new().head(u.clone()).send().await?;
+        let status = res.status();
+        if !status.is_success() {
+            if u.as_str().contains('*') {
+                return Err(ObjectStoreSourceError::InvalidHttpStatus(format!(
+                    "Unexpected status code '{}' for url: '{}'. \
+                    Note that globbing is not supported for HTTP.",
+                    status, u,
+                )));
+            }
```
tychoish, line 52: I would have imagined that we'd do this somewhere other than the check for the content length?

crates/datasources/src/object_store/http.rs
```diff
@@ -140,3 +168,193 @@ impl ObjStoreAccess for HttpStoreAccess {
...
+        Ok(GetResult {
+            payload,
+            meta: ObjectMeta {
+                location,
+                last_modified: Utc::now(),
+                size,
+                e_tag: None,
+                version: None,
+            },
+            range: 0..size,
```
tychoish, line 212: is this an array for every byte?

crates/datasources/src/object_store/http.rs
```diff
@@ -140,3 +168,193 @@ impl ObjStoreAccess for HttpStoreAccess {
...
+        let contents = res.bytes().await?;
+
+        let size = contents.len();
+
+        let stream = async_stream::stream! {
+            let res: Result<Bytes, object_store::Error> = Ok(contents);
+            yield res;
+        };
+
+        let payload = GetResultPayload::Stream(Box::pin(stream));
```
tychoish, line 201: my concern here is less that we should write it to a file (or not) but that we actually/unintentionally end up reading the whole stream into memory here... That's the main risk in my mind...

tychoish (approved):
there are definitely some edge cases here that we should think about a bit before merging, but the code seems fine.

crates/datasources/src/object_store/http.rs
```diff
@@ -140,3 +172,232 @@ impl ObjStoreAccess for HttpStoreAccess {
...
+        }
+
+        let file_meta = download_file.metadata().await?;
+        let size = file_meta.len() as usize;
+
+        // Seek the downloaded file to start (before reading).
+        download_file.seek(SeekFrom::Start(0)).await?;
+
+        let stream = async_stream::stream! {
+            // Delete when the stream is dropped.
```
tychoish, line 221: don't we need to call this with move? I've been tending to use the `futures::stream::once()` rather than the macro to make this more clear, but if it's in an `Arc` then I think it ends up working ok?

tychoish (approved):

crates/datasources/src/object_store/http.rs
vrongmeal: The macro expands to `async move`. Moreover, the compiler would just error if it didn't here.

PR #2793, opened 2024-03-15
fix: resolve aliased catalogs

### LINKED ISSUE #2779: bug: infer db name on SELECT and not just on CREATE
## Description
This is a strange issue I'm hitting when trying to run dbt with GlareDB Cloud. The short of it is that this code works:

```sql
  create or replace view "empty_rain"."public"."my_view"
```
but this code does not:

```sql
select *
from "empty_rain"."public"."my_view"
```
where `"empty_rain"` is the name of my GlareDB Cloud deployment.

For the SELECT query, this **will** work:

```sql
select *
from "default"."public"."my_view"
```

So essentially, we have `deployment name` -> `db name` (default) -> `schema name` -> `table or view name`.

So I think what is happening is that when we create a table or a view, when we pass in `deployment_name`, we are able to assume the `default` DB name. But when we try this in a `SELECT` statement, it does not work. Incidentally, the below code also DOES NOT work:

```sql
select *
from "empty_rain"."default"."public"."my_view"
```

All of this matters, because when I configure my credentials for dbt, it asks me for my database name. When I configure this, I need to pass in my deployment name, or else I get a connection error. But if I try and select from dbt_model_A in dbt_model_B, when dbt compiles my queries, it looks for any existing tables at "empty_rain"."public"."dbt_model_A", which fails.

universalmind303:
closes #2779 

Note for Reviewers: 

There isn't a real good way that I know of to add a test for this, so i will just share the steps I took to manually test it. 

1. go to console.glaredb.com
2. create a view in there: `create view "my_view" as select 1;`
3. connect to cloud via local
```sh
> just run local
> \open glaredb://<USER>:<PASS>@o_xTz3cuY.remote.glaredb.com:6443/quiet_frog
Connected to Cloud deployment (TLS enabled): quiet_frog
> select * from "quiet_frog"."public"."my_view";
┌──────────┐
│ Int64(1) │
│       ── │
│    Int64 │
╞══════════╡
│        1 │
└──────────┘
```

it also works the other way around. Previously a create statement `create view quiet_frog.public.my_view_2` would save it to the "default" db, now it properly saves it under the alias name

```sql
> create view "quiet_frog"."public"."my_view_2" as select 2;
View created
> select * from my_view_2;
┌──────────┐
│ Int64(2) │
│       ── │
│    Int64 │
╞══════════╡
│        2 │
└──────────┘
> select * from "quiet_frog"."public"."my_view_2";
┌──────────┐
│ Int64(2) │
│       ── │
│    Int64 │
╞══════════╡
│        2 │
└──────────┘
>
```

### DIFF
--- a/crates/catalog/src/session_catalog.rs
+++ b/crates/catalog/src/session_catalog.rs
@@ -47,6 +47,8 @@ pub struct ResolveConfig {
 /// from the remote state provided by metastore.
 #[derive(Clone, Debug)]
 pub struct SessionCatalog {
+    /// Optional alias for referencing objects in this catalog
+    alias: Option<String>,
     /// The state retrieved from a remote Metastore.
     state: Arc<CatalogState>,
     /// Map database names to their ids.
@@ -69,6 +71,7 @@ impl SessionCatalog {
     /// Create a new session catalog with an initial state.
     pub fn new(state: Arc<CatalogState>, resolve_conf: ResolveConfig) -> SessionCatalog {
         let mut catalog = SessionCatalog {
+            alias: None,
             state,
             database_names: HashMap::new(),
             tunnel_names: HashMap::new(),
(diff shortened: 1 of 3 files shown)


scsmithr (approved):

vrongmeal:
Should we also restrict creating an external database with the same name as catalog alias? Seems like something that could lead to conflicts/issues with not being able to access an external source.

Issue #2746, opened 2024-03-06
bug: InternalError: (psycopg2.errors.InternalError_) Casting expressions to regclass unsupported


talagluck:
## Description
See also #1577 

When attempting to use Great Expectations with GlareDB (using a Postgres connection string), I get
```
InternalError: (psycopg2.errors.InternalError_) Casting expressions to regclass unsupported
```

Full error:
```shell
E               [SQL: SELECT pg_catalog.pg_attribute.attname AS name, pg_catalog.format_type(pg_catalog.pg_attribute.atttypid, pg_catalog.pg_attribute.atttypmod) AS format_type, (SELECT pg_catalog.pg_get_expr(pg_catalog.pg_attrdef.adbin, pg_catalog.pg_attrdef.adrelid) AS pg_get_expr_1 
E               FROM pg_catalog.pg_attrdef 
E               WHERE pg_catalog.pg_attrdef.adrelid = pg_catalog.pg_attribute.attrelid AND pg_catalog.pg_attrdef.adnum = pg_catalog.pg_attribute.attnum AND pg_catalog.pg_attribute.atthasdef) AS "default", pg_catalog.pg_attribute.attnotnull AS not_null, pg_catalog.pg_class.relname AS table_name, pg_catalog.pg_description.description AS comment, pg_catalog.pg_attribute.attgenerated AS generated, (SELECT json_build_object(%(json_build_object_2)s, pg_catalog.pg_attribute.attidentity = %(attidentity_1)s, %(json_build_object_3)s, pg_catalog.pg_sequence.seqstart, %(json_build_object_4)s, pg_catalog.pg_sequence.seqincrement, %(json_build_object_5)s, pg_catalog.pg_sequence.seqmin, %(json_build_object_6)s, pg_catalog.pg_sequence.seqmax, %(json_build_object_7)s, pg_catalog.pg_sequence.seqcache, %(json_build_object_8)s, pg_catalog.pg_sequence.seqcycle) AS json_build_object_1 
E               FROM pg_catalog.pg_sequence 
E               WHERE pg_catalog.pg_attribute.attidentity != %(attidentity_2)s AND pg_catalog.pg_sequence.seqrelid = CAST(CAST(pg_catalog.pg_get_serial_sequence(CAST(CAST(pg_catalog.pg_attribute.attrelid AS REGCLASS) AS TEXT), pg_catalog.pg_attribute.attname) AS REGCLASS) AS OID)) AS identity_options 
[...]

PR #2795, opened 2024-03-18
chore: add better regclass and oid cast support

universalmind303:
partially addresses https://github.com/GlareDB/glaredb/issues/2746


testdata/sqllogictests/cast/oid.slt
```diff
@@ -0,0 +1,30 @@
+
+statement error
```
tychoish, line 2: should we put the error message here?

testdata/sqllogictests/cast/oid.slt
```diff
@@ -0,0 +1,30 @@
...
+
+statement ok
+select cast('my_table' as OID);
+
+
+# We don't support expressions other than strings right now.
+statement error Casting expressions to oid unsupported
+select cast(1 as oid);
+
+statement error Casting expressions to oid unsupported
```
tychoish, line 26: should this error message be phrased as "only strings can be cast to OIDs"

testdata/sqllogictests/cast/oid.slt
```diff
@@ -0,0 +1,30 @@
...
+select 'my_table'::OID;
+
+statement ok
+select cast('my_table' as oid);
+
+statement ok
+select cast('my_table' as OID);
+
+
+# We don't support expressions other than strings right now.
```
tychoish, line 22: is there a point to supporting more ever/eventually?

tychoish (commented):
few questions, seems reasonable. 

(shouldn't this be a fix not a chore?)

testdata/sqllogictests/cast/oid.slt
universalmind303: idk, most of this was copied from the `regclass` stuff, so i didn't want to change any more than absolutely necessary to implement the new functionality

tychoish:
The existing regclass error is only used once, and is ambiguous from the user's perspective (and ours, when debugging) if they ever run into it, particularly now that we (presumably) support integer and string casts. If it's easy enough to provide a bit more context, it seems worthwhile. 

I don't see any particular reason to treat our error messages as a stable API, unless I'm missing something or this is a standard/copied error message.

vrongmeal (approved):

Issue #1065, opened 2023-05-31
Unable to add gcs and s3 objects as external tables


scsmithr:
## Context

Outside google cloud:
```
2023-05-31T17:42:49.731656Z ERROR                 main ThreadId(01) testing::slt::cli: crates/testing/src/slt/cli.rs:244: Error while running test `sqllogictests/demos/replicant` error=test fail: statement failed: db error: ERROR: External table validation failed: Generic GCS error: Error performing get request tyrell/voight_kampff.csv: response error "{
  "error": {
    "code": 401,
    "message": "Invalid Credentials",
    "errors": [
      {
        "message": "Invalid Credentials",
        "domain": "global",
        "reason": "authError",
        "locationType": "header",
        "location": "Authorization"
      }
    ]
  }
}
", after 0 retries: HTTP status client error (401 Unauthorized) for url (https://storage.googleapis.com/storage/v1/b/glaredb%2Ddemos/o/tyrell%2Fvoight%5Fkampff%2Ecsv?alt=json)
```

Inside google cloud:
```
External table validation failed: Generic GCS error: Error performing get request tyrell/voight_kampff.csv: response error "{ "error": { "code": 412, "message": "The type of authentication token used for this request requires that Uniform Bucket Level Access be enabled.", "errors": [ { "message": "The type of authentication token used for this request requires that Uniform Bucket Level Access be enabled.", "domain": "global", "reason": "conditionNotMet", "locationType": "header", "location": "If-Match" } ] } } ", after 0 retries: HTTP status client error (412 Precondition Failed) for url (https://storage.googleapis.com/storage/v1/b/glaredb%2Ddemos/o/tyrell%2Fvoight%5Fkampff%2Ecsv?alt=json)
```

The object_store crate tries to dial the metadata service when not provided application default credentials or a service account. For objects open to public, this is unnecessary, but the version we're using (0.5.6) does not have a way of disabling that dial.

Version 0.6 does allow slotting in a custom authenticator which we could used. However, datafusion is stuck on 0.5.6 right now, so we're kinda stuck for now.

vrongmeal:
This works with the updated version. No changes are required, right?

```
glaredb=> create external table def from gcs options ( bucket = 'vrongmeal-public-test', location = 'userdata1.parquet' );
CREATE TABLE
```

vrongmeal:
I added a `NullCredentailProvider` to resolve this.

[...]

PR #2798, opened 2024-03-20
fix: Support null credentials for object store access.

vrongmeal:
GCS works (raising a PR for NULL Credentials):

```
> select * from 'gs://vrongmeal-public-test/data.csv';
┌───────┬─────────┐
│    id │ name    │
│    ── │ ──      │
│ Int64 │ Utf8    │
╞═══════╪═════════╡
│     1 │ vaibhav │
│     2 │ sean    │
│     3 │ grey    │
└───────┴─────────┘
```

Need to make a similar change upstream for S3 and Azure.

Fixes: #1065

### DIFF
--- a/crates/datasources/src/object_store/azure.rs
+++ b/crates/datasources/src/object_store/azure.rs
@@ -2,10 +2,11 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::sync::Arc;
 
+use async_trait::async_trait;
 use datafusion::execution::object_store::ObjectStoreUrl;
-use object_store::azure::{AzureConfigKey, MicrosoftAzureBuilder};
+use object_store::azure::{AzureConfigKey, AzureCredential, MicrosoftAzureBuilder};
 use object_store::path::Path as ObjectStorePath;
-use object_store::ObjectStore;
+use object_store::{CredentialProvider, ObjectStore};
 use protogen::metastore::types::options::StorageOptions;
 
 use super::errors::{ObjectStoreSourceError, Result};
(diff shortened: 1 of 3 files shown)


crates/datasources/src/object_store/azure.rs
```diff
@@ -70,18 +71,38 @@ impl ObjStoreAccess for AzureStoreAccess {
...
+            if matches!(
+                key,
+                AzureConfigKey::AccountName
+                    | AzureConfigKey::AccessKey
+                    | AzureConfigKey::ClientId
+                    | AzureConfigKey::ClientSecret
+                    | AzureConfigKey::SasKey
+                    | AzureConfigKey::Token
+            ) {
+                creds = true;
```
tychoish, line 86: should break here?

crates/datasources/src/object_store/azure.rs
```diff
@@ -70,18 +71,38 @@ impl ObjStoreAccess for AzureStoreAccess {
...
         }
 
         if let Some(access_key) = &self.access_key {
+            creds = true;
             builder = builder.with_access_key(access_key);
         }
 
+        if !creds {
+            builder = builder.with_credentials(Arc::new(NullCredentialProvider));
+        }
```
tychoish, line 104: could we create a build with null credentials, and then override it if/as we find credentials, and avoid the bool?

tychoish (commented):
this pr includes changes for gcs which is inconsistent with the PR description, just confused about what needs to be in the upstream vs here?

universalmind303 (approved):

PR #2802, opened 2024-03-20
fix: properly alias all sessions

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2779

manually tested in cloud via:

```sql
create view "broken_sky"."public"."my_view" as select 1;
```

```sql
select * from "broken_sky"."public"."my_view";
```

### DIFF
--- a/crates/sqlexec/src/engine.rs
+++ b/crates/sqlexec/src/engine.rs
@@ -420,8 +420,6 @@ impl Engine {
         vars: SessionVars,
         storage: SessionStorageConfig,
     ) -> Result<TrackedSession> {
-        println!("new_local_session_context");
-
         let session = self.new_untracked_session(vars, storage).await?;
 
         let prev = self.session_counter.fetch_add(1, Ordering::Relaxed);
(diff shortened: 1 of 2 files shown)


crates/sqlexec/src/engine.rs
```diff
@@ -446,14 +446,26 @@ impl Engine {
...
         );
 
+        // If the database name is in the form of `<UUID>/<db_id>`, then we
+        // should use the <db_id> as the alias for the database.
+        let database_name = vars.database_name();
+        if let Some((org_id, db_id)) = database_name.split_once('/') {
+            if Uuid::parse_str(org_id).is_ok() {
+                catalog = catalog.with_alias(db_id.to_string());
+            }
+        // else, we should use the entire database name as the alias if it's not empty
```
tychoish, line 464: what happens in the the case that there's a problem parsing the org id? No alias? shouldn't we just ignore the org_id?

tychoish (commented):
one question, other than that this seems good.

crates/sqlexec/src/engine.rs
universalmind303: are you referring to the edge case where the dbname contains a `/` but is not `<uuid>/db_id` from cloud?

tychoish: yes

tychoish (approved):

PR #2806, opened 2024-03-20
fix: reduce flightsql test parallelism for minio

tychoish:


greyscaled:
> Looks like main passed; let's hold off on merging this ...

it passed after I re-ran

PR #2810, opened 2024-03-21
fix: Iceberg fixes for reading table metadata

### LINKED ISSUE #2603: error reading iceberg table
image

data can be read fine using duckdb

vrongmeal:
Fixes #2603


tychoish:
What more do we need here? (I think there's some more test coverage that we could use here?)

vrongmeal:
> What more do we need here? (I think there's some more test coverage that we could use here?)

Added test coverage. There are issues though. If we overwrite the existing data (say running the `generate_pyiceberg.py` again without deleting the existing data), GlareDB returns twice the number of rows whereas the data is appropriately overwritten according to iceberg.

testdata/generate_pyiceberg.py
```diff
@@ -0,0 +1,55 @@
+"""
+How to run this script:
+======================
+
+$ python3 -m venv venv
+$ source venv/bin/activate
+$ pip install "pyiceberg[pyarrow,sql-sqlite]"
+$ pip install botocore
+$ python ./testdata/generate_pyiceberg.py
```
tychoish, line 9: I  think the data set ends up being pretty big, and while I don't think it _really_ matters, I  think it'd be good to avoid just committing lots of test data to the repo _just cause_, when we have other options (e.g. uploading the data to the test bucket, or generating as part of a fixture in one of the pytests).

vrongmeal: Yeah, thinking the same thing.

vrongmeal:
@tychoish made the required changes for pytest

testdata/iceberg/source_data/yellow_tripdata_2023-01.parquet
tychoish, line 1: 45mb feels big for a test file.

tychoish (commented):
looks great, one concern about the size of test data. 

the current test failure is orthogonal to this PR and I can push through it. 

is there anything that needs to be done here?

testdata/iceberg/source_data/yellow_tripdata_2023-01.parquet
vrongmeal: Yeah, I'll upload it on GCS. We have a script to handle big data files.

vrongmeal: It was a simple HTTP url, added it into the fixture to download the data.

tychoish (approved):

Issue #2811, opened 2024-03-21
`COPY TO` perf discrepancies


universalmind303:
## Description
> basically a bunch of benchmark tests for copying 30,000,000 rows of data to a file with Clickhouse, DuckDB, DataFusion, etc. He linked to a Google Collab notebook where anybody can run it. I tried it with GlareDB in the same notebook, and it looked like it was comparatively really fast (and faster even than DataFusion), and I was wondering how that could be possible. I ran this code:

```py
%%time
import glaredb as gdb
con = gdb.connect()
con.sql('''COPY  (select * from './lineitem/*.parquet' order by l_shipdate ) to './sorted_glaredb.parquet' ''')
```

https://www.linkedin.com/posts/mimounedjouallah_duckdb-pyarrow-clickhouse-activity-71723891

PR #2814, opened 2024-03-21
fix: properly satisfy enforcesorting rules

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2811

### DIFF
--- a/crates/datafusion_ext/src/metrics.rs
+++ b/crates/datafusion_ext/src/metrics.rs
@@ -8,9 +8,9 @@ use std::task::{Context, Poll};
 
 use datafusion::arrow::datatypes::{Schema, SchemaRef};
 use datafusion::arrow::record_batch::RecordBatch;
-use datafusion::error::Result;
+use datafusion::error::{DataFusionError, Result};
 use datafusion::execution::TaskContext;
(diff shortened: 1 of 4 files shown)


crates/datafusion_ext/src/metrics.rs
```diff
@@ -248,6 +256,75 @@ impl<T: DataSourceMetricsOptsType> ExecutionPlan for DataSourceMetricsExecAdapte
...
+    }
+
+    fn required_input_ordering(&self) -> Vec<Option<Vec<PhysicalSortRequirement>>> {
+        // the input ordering must match the output ordering of the source
+        // without this, the sort may get removed by the optimizer
+        vec![self
+            .output_ordering()
+            .map(PhysicalSortRequirement::from_sort_exprs)]
+    }
+}
```
universalmind303, line 296: the `WriteOnlyDataSourceMetricsExecAdapter` needs an explicit check on the input ordering, otherwise the `EnforceSorting` rule may remove it. The `ReadOnly...` doesn't care, so this needed to be split out. 

This states that if the child has an explicit `output_ordering`, then this node needs a matching `input_ordering`.

tychoish (approved):

Issue #2796, opened 2024-03-18
unable to `drop credential`


universalmind303:
## Description
```sql
> CREATE CREDENTIAL debug_creds PROVIDER debug
::: OPTIONS (table_type = 'never_ending');

> DROP CREDENTIAL debug_creds;
Error: sql parser error: Expected TABLE, VIEW, INDEX, ROLE, SCHEMA, FUNCTION, STAGE or SEQUENCE after DROP, found: credential
```

PR #2823, opened 2024-03-25
feat: deprecate "drop credentials"

universalmind303:
closes https://github.com/GlareDB/glaredb/issues/2796

```sh
> drop credential if exists debug_creds;
Credentials dropped
> drop credentials exists debug_creds;
Credentials dropped
DEPRECATION WARNING: `DROP CREDENTIALS` is deprecated and will be removed in a future release. Use `DROP CREDENTIAL` instead.
```

## Note for reviewers

The deprecation message itself is purposely not tested in the SLT's as it will be removed in a future release.

### DIFF
--- a/crates/sqlexec/src/planner/physical_plan/drop_credentials.rs
+++ b/crates/sqlexec/src/planner/physical_plan/drop_credentials.rs
@@ -20,13 +20,21 @@ use datafusion::physical_plan::{
 use futures::stream;
 use protogen::metastore::types::service::{self, Mutation};
 
-use super::{new_operation_batch, GENERIC_OPERATION_PHYSICAL_SCHEMA};
+use super::{
+    new_operation_batch,
+    new_operation_with_deprecation_batch,
+    GENERIC_OPERATION_PHYSICAL_SCHEMA,
+};
+
+const DEPRECATION_MESSAGE: &str = "`DROP CREDENTIALS` is deprecated and will be removed in a future release. Use `DROP CREDENTIAL` instead.";
 
 #[derive(Debug, Clone)]
 pub struct DropCredentialsExec {
     pub catalog_version: u64,
     pub names: Vec<String>,
     pub if_exists: bool,
+    /// If true, show a deprecation warning when the operation is executed.
+    pub show_deprecation_warning: bool,
 }
 
 impl ExecutionPlan for DropCredentialsExec {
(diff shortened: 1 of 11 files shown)


crates/sqlexec/src/planner/physical_plan/drop_credentials.rs
```diff
@@ -20,13 +20,22 @@ use datafusion::physical_plan::{
...
+
+const DEPRECATION_MESSAGE: &str = "`DROP CREDENTIALS` is deprecated and will be removed in a future release. Use `DROP CREDENTIAL` instead.";
 
 #[derive(Debug, Clone)]
 pub struct DropCredentialsExec {
     pub catalog_version: u64,
     pub names: Vec<String>,
     pub if_exists: bool,
+    /// If true, show a deprecation warning when the operation is executed.
+    pub show_deprecation_warning: bool,
```
tychoish, line 38: To my mind, the deprecation is a static/implementation property of the type, and "show deprecation warning" is a property of the session/client and not the physical plan. 

Is there a reason that I've missed that means that the display/output of the struct has to be mutable. 

-----

I've read this again, and I think my confusion is that this struct is `DropCredntials` and should be renamed to match the output. I get why this is here, but it feels like it makes it difficult to generalize, and that we'll have to do this work basically every time we want to change syntax or deprecate something.

tychoish (commented):
- how long is our deprecation period: we should note what that is on the ticket and in the code. (it can be the first 0.x.0 release after <date> July 1 2024. (or whatever)). 
- will we want to have a period where deprecation warnings receive a special "redirecting" error message? relative to the generic "this syntax isn't supported" 
- would it make sense to have a CLI flag, or connection mode that turns deprecations into errors so that people can validate that their applications don't use deprecated methods?

Just a couple of questions and a proposed rename

tychoish:
(the test failure is material to the PR, was checking because of the minio failures we see... it's interesting to me that this only impacted minio. I'm not sure why this is only picked up on minio though)

crates/sqlexec/src/planner/physical_plan/drop_credentials.rs
universalmind303: yeah I considered renaming it to `DropCredential`, but didn't want to overcomplicate the PR. Will update with the rename.

PR #2825, opened 2024-03-25
feat: query google sheet

hailelagi:
WIP

### DIFF
--- a/crates/datasources/src/google_sheet/client.rs
+++ b/crates/datasources/src/google_sheet/client.rs
@@ -1,0 +1,9 @@
+// TODO: wrap the google cloud console API or find a good third party client in rs
+// there is: https://crates.io/crates/sheets but that means using OAuth
+// OR curl https://sheets.googleapis.com/v4/spreadsheets/1f5epAPxP_Yd3g1TunEMdtianpVAhKS0RG6BKRDSLtrk\?key=<API_KEY>
+// a small wrapper with reqwest might not be too involved
+
+// see: https://developers.google.com/sheets/api/samples/reading
+// see: https://developers.google.com/apis-explorer/#p/sheets/v4/
+
+// stream over HTTP/JSON or re-use gRPC interface?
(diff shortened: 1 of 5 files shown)


talagluck (commented):
I'm also wondering if we could default the output type to CSV if a user passes in a general Google Sheets link. It looks like CSV is probably the most general purpose output format (the only other data formats seem to be TSV and XLSX).

Similar to read_csv, we will probably want to default to assuming there are headers, and allow the user to specify if there are none.

One last thing which is a nice-to-have, but not critical, would be allowing the user to specify the cell range for the data (e.g.

```sql
SELECT COUNT(*) FROM read_google_sheet('https://docs.google.com/spreadsheets/d/1f5epAPxP_Yd3g1TunEMdtianpVAhKS0RG6BKRDSLtrk/', 'January', 'A1:AF20');
```

talagluck:
Ah, just to call this out with explicit keyword arguments also:
```sql
SELECT COUNT(*) FROM read_google_sheet(
  link => 'https://docs.google.com/spreadsheets/d/1f5epAPxP_Yd3g1TunEMdtianpVAhKS0RG6BKRDSLtrk/', 
  sheet => 'January', 
  range => 'A1:AF20',
  file_format => 'csv'
)
```

I'm open to feedback on the keyword names.

hailelagi:
these generally seem like reasonable suggestions and I'll try to work them in. Perhaps rather than `link` it'd be maybe `location`? and a few others but let's see as the pr evolves

talagluck:
Sounds good. I agree that `link` isn't ideal and think that `location` is better. I also like `url` or `google_sheets_url`. But we can wait on that decision.

hailelagi:
accidentally closed this when I forced a rebase to catchup to main, this has been in my todo for a bit too long, I plan on taking a look at some point soon this week, almost certainly definitely maybe

hailelagi:
so, I gave this another good long look this morning, and have decided to close this. While the functionality itself is not necessarily hard to add, there's currently far too many moving parts architecturally involved in adding a datasource that I have to tip-toe around -- seemingly parsing out table options like `range` and allowing for either an `oauth` or api key is... not fun. That said big fan of glaredb, I'll likely revisit in awhile <3 and see what's new :)

PR #2831, opened 2024-03-28
chore: engine constructor cleanup

tychoish:


crates/sqlexec/src/engine.rs
```diff
@@ -323,6 +323,15 @@ pub struct Engine {
     _task_executors: Vec<TaskExecutor>,
 }
 
+pub enum EngineBackend {
```
scsmithr, line 326: I'd probably go with EngineStorage or similar. I originally read this as we were going to be connecting a remote engine, when really when just using remote storage.

scsmithr (approved):

PR #2832, opened 2024-03-28
feat: glaredb driver implementation

tychoish:
building up parts of the sdk constructor, will do a bit more here.


crates/glaredb/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+        Operation {
+            op: OperationType::Prql,
+            query: query.into(),
+            conn: Arc::new(self.clone()),
+            schema: None,
+        }
+    }
+}
+
+pub struct RecordStream(Pin<Box<dyn Stream<Item = Result<RecordBatch, DataFusionError>> + Send>>);
```
scsmithr, line 175: I would have this wrap a `SendableRecordBatchStream` since that's really just an alias for this. The `From` impl below will result in this the stream being double boxed, which probably isn't what we want.

crates/glaredb/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+    query: String,
+    conn: Arc<Connection>,
+    schema: Option<Arc<Schema>>,
+}
+
+impl Debug for Operation {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "Operation<{:?}>(\"{:?}\")", self.op, self.query)
+    }
+}
```
scsmithr, line 229: Recommend using the `debug...` methods instead of manually formatting these: https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct

crates/glaredb/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+    Execute,
+}
+
+#[derive(Clone)]
+#[must_use = "operations do nothing unless call() or execute() run"]
+pub struct Operation {
+    op: OperationType,
+    query: String,
+    conn: Arc<Connection>,
+    schema: Option<Arc<Schema>>,
```
scsmithr, line 222: What's schema for?

crates/glaredb/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+        write!(f, "Operation<{:?}>(\"{:?}\")", self.op, self.query)
+    }
+}
+
+impl Operation {
+    pub fn schema(&self) -> Option<Arc<Schema>> {
+        self.schema.clone()
+    }
+
+    pub async fn execute(&mut self) -> Result<SendableRecordBatchStream, ExecError> {
```
scsmithr, line 236: What's the rationale for `&mut self` instead of `&self`?

Personally I'd expect an operation to be executed (or called) at most once, but I don't feasibility of this accepting a `self`.

crates/sqlexec/src/context/local.rs
```diff
@@ -217,12 +217,12 @@ impl LocalSessionContext {
         Ok(())
     }
 
-    pub fn register_env_reader(&mut self, env_reader: Box<dyn EnvironmentReader>) {
-        self.env_reader = Some(env_reader);
+    pub fn register_env_reader(&mut self, reader: Option<Arc<Box<dyn EnvironmentReader>>>) {
+        self.env_reader = reader;
```
scsmithr, line 221: I believe this can be changed back if we make the above change to have `connect` accept `self` (which would remove the need for an Arc).

crates/sqlexec/src/engine.rs
```diff
@@ -321,6 +322,7 @@ pub struct Engine {
     task_scheduler: Scheduler,
     /// Task executors.
     _task_executors: Vec<TaskExecutor>,
+    _tmp_dir: Option<tempfile::TempDir>,
```
scsmithr, line 325: Remove underscore (also should probably remove underscore for `_task_executors` too).

crates/sqlexec/src/remote/client.rs
```diff
@@ -122,18 +122,27 @@ pub struct AuthenticateClientError {
     pub msg: String,
 }
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
```
scsmithr, line 125:
```suggestion
#[derive(Debug, Clone, Copy)]
```

crates/glaredb/src/lib.rs
tychoish: It was mutable, because of the schema thing, which was for an earlier iteration where the operation itself was a stream.

I've addressed that part of this, leaving the "re-executable" as a discussion (we can also rename `Output` to something else, as I think the naming is just bad, `Execution` makes sense, but that has the word "execute" three times in a single options which seems beyond the pail too.

crates/sqlexec/src/engine.rs
tychoish: done, `_task_executors` still isn't used

crates/glaredb/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+    #[builder(setter(strip_option))]
+    pub disable_tls: Option<bool>,
+    #[builder(default = "Some(\"https://console.glaredb.com\".to_string())")]
+    #[builder(setter(into, strip_option))]
+    pub cloud_addr: Option<String>,
+    #[builder(default = "Some(RemoteClientType::Cli)")]
+    #[builder(setter(strip_option))]
+    pub client_type: Option<RemoteClientType>,
+    #[builder(setter(strip_option))]
+    pub environment_reader: Option<Arc<Box<dyn EnvironmentReader>>>,
```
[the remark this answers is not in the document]
tychoish: otherwise it's not cloneable, and in order to pass the same environment reader from the language environment, to the builder/options struct to the session it needs to be cloneable. Previously we just constructed the python env reader during connection.

I tried removing the arc, and I suppose we could pass it in without an arc to the connection object but that feels like it would happen "late", and the Arc seems clean.

crates/glaredb/src/lib.rs
```diff
@@ -0,0 +1,396 @@
...
+            location: None,
+            connection_target: None,
+            ..Default::default()
+        }
+    }
+
+    pub fn with_env_reader(&mut self, reader: Arc<Box<dyn EnvironmentReader>>) -> &mut Self {
+        self.environment_reader = Some(reader);
+        self
+    }
```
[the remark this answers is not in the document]
tychoish: was vestigial, omitted.

crates/glaredb/src/lib.rs
tychoish: vestigial, removed

crates/glaredb/src/lib.rs
tychoish: yeah, I moved this up: because Operation wraps (ultimately)  Engine, the derive-able implementation didn't work. Anyway, I removed it from here and implemented debug for engine in a way that was more reasonable.

crates/sqlexec/src/context/local.rs
tychoish: I think we need the arc in order to allow it to be cloneable, and passed in from the bindings' package.

crates/glaredb/src/lib.rs
tychoish: ok, having tried to unwind this the train of thought is:
- we need to be able to treat operations as tables, and implement table provider. the requirements of this functionality require:
  - the operation to know what the output schema is (lazily)
  - be re-runable in case it's reused in different operations.
- the partition wrapper needs to be sync, and so we can't just pass the stream in directly to it.

This explains all of the fraught Arc/Mutex shenanigans.
- the operation needs to be mutable so that we can set schema when the operation evaluates not when it's called.
- the table object needs to be sync

All of this adds up to the current implementation.

crates/sqlexec/src/context/local.rs
tychoish: I think the arc makes a lot of sense here because ideally you never want a situation where the environment reader, once set, is duplicated...

scsmithr (approved):

PR #2848, opened 2024-03-29
feat: iterator helpers for driver interfaces

tychoish:
this is just a little thing that I didn't want to get in the way of anything else. Seems like this API would be useful in a number of cases.

### DIFF
--- a/crates/glaredb/src/lib.rs
+++ b/crates/glaredb/src/lib.rs
@@ -333,3 +333,44 @@ impl Into<RecordStream> for Result<SendableRecordBatchStream, DataFusionError> {
         }
     }
 }
+
+impl RecordStream {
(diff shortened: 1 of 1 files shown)


tychoish:
> What's the use case? Using `block_on` for getting batches from a stream probably isn't what we want to be doing.

yeah I pulled this out of the previous PR because I thought it would be more controversial.

I wanted to have an async and a sync interface, just for ease of use (mostly these are going to be used in testing/etc.... this (relative to the other PRs in this sequence is optional)

scsmithr:
I don't think it's worth trying to have a sync interface for this. `block_on` can easily lead to deadlocks and generally pretty inefficient here (`all` would block the entire thread even if the stream isn't producing batches).

`all` could also be replaced with `collect`: https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.collect (still async though)

PR #2850, opened 2024-04-01
chore: integrate new rust sdk into python bindings

tychoish:
This is stuck in some borrow checker fun times. My attempt to collapse the `PyLogicalPlan` and the `PyExecutionResult` seems to have run into a wall. 

The crux of the problem is that that:
- `PyObjects` need to be `+ Clone + Sync`
- `SendableRecordStream` don't work with that, and all of my attempts to hack around this haven't succeeded.


bindings/python/src/logical_plan.rs
```diff
@@ -1,86 +1,191 @@
...
+use crate::util::pyprint;
 
 #[pyclass]
-#[derive(Clone, Debug)]
 pub struct PyLogicalPlan {
-    pub(super) lp: LogicalPlan,
-    pub(super) session: PyTrackedSession,
-    pub(super) op: OperationInfo,
+    schema: SchemaRef,
+    stream: SendableRecordBatchStream,
```
scsmithr, line 30: I don't understand why there's a stream on this since we only get a stream after executing the plan.

scsmithr:
> My attempt to collapse the PyLogicalPlan and the PyExecutionResult seems to have run into a wall.

What's the rationale for this? These feel like two distinct things to me.

tychoish:
> > My attempt to collapse the PyLogicalPlan and the PyExecutionResult seems to have run into a wall.
> 
> What's the rationale for this? These feel like two distinct things to me.

Well, sort of, (right?): 
- if it's a DDL op then you get this kind of fake plan, because we don't want DDL changes to be lazy.
- from the caller's perspective, the result of an operation is [at least conceptually] a thunk (lazy value), the difference between a plan and a stream that executes the plan on the first op is not so great. At least in my understanding streams shouldn't do work before they're iterated on/called. 

We could back out, and pass the inner driver, and the query string (but then we'd have to replicate the logic about "is this a DDL," in the python/client layer, which means needing a lower-level implementation inside of the connection, which we can do but that means a slightly different architecture all around, (totally fine)

bindings/python/src/logical_plan.rs
tychoish: https://github.com/GlareDB/glaredb/pull/2832/files#diff-87aa6c7f95143a14b534fe4e79556ba4cfe9df4467c18b12f1d6d578e4882f50R130

universalmind303: I agree, the `PyLogicalPlan` should be a thin wrapper around a `LogicalPlan`, not a `SendableRecordBatchStream`

tychoish: There are two reasons that I can see to have a public interface that has a concept of logical plans: 
- logical plans mirror the way that the db/datafusion operates internally, 
- maintaining the implementation details of the existing implementation.
Neither of these are, on their own particularly compelling to me, though perhaps I'm missing something. I think the naming is confusing and setting up a weird expectation, so I've cleaned that up. 

As I understand it, the public interface, exposed by the bindings, supports the following operations:
- "execute this query immediately in all cases, and provide access to the results."
- "parse this query, and if it's DDL operations, execute the operation immediately, and provide access to the results."
- "parse the query and plan it's execution, and provide access to the results, but do not execute the query until the results are accessed".
Where "provide access to the results" is typically:
- some sort of stream/iterator/cursor interface to a sequence of arrow record batches
- an object that can be referenced and used as a table in future queries.

Is there another user story or way of accessing the results?

bindings/python/tests/test_logical_plan.py
```diff
@@ -9,9 +9,34 @@ def test_compose():
...
+        #
+        # In an attempt to match the previous behavior I've added the
+        # `resolve_table` method which _should_ give us a new table
+        # object (and therefore re-run the query). That's a problem
+        # for DDL/DML queries, but that's not a new issue.  The
+        # current implementation, if we run the test with these lines
+        # uncommented produces a "`pyo3_runtime.PanicException: Cannot
+        # start a runtime from within a runtime`" error, when the
+        # variable is accessed a second time. This is probably
+        # tractable, but would need more time.
```
tychoish, line 35: would love some feedback on this (particularly around user expectations. @talagluck @scsmithr

talagluck: I'm not following the differences currently. Let's discuss today.

bindings/python/src/execution.rs
```diff
@@ -0,0 +1,224 @@
...
+use pyo3::types::PyTuple;
+
+use crate::error::PyGlareDbError;
+use crate::runtime::wait_for_future;
+use crate::util::pyprint;
+
+#[pyclass]
+#[derive(Debug, Clone)]
+pub struct PyExecution {
+    op: Arc<Mutex<Operation>>,
```
scsmithr, line 30: I asked about the rationale for taking a `&mut self` for the Operation methods in the other pr. I think if we make a change to make those take `&self` instead, we can remove the Arc and Mutex here.

bindings/python/src/execution.rs
```diff
@@ -0,0 +1,224 @@
...
+
+        let batches = batches
+            .into_iter()
+            .map(|rb| rb.to_pyarrow(py))
+            .collect::<Result<Vec<_>, _>>()?;
+
+        Ok((schema.to_pyarrow(py)?, batches.to_object(py)))
+    }
+
+    fn resovle_operation(
```
scsmithr, line 129:
```suggestion
    fn resolve_operation(
```

bindings/python/src/execution.rs
tychoish: I removed the mutability and will rebase this in a bit. I still think that we need clone+copy here because of the python runtime constraints, but I'll play with it

bindings/python/src/connection.rs
```diff
@@ -1,50 +1,42 @@
...
-use futures::lock::Mutex;
 use once_cell::sync::OnceCell;
 use pyo3::prelude::*;
 use pyo3::types::PyType;
-use sqlexec::engine::{Engine, EngineStorage, SessionStorageConfig, TrackedSession};
-use sqlexec::{LogicalPlan, OperationInfo};
-
-use crate::execution_result::PyExecutionResult;
-
-pub(super) type PyTrackedSession = Arc<Mutex<TrackedSession>>;
```
tychoish, line 14: In a lot of ways, this `Arc<Mutex<>>` thing here serves the same purpose that the `Arc<Mutex<>>` around the `Operation` in the `PyExecution` object.

bindings/python/src/execution.rs
```diff
@@ -0,0 +1,224 @@
...
+    async fn scan(
+        &self,
+        _ctx: &SessionState,
+        projection: Option<&Vec<usize>>,
+        _filters: &[Expr],
+        _limit: Option<usize>,
+    ) -> Result<Arc<dyn ExecutionPlan>, DataFusionError> {
+        let op = self.op.lock().unwrap();
+        let schema = op.schema().expect("table must be resolved");
+        Ok(Arc::new(StreamingTableExec::try_new(
```
tychoish, line 193: using streaming table exec and the partition here is definitely a shortcut.

bindings/python/src/connection.rs
```diff
@@ -110,39 +101,11 @@ impl Connection {
     /// con = glaredb.connect()
     /// con.sql('create table my_table (a int)').execute()
     /// ```
-    pub fn sql(&mut self, py: Python<'_>, query: &str) -> PyResult<PyLogicalPlan> {
-        let cloned_sess = self.session.clone();
+    pub fn sql(&mut self, py: Python<'_>, query: &str) -> PyResult<PyExecution> {
```
universalmind303, line 104: does this need to take a mutable self `&mut self`? 

if we want to support:

> which has a few quirks in service of making it possible to reference python variables as tables in later SQL queries

wouldn't it make more sense to just clone `self.inner` on each call instead? It seems like a pretty cheap clone to make

bindings/python/src/execution.rs
```diff
@@ -0,0 +1,224 @@
...
+use pyo3::prelude::*;
+use pyo3::types::PyTuple;
+
+use crate::error::PyGlareDbError;
+use crate::runtime::wait_for_future;
+use crate::util::pyprint;
+
+#[pyclass]
+#[derive(Debug, Clone)]
+pub struct PyExecution {
```
universalmind303, line 29: what is a `PyExecution`? I don't understand the naming. Usually when creating binding wrapper structs, they are transparent

```rs
pub struct MyStruct {}

pub struct PyMyStruct {
  inner: MyStruct // or Arc<MyStruct>
}
```

```suggestion
pub struct PyOperation {
```

bindings/python/src/connection.rs
tychoish: done

bindings/python/src/execution.rs
tychoish: I'm not sure that operation is a good name either, though none of this naming is exposed to users.

universalmind303: in duckdb's python api, this is referred to as [DuckDBPyRelation](https://duckdb.org/docs/api/python/reference/#duckdb.DuckDBPyRelation). maybe we could just go with that? It seems to follow the same semantics


> Run a SQL query. If it is a SELECT statement, create a relation object from the given SQL query, otherwise run the query as-is.

tychoish: Relation makes about as much sense as LogicalPlan, which is to say, not much unless you know what's happening under the covers. I opted for `ExecutionOutput` in the latest version. 

Previously the bindings had this `PyExecutionResult` and I think having Result in that name is misleading, and also it's not a result as much as a promise... So I got `PyExecution` from there, which works-ish...

PR #2853, opened 2024-04-03
chore: df 36 upgrade

universalmind303:
Still a few items todo

- [x] get functions removed from `BuiltinScalarFunctions` added back in (Isnan, Encode, ...)
- [x] doublecheck `array_append` functionality


testdata/sqllogictests/datatypes.slt
```diff
@@ -27,12 +27,13 @@ select distinct arrow_typeof(f) from test
...
-
 # array functions don't work on fixed size lists
 statement error
 select array_element(1, f) from test;
 
-statement error 
+# but this one does work? 
+# TODO: should this be allowed?
+statement ok
 select array_append(f, 3) from test;
```
tychoish, line 37: what more should we do here before merging

testdata/sqllogictests/functions/hashing.slt
```diff
@@ -58,7 +58,7 @@ select siphash('42');
 query I
 select siphash([0, 100, 100, 300, 500, 800]);
 ----
-4101357204800564241
+4492675455749706323
```
tychoish, line 61: why would this change?

tychoish (commented):
just questions, otherwise seems good.

testdata/sqllogictests/datatypes.slt
universalmind303: I don't think mutating fixed sized arrays should be allowed. If it's a **fixed size**, then `array_append` doesn't seem like it should be valid. I'm not sure what we want to do with this one yet, maybe upstream something to datafusion? or monkeypatch the function to not work on fsl's?

universalmind303: I think we should just leave this for now. If it works in datafusion, i think it's fine for it to work in our system too. It just implicitly coerces the fsl to a regular list. 

I also think this a pretty rarely used function (especially on FSL's). I'll open up a separate issue to track this, but I don't think we need to act on it now.

universalmind303:
I want to wait to hear back about this [datafusion regression](https://github.com/apache/arrow-datafusion/issues/9951) before merging this.

PR #2857, opened 2024-04-05
feat: indirection for operations in rust SDK

tychoish:
This provides an extra layer of indirection in the rust sdk API and moves the operation into it's own type. We don't want to mirror this in the bindings exactly, but having this capability will let bindings re-execute "plans" without any funny business. 

I may want to make `DirectConnection`  and `StreamingConnection` which provide the same interface as connection but that pass through the operations.

Will rehome #2850 on top of this.

### DIFF
--- a/crates/glaredb/src/lib.rs
+++ b/crates/glaredb/src/lib.rs
@@ -138,68 +138,119 @@ impl ConnectOptions {
     }
 }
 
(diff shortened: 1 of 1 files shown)


tychoish:
merged this into #2832 independently.

PR #2871, opened 2024-04-08
feat: documentation for Rust SDK crate

tychoish:
Wrote docstrings for every public method/field. Hopefully, this will increase clarity for the SDK interface and also highlight some obvious places for improvement. 

I'm not wild about `Operation` ("results" would be better but that's too close to `Result`) other alternatives weren't compelling. I'm also not wild about the semantics/ergonomics of `execute().execute()` maybe the second one should be `.evaluate()` (to contrast with `.call()`.

Probably worth reviewing this before/during #2832.


crates/glaredb/src/lib.rs
```diff
@@ -1,3 +1,14 @@
+//! GlareDB is a database engine designed to provide the user
+//! experience and ergonomic embedded databases with the compute power
+//! and flexibility of large scale distributed serverless compute engines.
+//!
+//! The GlareDB Rust SDK is a set of high-level wrappers for a GlareDB
+//! instance as either a client or an embedded database. The
+//! implementation primarily underpins the implementations of the
+//! Python and Node.JS bindings, but may be used/useful directly for
```
greyscaled, line 8:
```suggestion
//! Python and Node.js bindings, but may be used/useful directly for
```

crates/glaredb/src/lib.rs
```diff
@@ -1,3 +1,14 @@
...
+//! testing GlareDB from within Rust tests, and even inside of rust
```
greyscaled, line 9:
```suggestion
//! testing GlareDB from within Rust tests, and even inside of Rust
```

crates/glaredb/src/lib.rs
```diff
@@ -18,38 +29,81 @@ use derive_builder::Builder;
...
-use sqlexec::environment::EnvironmentReader;
+pub use sqlexec::environment::EnvironmentReader;
 use sqlexec::errors::ExecError;
 use sqlexec::remote::client::RemoteClientType;
 use sqlexec::session::ExecutionResult;
 use sqlexec::OperationInfo;
 use url::Url;
 
-
+/// ConnectionOptions are the set of options to configure a GlareDB
```
greyscaled, line 39:
```suggestion
/// ConnectOptions are the set of options to configure a GlareDB
```

crates/glaredb/src/lib.rs
```diff
@@ -18,38 +29,81 @@ use derive_builder::Builder;
...
+/// instance, and are an analogue to the commandline arguments to
+/// produce a "running database". The ConnectOptionsBuilder provides a
+/// chainable interface to construct these values and produce as
```
greyscaled, line 42: either 'a' or maybe 'as a' ?

crates/glaredb/src/lib.rs
```diff
@@ -18,38 +29,81 @@ use derive_builder::Builder;
...
+    /// operations to local disk (sorts, large joins, etc.)
     #[builder(setter(into))]
     pub spill_path: Option<String>,
+    /// Defines the options used to configure the object store
+    /// (credentials, etc.)
     #[builder(setter(into), default = "HashMap::new()")]
     pub storage_options: HashMap<String, String>,
 
+    /// By default, the client will connect to the GlareDB service
+    /// using TLS. When this option is specified (and true), the this
```
greyscaled, line 80:
```suggestion
    /// using TLS. When this option is specified (and true), then this
```

crates/glaredb/src/lib.rs
```diff
@@ -18,38 +29,81 @@ use derive_builder::Builder;
...
+    /// GlareDB instance will establish an insecure connection. Use for
+    /// testing and development.
     #[builder(setter(strip_option))]
     pub disable_tls: Option<bool>,
+    /// Location of the GlareDB clout instance used by GlareDB
+    /// negotiate out-of-band certificate provisioning.
```
greyscaled, line 86:
```suggestion
    /// to negotiate out-of-band certificate provisioning.
```

crates/glaredb/src/lib.rs
```diff
@@ -18,38 +29,81 @@ use derive_builder::Builder;
...
     #[builder(default = "Some(\"https://console.glaredb.com\".to_string())")]
     #[builder(setter(into, strip_option))]
     pub cloud_addr: Option<String>,
+    /// Client type distinguishes what kind of remote client this is,
+    /// and is used for logging and introspection.
     #[builder(default = "Some(RemoteClientType::Cli)")]
     #[builder(setter(strip_option))]
     pub client_type: Option<RemoteClientType>,
+    /// Specify an optional environment reader, which GlareDB uses in
+    /// embedded cases so that queries can bindings to extract tables
```
greyscaled, line 96: This sentence is a bit off, not 100% sure what was intended, maybe something like:

```
can bind to extracted tables
from variables in scope of data frames or the
output of a query
```

crates/glaredb/src/lib.rs
```diff
@@ -18,38 +29,81 @@ use derive_builder::Builder;
...
+    /// from variables in the binding's scope that data frames, or the
+    /// output of a query.
     #[builder(setter(strip_option))]
     pub environment_reader: Option<Arc<Box<dyn EnvironmentReader>>>,
 }
 
 impl ConnectOptionsBuilder {
+    /// Adds a single option (key/value pair) to the builder for the
+    /// storage options map. All keys must be unique, and setting the
+    /// same otpion more than once.
```
greyscaled, line 106:
```suggestion
    /// storage options map. All keys must be unique, and setting the
    /// same option more than once.
```

greyscaled, line 106: What happens if setting more than once (sentence needs completion)?

crates/glaredb/src/lib.rs
```diff
@@ -229,10 +323,20 @@ impl Debug for Operation {
 }
 
 impl Operation {
+    /// Schema returns the schema of the results. This method returns
+    /// `None` before the query executes.
     pub fn schema(&self) -> Option<Arc<Schema>> {
         self.schema.clone()
     }
 
+    /// Executes the query, according the semantics of the operation's
```
greyscaled, line 332:
```suggestion
    /// Executes the query, according to the semantics of the operation's
```

crates/glaredb/src/lib.rs
```diff
@@ -25,31 +36,74 @@ use sqlexec::session::ExecutionResult;
 use sqlexec::OperationInfo;
 use url::Url;
 
-
+/// ConnectOptions are the set of options to configure a GlareDB
+/// instance, and are an analogue to the commandline arguments to
+/// produce a "running database". The ConnectOptionsBuilder provides a
+/// chainable interface to construct these values and produce as
```
greyscaled, line 42:
```suggestion
/// chainable interface to construct these values and produce a
```

greyscaled (approved):
Just those last two spots, but looks great and sending ✅

PR #2873, opened 2024-04-09
fix: in memory db should not be global

### LINKED ISSUE #2867: `test_execute_is_eager` in python bindings is incorrect
In this file https://github.com/GlareDB/glaredb/blob/main/bindings/python/tests/test_execution_model.py, there are a number of issues, 

- a `CREATE TABLE` operation (should) run more than once, but does not error on the second time. (`test_eager_ddl`)
- an `INSERT` operation should run more than once, but only seems to run once. (`test_eager_ddl`)
- a `CREATE TABLE` operation should produce an empty table, but may leak state from another test (`test_execute_is_eager`) 

This semantic was discovered while implementing #2850 and #2832, the former addresses (and fixes) the first two issues, but does not resolve the final issue.

tychoish:
This leads to confusing problems in the bindings.

Want to run this through CI, probably a couple of bindings tests will
break (and will be fixed.) Also need to update documentation of
bindings.

Closes #2867

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5761,7 +5761,6 @@ dependencies = [
  "logutil",
  "moka 0.12.5",
  "object_store",
- "once_cell",
  "tempfile",
  "thiserror",
  "tokio",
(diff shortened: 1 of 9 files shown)


scsmithr (approved):
Test failure is what we'd expect.

Issue #2828, opened 2024-03-26
bug: issue with read_json over http


talagluck:
## Description
I'm trying the following query:

```sql
select * From read_json('https://data.cityofnewyork.us/resource/2npr-yv2b.json?$limit=1000');
```

And I'm getting the following error:

```shell
ExecutionException: External error: Exec error: Remote session error: unable to dispatch table access: status: Unknown, message: "Missing session for id c3031c45-efcd-4491-b863-60f9570a88fe", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "content-length": "0", "date": "Tue, 26 Mar 2024 16:32:02 GMT"} }
```

When I go directly to the url, I can see the data.

And when I run this query, it works:

```sql
select * From read_json('https://opdb.org/api/search/typeahead?q=*');
```

tychoish:
Some debugging notes here: 
- the `$` in the URL seems to be potentially an issue. or at least that's a thing that's different between the thing that works and the thing that doesn't (this is unlikely to be the thing)
- the session error seems like a "running in cloud" problem, so I ran it locally in the CLI off of (approximately) `main` and got the following response
  ```psql
    > select * From read_json('https://data.cityofnewyork.us/resource/2npr-yv2b.json?$limit=1000');
    Error: External error: Unexpected status code '400 Bad Request' for url: 'https://data.cityofnewyork.us/resource/2npr-yv2b.json?$limit=1000'
   ```
   I don't think this is a cloud problem, or at least it's not a cloud problem worth looking at yet.
- I did some poking at this more and `curl` gives me a 400 when I attempt to do a `HEAD` request, so that's probably worth looking into.
- The error reporting from remote glaredb instances probably could be better. 

We definitely should investigate more here if there's something easy to get these to work (like `HEAD` request avoidance) but some web servers/CDNs might not be compatible for a bit.

talagluck:
Totally fair - and to be clear, this isn't necessarily a "let's fix now" but an "I see something that doesn't work and it would be nice if it did."

PR #2874, opened 2024-04-10
fix: Ignore HTTP store client errors trying to get content length

vrongmeal:
The `HEAD` method might not be supported. Ignore client side errors and assume we have `0` content length.

Fixes: #2828


crates/datasources/src/object_store/http.rs
```diff
@@ -326,10 +334,19 @@ impl ObjectStore for SimpleHttpStore {
             if content_length != 0 {
                 self.obj_store.get_opts(location, options).await
             } else {
-                self.simple_get_req(location.clone()).await.map_err(|e| {
+                self.simple_get_req(location.clone()).await.map_err(|err| {
+                    let err = if self.url.as_str().contains('*') {
+                        ObjectStoreSourceError::String(format!(
+                            "{err}\n\
+                            Note that globbing is not supported for HTTP.",
```
tychoish, line 341: I would avoid newlines in errors.

testdata/sqllogictests/functions/json_scan.slt
```diff
@@ -40,11 +40,6 @@ select * from ndjson_scan([]);
 
 # Glob patterns not supported on HTTP
 
-statement error Unexpected status code '404 Not Found'
-select * from ndjson_scan(
-  'https://raw.githubusercontent.com/GlareDB/glaredb/main/testdata/sqllogictests_datasources_common/data/*.ndjson'
```
tychoish, line 45: why did this error before and doesn't any more?

tychoish (commented):
nits

crates/datasources/src/object_store/http.rs
vrongmeal: The error is usually very long. Moreover, we just want to append a "note" here for the user (easier to read).

testdata/sqllogictests/functions/json_scan.slt
vrongmeal: So the previous error had both: status code and glob. Now the error message has changed...

crates/datasources/src/object_store/http.rs
tychoish: I mean, I think it's fine.

If it's very long it's going to wrap anyway and adding our own new line character isn't going to actually be a great place in all cases.

My concern is that it's going to end up in log messages with a new line that's going to be hard to read and maybe hard to parse. 

I'd also change it to put the summary at the beginning: `HTTP sources do not support globbing: {err}`

vrongmeal:
A test fails with a less precise error, should we just let it be for now. There isn't a better work around to improve the error message...

talagluck:
It looks like we're removing tests here - are there any tests we could add?

vrongmeal:
> It looks like we're removing tests here - are there any tests we could add?

So, to make most of the things work, we had to do a bit of hacky stuff which messes up the error messages. So the tests removed are simply not checking some particular error messages (related to HTTP). We still have the same tests below checking another error message so our coverage is still the same.

tychoish (approved):

PR #2875, opened 2024-04-10
fix: inserts clean up on failure

universalmind303:
This fixes a bug introduced by datafusion 36

previously some operations that failed during optimization such as invalid casts now fail at runtime. 

Since they now fail at runtime, it means that we would still create the catalog and table and only fail during insert afterwards. This left both the catalog and the storage in a bad state that didn't accurately reflect the operation. e.g.

```sql
create table invalid_ctas as (select cast('test' as int) as 'bad_cast');
```

This updated the catalog and created a table for `invalid_ctas`, but when you'd query it you would get an error. 

This PR makes sure that the operation is successful before committing the changes. It does so by exposing some new methods on the catalog client. `commit_state` `mutate_and_commit` and `mutate` instead of the previous `mutate`. 

The existing code was refactored to use the `mutate_and_commit` which is the same as the old `mutate`. The code that requires the commit semantics (create table) now uses `mutate` to first get an uncommitted catalog state with those changes, then does all of it's other actions 

- create the "native" table
- _Optional_ inserts into the table
- commits the catalog state

If any of the operations before the commit fail, then the catalog mutations are never committed.


crates/catalog/src/mutator.rs
```diff
@@ -30,12 +30,52 @@ impl CatalogMutator {
...
+                // Go ahead and refetch the catalog and retry the mutation.
+                //
+                // Note that this relies on metastore _always_ being stricter
+                // when validating mutations. What this means is that retrying
+                // here should be semantically equivalent to manually refreshing
+                // the catalog and rerunning and replanning the query.
+                debug!(error_message = msg, "retrying mutations");
+
+                client.refresh_cached_state().await?;
+                let state = client.get_cached_state().await?;
```
tychoish, line 60: does the `refresh_cached_state` call also apply whatever updates we're trying to commit here? It definitely reads as if we reset the state and then try and commit the state to what it already is. Either this is a bug or one of these methods could be more clearly named.

crates/catalog/src/mutator.rs
```diff
@@ -30,12 +30,52 @@ impl CatalogMutator {
         self.client.as_ref()
     }
 
+    /// Commit the catalog state.
+    /// This persists the state to the metastore.
+    pub async fn commit_state(
+        &self,
+        catalog_version: u64,
```
tychoish, line 37: does this expect the version to be the current version of the catalog upon which we want to apply the edits, or the version of the catalog we're trying update? I think the ambiguity/possibility for off-by-one errors in this case makes things confusing.

This is made further confusing because there's a version number in the catalog state, and I'm not sure how they should be compared, validated, or reconciled.

crates/catalog/src/mutator.rs
```diff
@@ -30,12 +30,52 @@ impl CatalogMutator {
...
+            Err(CatalogError {
+                msg,
+                strategy: Some(ResolveErrorStrategy::FetchCatalogAndRetry),
+            }) => {
+                // Go ahead and refetch the catalog and retry the mutation.
+                //
+                // Note that this relies on metastore _always_ being stricter
+                // when validating mutations. What this means is that retrying
+                // here should be semantically equivalent to manually refreshing
+                // the catalog and rerunning and replanning the query.
```
tychoish, line 56: but only once? if this fails do we bail out and retry at a higher level anyway? Why not implement this logic in the commit_state method rather than here? is there a limit to the depth of retries that we should allow, or does this just wait to hit the stack or recursion limit?

The debug statement should also, probably, include slightly more information: which deployment/database-id/catalog as well as the call depth/number of retries, or version number, if possible.

crates/metastore/src/client.rs
```diff
@@ -323,6 +323,28 @@ impl StatefulWorker {
...
+                });
+
+                let result = self.client.commit_catalog(commit_request).await.unwrap();
+                let result = result.into_inner();
+                let state: CatalogState = result.catalog.unwrap().try_into().unwrap();
+                self.set_cached_state(state);
+                let res = Ok(self.cached_state.clone());
+
+                if response.send(res).is_err() {
+                    error!("failed to respond to commit");
```
tychoish, line 345: should we report more context here (about which DB/catalog failed, or the error itself?

crates/metastore/src/database.rs
```diff
@@ -358,13 +342,34 @@ struct State {
...
+        State::from_persisted(persisted)
+    }
     /// Create a new state from a persisted catalog.
     ///
     /// The state will be combined with a predefinend builtin catalog objects.
     ///
     /// This will build the schema names and objects maps.
     fn from_persisted(persisted: PersistedCatalog) -> Result<State> {
         let mut state = persisted.state;
+        let counter = persisted.extra.oid_counter;
```
tychoish, line 372: this change seems to be a noop, right?

crates/metastore/src/database.rs
```diff
@@ -1419,6 +1423,25 @@ impl BuiltinCatalog {
...
 
+#[cfg(test)]
+impl DatabaseCatalog {
+    /// Try to mutate the catalog and immediately commit the changes.
+    ///
+    /// Errors if the provided version doesn't match the version of the current
+    /// catalog.
+    ///
+    /// On success, a full copy of the updated catalog state will be returned.
+    // TODO: All or none.
```
tychoish, line 1434: I think this comment is stale, also not sure why this is test-only.

tychoish (commented):
are there existing tests concurrent writes to the catalogs/conflict handling?

crates/catalog/src/mutator.rs
universalmind303: Honestly, i don't have a super good understanding of this code, it was just copied from the original function.

crates/metastore/src/client.rs
universalmind303: no, this is in par with the other error messages

```rs
error!("failed to respond to ping");

error!("failed to send cached state");
```

crates/metastore/src/database.rs
universalmind303: The reason this isnt exposed elsewhere is that the actions need to go over grpc and instead of creating 'mutate', 'mutate and commit', and 'commit' actions, all we really need on the GRPC service is 'mutate' and 'commit'. So the interfaces on the other side of things use a 'mutate_and_commit' that makes both grpc calls and only serves as a utility function. 


the same applies here, but it is on the other side of grpc, so it's not actually "used" anywhere else beyond testing. 

There were a lot of instances where this was used in testing, so it made sense to not manually rewrite all of the tests.

crates/catalog/src/mutator.rs
universalmind303: yeah i found this a bit confusing too. I don't think this is directly related to this PR functionality though as this was already prevalent in the existing code. 

I do think we can clean up these interfaces a bit, but I think it'd detract from the core logic needed to fix this bug. Maybe we can open up an issue to follow up on this?

crates/metastore/src/database.rs
tychoish: it'd make sense to want to have the combined operation _behind_ the gRPC interface so that you wouldn't need to have an extra round trip in-between the operations (thereby reducing the risk of conflicts), but I am relatively unfamiliar with the code.

crates/catalog/src/mutator.rs
scsmithr: In the original `mutate` method, this retry exists mostly for cloud + HA metastores. A glaredb instance might have an out-of-date catalog if some other instance previously made a change and we haven't updated our cached state yet. The comment is just saying that we retry the mutation against the now up-to-date version, and would (should) semantically be the same as the user running the query again.

This is only retried once since catalog changes are infrequent, and the likelihood that a separate change is made between when we get the updated state back and when we try to apply the mutations again is low.

crates/catalog/src/mutator.rs
scsmithr: This is an important question to answer. `refresh_cached_state` is responsible for refreshing the cached catalog state on this node by getting the latest version from metastore. `get_cached_state` just gets a reference to it.

Can metastore return uncommitted states? (directly impacting what `refresh_cached_state` will do)

crates/catalog/src/mutator.rs
scsmithr: > yeah i found this a bit confusing too. I don't think this is directly related to this PR functionality though as this was already prevalent in the existing code.

This is directly related to this pr. What should this version be when committing a catalog state? Does it match `state.version`?

crates/sqlexec/src/planner/physical_plan/create_table.rs
```diff
@@ -192,16 +192,30 @@ impl CreateTableExec {
...
+            })?;
+            return Err(e);
+        } else {
+            mutator
+                .commit_state(catalog_version, state.as_ref().clone())
+                .await
+                .map_err(|e| {
+                    DataFusionError::Execution(format!("failed to commit catalog state: {e}"))
+                })?;
+        }
```
scsmithr, line 218:
```suggestion
        if let Err(e) = insert_res {
            storage.delete_table(ent).await.map_err(|e| {
                DataFusionError::Execution(format!("failed to clean up table: {e}"))
            })?;
            return Err(e);
        }
        
        mutator
            .commit_state(catalog_version, state.as_ref().clone())
            .await
            .map_err(|e| {
                DataFusionError::Execution(format!("failed to commit catalog state: {e}"))
            })?;
```

Clearer to follow.

crates/catalog/src/mutator.rs
universalmind303: added some comments, and cleaned this up

crates/catalog/src/mutator.rs
universalmind303: added some comments and cleaned this up to be clearer.

crates/catalog/src/mutator.rs
tychoish: I actually think that it's retried infinitely because of the recursive call.

universalmind303: > I actually think that it's retried infinitely because of the recursive call.

it's not calling `self.commit_state` but instead `client.commit_state` so this isn't recursive

scsmithr:
Some tests for `catalog/src/client.rs`

```rust
    #[tokio::test]
    async fn mutate_no_commit() {
        // Test to ensure uncommitted states aren't accessible.

        let client = new_local_metastore().await;
        let supervisor = MetastoreClientSupervisor::new(client, DEFAULT_METASTORE_CLIENT_CONFIG);

        let db_id = Uuid::nil();
        let c1 = supervisor.init_client(db_id).await.unwrap();
        let c2 = supervisor.init_client(db_id).await.unwrap();

        // Both clients have the same state.
        let s1 = c1.get_cached_state().await.unwrap();
        let s2 = c2.get_cached_state().await.unwrap();
        assert_eq!(s1.version, s2.version);

        // Client 1 mutates, no commit.
        let _ = c1
            .try_mutate(
                s1.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "wario".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();


        // Client 2 should still get latest committed version.
        c2.refresh_cached_state().await.unwrap();
        let refreshed2 = c2.get_cached_state().await.unwrap();
        assert_eq!(s2.version, refreshed2.version);

        // Client 1 should also get the latest committed version, not the
        // uncommitted version.
        c1.refresh_cached_state().await.unwrap();
        let refreshed1 = c1.get_cached_state().await.unwrap();
        assert_eq!(s1.version, refreshed1.version);
    }

    #[tokio::test]
    async fn double_mutate() {
        // Test to ensure a client can mutate twice without having the
        // intermediate mutate affect anything.
        //
        // This represents the "rollback" case for when a query fails and
        // doesn't commit.

        let client = new_local_metastore().await;
        let supervisor = MetastoreClientSupervisor::new(client, DEFAULT_METASTORE_CLIENT_CONFIG);

        let db_id = Uuid::nil();
        let c = supervisor.init_client(db_id).await.unwrap();
        let state = c.get_cached_state().await.unwrap();

        // First mutation. This one doesn't get committed.
        let _ = c
            .try_mutate(
                state.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "wario".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();

        // Second mutation, this one gets committed.
        let new_state = c
            .try_mutate(
                state.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "mario".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();

        let _ = c
            .commit_state(state.version, new_state.as_ref().clone())
            .await
            .unwrap();


        // Client should now have the new committed state.
        c.refresh_cached_state().await.unwrap();
        let refreshed = c.get_cached_state().await.unwrap();
        assert_eq!(new_state.version, refreshed.version);

        // New state version should not be the same as the original version.
        assert_ne!(state.version, refreshed.version);
    }

    #[tokio::test]
    async fn commit_arbitrary_state_version() {
        // Test trying to commit to an arbitrary version.

        let client = new_local_metastore().await;
        let supervisor = MetastoreClientSupervisor::new(client, DEFAULT_METASTORE_CLIENT_CONFIG);

        let db_id = Uuid::nil();
        let c = supervisor.init_client(db_id).await.unwrap();
        let state = c.get_cached_state().await.unwrap();

        // Mutate...
        let new_state = c
            .try_mutate(
                state.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "wario".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();

        // Then commit with arbitrary version.
        let _ = c
            .commit_state(9999, new_state.as_ref().clone())
            .await
            .unwrap_err();
    }

    #[tokio::test]
    async fn concurrent_commits() {
        // Test to ensure two clients committing at the same errors as
        // appropriate.

        let client = new_local_metastore().await;
        let supervisor = MetastoreClientSupervisor::new(client, DEFAULT_METASTORE_CLIENT_CONFIG);

        let db_id = Uuid::nil();
        let c1 = supervisor.init_client(db_id).await.unwrap();
        let c2 = supervisor.init_client(db_id).await.unwrap();

        // Both clients have the same state.
        let s1 = c1.get_cached_state().await.unwrap();
        let s2 = c2.get_cached_state().await.unwrap();
        assert_eq!(s1.version, s2.version);

        // Client 1 mutation...
        let new_state1 = c1
            .try_mutate(
                s1.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "wario1".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();

        // Client 2 mutation...
        let new_state2 = c2
            .try_mutate(
                s2.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "wario2".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();

        // Client 1 commit should succeed.
        let _ = c1
            .commit_state(s1.version, new_state1.as_ref().clone())
            .await
            .unwrap();

        // Client 2 commit should error.
        let _ = c2
            .commit_state(s2.version, new_state2.as_ref().clone())
            .await
            .unwrap_err();
    }

    #[tokio::test]
    #[ignore]
    async fn error_on_commit_to_committed_catalog() {
        // Test to ensure we error when trying to commit an already committed
        // catalog.
        //
        // Once committed, a catalog version shouldn't be committed to again.
        // Other wise we could end up in a scenario where two nodes commit the
        // same version, with one overwriting the other's version.

        let client = new_local_metastore().await;
        let supervisor = MetastoreClientSupervisor::new(client, DEFAULT_METASTORE_CLIENT_CONFIG);

        let db_id = Uuid::nil();
        let c = supervisor.init_client(db_id).await.unwrap();
        let state = c.get_cached_state().await.unwrap();

        // Mutate...
        let new_state = c
            .try_mutate(
                state.version,
                vec![Mutation::CreateSchema(CreateSchema {
                    name: "wario1".to_string(),
                    if_not_exists: false,
                })],
            )
            .await
            .unwrap();

        // Commit as normal.
        let _ = c
            .commit_state(state.version, new_state.as_ref().clone())
            .await
            .unwrap();

        // Try to commit again.
        c.refresh_cached_state().await.unwrap();
        let refreshed = c.get_cached_state().await.unwrap();
        let _ = c
            .commit_state(refreshed.version, new_state.as_ref().clone())
            .await
            .unwrap_err();
    }
```

The ignored one is a test I'd expect to pass.

universalmind303:
@scsmithr I added those tests & added some checks to get `error_on_commit_to_committed_catalog` to pass.

crates/catalog/src/mutator.rs
```diff
@@ -74,6 +109,22 @@ impl CatalogMutator {
...
+    /// This will retry mutations if we were working with an out of date
+    /// catalog.
+    pub async fn mutate_and_commit(
+        &self,
+        catalog_version: u64,
+        mutations: impl IntoIterator<Item = Mutation>,
+    ) -> Result<Arc<CatalogState>> {
+        let state = self.mutate(catalog_version, mutations).await?;
+        self.commit_state(catalog_version, state.as_ref().clone())
+            .await
```
tychoish, line 126: though it's not _enough_ (I think), if there's going to be a version conflict, we could know/fix it here.

crates/catalog/src/mutator.rs
```diff
@@ -30,12 +30,47 @@ impl CatalogMutator {
...
+        {
+            Err(CatalogError {
+                msg,
+                strategy: Some(ResolveErrorStrategy::FetchCatalogAndRetry),
+            }) => {
+                debug!(error_message = msg, "retrying commit");
+                // refresh the cached state and retry the commit
+                client.refresh_cached_state().await?;
+
+                client.commit_state(state.version, state).await
```
tychoish, line 60: I think I don't understand how this retry helps.

If we get here it's probably because the _service_ failed to commit, because the state The version in `state` hasn't changed since the last attempt, the version in `state`is likely to be _more_ wrong now. 

I think the desired flow for this is: 

- hand the client the state that you want to commit and the version of the catalog that you modified to get the new state. 
- if we get an error that should be retried, discard the state, reapply the mutation to the _new_ state of the catalog fetched after the last failure, and attempt to commit this state. 

Right now, based on my reading of the code, I think: 
- we get a failure from the service, presumably because the new state of the catalog is higher than the state that we're trying to apply
- we apply the same state, and the client rejects before sending it to the service, it because it's stale (this is correct, but it means that the error will either:
  - always bubble up to the user.
  - recursively retry forever.

crates/metastore/src/client.rs
```diff
@@ -323,6 +323,36 @@ impl StatefulWorker {
...
+
+                let commit_request = tonic::Request::new(CommitRequest {
+                    db_id: self.db_id.into_bytes().to_vec(),
+                    catalog_version: current_catalog_version,
+                    catalog: Some(state.try_into().unwrap()),
+                });
+
+                let result = self.client.commit_catalog(commit_request).await.unwrap();
+                let result = result.into_inner();
+                let state: CatalogState = result.catalog.unwrap().try_into().unwrap();
```
tychoish, line 348: are these unwraps safe?

crates/catalog/src/mutator.rs
universalmind303: no recursion takes place here. It only tries to recommit it once. We can pretty easily remove this, it was added as the `mutation` path had this same retry logic and didn't want to deviate too much from the original code.

crates/metastore/src/client.rs
universalmind303: If these ever do panic, they won't crash the application as they're over the grpc client which has some built in panic handling.

universalmind303: our other endpoints are also unwrapping `client.ping().await.unwrap();`

crates/catalog/src/mutator.rs
universalmind303: i don't understand what you are saying/asking.

tychoish: if `state.version != (catalog_version -1)` then that _might_ (probably?) would be known here and we wouldn't need to run `commit_state()` (because we know it'd error)... similarly, if the mutation applied cleanly, (e.g. 125 does not error), and the commit fails, then it seems like retrying here would be safe.)

universalmind303: We already have that check further down `metastore/src/client.rs:332`. 

we _could_ add that check here as well to avoid the retries, but it seems like an optimization.

tychoish: the check in the metastore client (and the retry) doesn't (I  think?) reapply the mutation to the fresh state, so the retry, so while this code, in both cases is _correct_, I'm not sure there's a case where the retry in the metastore client could ever succeed, unless there was some kind of very transient networking problem, and a retry here has some chance of succeeding.

universalmind303: removed the retry

tychoish:
modulo some questions about this retry logic, I think @scsmithr should take another pass, but I think this feels correct. 

[also just saying this here, if the insert is interrupted (process restart), there's still an inconsistency because the data is partially written and the catalog hasn't committed... similarly if the insert succeeds but the catalog update fails, the data is written but the table doesn't exist. I don't think that's a blocking bug right now.]

scsmithr (approved):

PR #2879, opened 2024-04-11
chore: hoist node.js bindings on top of rust sdk

tychoish:
The commit history is wacky because I started this work before the SDK branch was fully merged *but* the diff is normal.

@greyscaled if you wanna take this for a spin, that'd be cool too.


bindings/nodejs/src/connection.rs
```diff
@@ -241,21 +116,11 @@ impl Connection {
...
     #[napi(catch_unwind)]
     pub async fn execute(&self, query: String) -> napi::Result<()> {
-        let sess = self.session.clone();
-        let mut sess = sess.lock().await;
-
-        let plan = sess
-            .create_logical_plan(&query)
+        self.inner
+            .execute(query)
+            .execute()
```
universalmind303, line 121: this seems wrong? `execute().execute()`

bindings/nodejs/src/connection.rs
```diff
@@ -241,21 +116,11 @@ impl Connection {
     /// ```
     #[napi(catch_unwind)]
     pub async fn execute(&self, query: String) -> napi::Result<()> {
-        let sess = self.session.clone();
-        let mut sess = sess.lock().await;
-
-        let plan = sess
-            .create_logical_plan(&query)
+        self.inner
+            .execute(query)
```
universalmind303, line 120: `.execute.execute` seems like it shouldn't be valid.

bindings/nodejs/src/connection.rs
```diff
@@ -1,131 +1,41 @@
...
-            .new_local_session_context(SessionVars::default(), SessionStorageConfig::default())
-            .await
-            .map_err(JsGlareDbError::from)?;
-        let con = Connection {
-            session: Arc::new(Mutex::new(sess)),
-            _engine: Arc::new(engine),
-        };
-
-        Ok(con.clone())
+        static DEFAULT_CON: OnceCell<Connection> = OnceCell::new();
```
universalmind303, line 24: we should probably just return a new instance each time. This seems like it could produce some unexpected behavior. 

such as. 

```js
const db1 = glaredb.connect() 

const db2 = glaredb.connect() 
```

I'd expect these to be two separate sessions, but this code looks like they'd use a shared connection.



JS isn't really known for it's performance, so the cost of spinning up a new session is likely not a concern. It's what we are already doing, and seems a bit safer a

bindings/nodejs/src/connection.rs
```diff
@@ -214,17 +96,10 @@ impl Connection {
...
-            .await
-            .map_err(JsGlareDbError::from)?;
-
-        let op = OperationInfo::new().with_query_text(query);
-
-        Ok(JsLogicalPlan::new(plan, cloned_sess, op))
+    pub async fn prql(&self, query: String) -> napi::Result<JsExecution> {
+        let mut op = self.inner.prql(query);
+        op.execute().await.map_err(JsGlareDbError::from)?;
+        Ok(op.into())
```
universalmind303, line 102: this doesn't make any sense. 

- :100 we build an operation
- :101 we execute the operation
- :102 we return the operation converted into `JsExecution`? 

the first two make sense, but I'd think we'd return the result of `execute` instead of mutating op, then returning that as something else. 

i would expect a flow that more closely resembles this

```rs
        let mut op = self.inner.prql(query);
        let result = op.execute().await.map_err(JsGlareDbError::from)?;
        Ok(result.into())

File: bindings/nodejs/src/execution.rs [hunk 5]
```diff
@@ -0,0 +1,116 @@
...
+use arrow_util::pretty;
+use datafusion::arrow::ipc::writer::FileWriter;
+use futures::stream::StreamExt;
+use glaredb::{RecordBatch, SendableRecordBatchStream};
+
+use crate::error::JsGlareDbError;
+
+#[napi]
+#[derive(Clone, Debug)]
+pub struct JsExecution {
```
universalmind303, line 12: what is a `JsExecution`?

bindings/nodejs/src/execution.rs
```diff
@@ -0,0 +1,116 @@
...
+    op: Arc<Mutex<glaredb::Operation>>,
```
universalmind303, line 13: why does this need to be `Arc<Mutex<>>` why not own `Operation` directly?

bindings/nodejs/src/connection.rs
```diff
@@ -166,38 +76,10 @@ impl Connection {
...
-
-                Ok(JsLogicalPlan::new(
-                    LogicalPlan::Noop,
-                    cloned_sess,
-                    Default::default(),
-                ))
-            }
-            _ => Ok(JsLogicalPlan::new(plan, cloned_sess, op)),
-        }
+    pub async fn sql(&self, query: String) -> napi::Result<JsExecution> {
```
universalmind303, line 79: same comment as `fn prql`

bindings/nodejs/src/connection.rs
tychoish: yeah, known issue, this is a rust sdk problem that I talked about with sean earlier. The API isn't public (yet) an I'm trying to figure out the right words for these things.

bindings/nodejs/src/connection.rs
tychoish: > I'd expect these to be two separate sessions, but this code looks like they'd use a shared connection.

This is complicated: 
- before https://github.com/GlareDB/glaredb/commit/138347b287c94d53cbe37ee3f90b9284a08c42ae, there was only one glaredb in memory database (particularly confusing for testing, I think, but also users.) 
- as of #2850, the python driver has (will have) the semantics you (and I) expect, as above.
- also as of that PR the `glaredb.sql()` and `glaredb.execute()` functions in python, will access this static/global shared db.
- I believe, that we could have a similar (global) functionality in node, but I'd be happy to drop it. my grep shows that this is never called.

tychoish:
So this patch mirrors the Python PR, which has a few quirks in service of making it possible to reference python variables as tables in later SQL queries (which means that the result of the `sql()` and `prql()` methods need to be re-executable.) 

Most of the quirks of this implementation are in service of mirroring the python implementation, but I think if we don't intend to have this API, then we can have a much simpler implementation here.

universalmind303 (approved):

PR #2902, opened 2024-04-16
feat: read_blob

### LINKED ISSUE #2683: `read_blob` function
## Description

duckdb style `read_blob` function

https://duckdb.org/docs/guides/import/read_file#read_blob

for some additional context, I think you could do some interesting semantic search given this kind of functionality.

An example of using `read_blob` and semantic search to build a mechanism for searching for files using semantic text

```sql
> create table doc_embeddings as (
  SELECT *, openai_embed(content::text) as embedding
  FROM read_blob('**/*.rs')
);

> select cosine_similarity(openai_embed('configuration object for various supported storage types'), embeddings), content;
```

universalmind303:
closes #2683


tychoish (approved):
I want someone else to also look at this, but this seems pretty straight forward.

crates/sqlbuiltins/src/functions/table/read_blob.rs
```diff
@@ -0,0 +1,361 @@
...
+                        }
+                    }
+                }
+                None => None,
+            };
+
+            for field in schema.fields() {
+                // we need to do this hacky option take because of the borrow checker.
+                // these matches should only ever match once, but the borrow checker doesn't know that.
+                // so we just wrap the value in an option to ensure it's only taken once.
```
vrongmeal, line 193: Instead of doing this, can we simply create the columns at the end? This code feels a bit hard to read. We could just have all the variables and then create different columns (and eventually record batch) from them at the end.

vrongmeal (commented):
Looks good! Just a suggestion about removing the hack.

crates/sqlbuiltins/src/functions/table/read_blob.rs
universalmind303: @vrongmeal I refactored this to instead build the columns with an index, then reorder them by the index. I think this is a lot more readable.

scsmithr (approved):

PR #2903, opened 2024-04-17
feat(sdk): row-based access to results in sdk

tychoish:
the bindings have similar functionality via roughly-native mechanisms
(pyarrow, etc), but if this is going to be useful for interacting with
results in tests, something like this is going going to be needed.


universalmind303 (changes requested):
Unless there is a very compelling usecase for this, I'd be very hesitant about exposing this functionality. This is an anti-pattern when working with columnar data. 

At a minimum, I'd want to see a docstring that explicitly states the performance implications of row based iteration. Something like: 

```md
Row iteration is not optimal as the underlying data is stored in columnar 

Note that this method should not be used in place of native (columnar) operations, due the high cost of materializing all data into a Map

it should be used only when you need to move the values out of columnar format or into other object that cannot operate directly with Arrow/Datafusion columns.
```

tychoish:
We could feature flag this for testing, but this isn't (and for the reasons you point out) shouldn't be used for production code, but we're not in this PR, and I'm definitely willing to use a different scalar value wrapper, or another conversion method if you have a better one in mind, but the performance can be improved...

Anyway, this is not radically different from a design perspective from the [to_pydict](https://github.com/apache/arrow/blob/be5f5b8f8f8294d43681c59e1e67f7434a10861c/python/pyarrow/table.pxi#L2135) method that we use in the tests for the bindings.

crates/glaredb/src/lib.rs
```diff
@@ -312,6 +313,76 @@ impl From<Result<SendableRecordBatchStream, ExecError>> for RecordStream {
...
+    fn try_from(batch: RecordBatch) -> Result<Self, Self::Error> {
+        let schema = batch.schema();
+        let mut out = Vec::with_capacity(batch.num_rows());
+
+        for row in 0..batch.num_rows() {
+            let mut record = RowMap::with_capacity(batch.num_columns());
+            for (idx, field) in schema.fields.into_iter().enumerate() {
+                record.insert(
+                    field.name().to_owned(),
+                    ScalarValue::try_from_array(batch.column(idx), row)?,
```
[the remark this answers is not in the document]
scsmithr: This seems fine to me. The bounds check is pretty much a constant assert, and the downcast needs to happen _somewhere_, and here is as good as any.

We use `try_from_array` for the pg stuff as well: https://github.com/GlareDB/glaredb/blob/main/crates/pgsrv/src/codec/server.rs#L327-L351 (wrapped, but internally this calls the same function)

crates/glaredb/src/lib.rs
```diff
@@ -312,6 +313,76 @@ impl From<Result<SendableRecordBatchStream, ExecError>> for RecordStream {
     }
 }
 
+/// RowMap represents a single record in an ordered map.
+type RowMap = indexmap::IndexMap<String, ScalarValue>;
+
+/// RowMapBatch is equivalent to a row-based view of a record batch.
+pub struct RowMapBatch(Vec<RowMap>);
```
scsmithr, line 320: I'd add a quick comment this should be used sparingly/for tests.

Also, we could use an `Arc<str>` instead of `String` for the key if we want to keep the size down.

scsmithr (approved):

crates/glaredb/src/lib.rs
tychoish: arc str! 

My other theory was to just use the bson library, but the bson value enum has fewer types than the scalar value type, so avoiding more downcasting than necessary seems good

PR #2904, opened 2024-04-17
refactor: remove datafusion dependency from 'parser' crate

universalmind303:
For some additional context. 

This entirely removes `datafusion` as a dependency from the `parser` crate. Since we're no longer bound to datafusion's sqlparser version, it opens us up to iterate faster on sql functionality by using the latest version of 'sqlparser' crate _(or even a fork)_


crates/parser/src/options.rs
```diff
@@ -1,10 +1,72 @@
...
+    ARROW,
+    /// Apache Avro file
+    AVRO,
+    /// Apache Parquet file
+    PARQUET,
+    /// CSV file
+    CSV,
+    /// JSON file
+    JSON,
+}
```
tychoish, line 19: this is (I think?) doesn't impact the similarly named enum in data fusion (because of the dependency between `FileType` and `FileFormat`... 

having a noe on this somewhere might be good, otherwise things become difficult to grep and make sense of things.

tychoish (approved):
just a nit about a comment, as long as the tests pass

vrongmeal (approved):

crates/parser/src/options.rs
```diff
@@ -1,10 +1,72 @@
...
+    type Err = ParserError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let s = s.to_uppercase();
+        match s.as_str() {
+            "ARROW" => Ok(FileType::ARROW),
+            "AVRO" => Ok(FileType::AVRO),
+            "PARQUET" => Ok(FileType::PARQUET),
+            "CSV" => Ok(FileType::CSV),
+            "JSON" | "NDJSON" => Ok(FileType::JSON),
```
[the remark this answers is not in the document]
universalmind303: No, this is a direct copy from datafusion's `FileType` for now. I needed to copy it over to break our dependency in `parser`. Will add a note that it's a direct equivalent to datafusion's `FileType`.

crates/parser/src/options.rs
universalmind303: This is a direct copy from datafusion's `FileType` for now. I needed to copy it over to break our dependency in `parser`. Will add a note that it's a direct equivalent to datafusion's `FileType`.

PR #2905, opened 2024-04-17
feat: support = operator in function args

universalmind303:
Now if functions support named args, we can use python style `=` assignment in the function instead of the lesser known `=>` assignment operator

ex: 

```sqj
# before
select * from read_csv('some_file', delimiter => ';');

# after
select * from read_csv('some_file', delimiter = ';');
```

Note: the `=>` operator still works.


vrongmeal (approved):

tychoish (approved):
lgtm modulo failing test

universalmind303:
so it turns out this breaks a lot of generated queries, such as the ones generated from prql. We'd likely need some way to configure this per function. Closing this for now as it's a lot more work than initially intended.

PR #2924, opened 2024-04-23
feat: read_text table function

universalmind303:


testdata/sqllogictests_object_store/gcs/read_text.slt
```diff
@@ -0,0 +1,15 @@
...
+PROVIDER gcp OPTIONS ( 
+  service_account_key '${GCP_SERVICE_ACCOUNT_KEY}' 
+);
+
+
+query T
+select filename from read_text('gs://${GCS_BUCKET_NAME}/multi_csv/**/*', gcp_creds);
+----
+multi_csv/bikeshare_stations.csv
+multi_csv/bikeshare_stations_2.csv
```
tychoish, line 12: is there a way to assert in an SLT that the value that we expect to have data/content isn't empty?

tychoish (approved):
I think this is fine as is but I have a few questions.

testdata/sqllogictests_object_store/gcs/read_text.slt
universalmind303: yeah we are testing that in `testdata/sqllogictests/functions/read_text.slt`

```sql
query I
select content from read_text('./testdata/blob/hello.txt');
----
hello world
```

PR #2932, opened 2024-04-24
refactor: use "arc<mutex<session>>" in repl

universalmind303:
very small PR to use `Arc<Mutex<>>` in the repl for `TrackedSession`. Just some prereq work for adding tab completions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1956,6 +1956,7 @@ dependencies = [
  "num_cpus",
  "object_store",
  "object_store_util",
+ "parking_lot",
  "parser",
  "pgrepr",
  "pgsrv",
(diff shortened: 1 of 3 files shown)


tychoish:
Not sure  I understand why this is needed, but not otherwise objectionable.,

universalmind303:
> Not sure I understand why this is needed, but not otherwise objectionable.

The completer needs a session to execute the queries for tables/schemas/etc. We can hold off on merging this until I have a bit of the other work done as well. This is mostly my attempt to keep the PR's smaller and more manageable for reviewing.

tychoish:
> The completer needs a session to execute the queries for tables/schemas/etc. We can hold off on merging this until I have a bit of the other work done as well. This is mostly my attempt to keep the PR's smaller and more manageable for reviewing.

Definitely appreciate it. Maybe let's let it sit and build a stack on top of it.

PR #2935, opened 2024-04-24
fix: insert metrics tracking

tychoish:
Saw this while working on #2931


crates/datasources/src/mongodb/insert.rs
```diff
@@ -135,6 +142,4 @@ impl ExecutionPlan for MongoDbInsertExecPlan {
     }
 
     fn metrics(&self) -> Option<MetricsSet> {
-        Some(self.metrics.clone_inner())
-    }
-}
```
greyscaled, line 140: @tychoish I think perhaps an extra brace got removed, based on the CI error?

PR #2950, opened 2024-05-03
feat: postgres implement column limit for postgres

tychoish:


tychoish:
This consistently fails because we can't consistently predict the schema of the results of the query based on the information we have.

PR #2959, opened 2024-05-07
fix: Don't preprocess database name when setting session catalog alias

scsmithr:
Previously we were doing some processing on the database name to try to extract out the "true" name from names that follow the "org/database" format. However that meant if someone connects to the db over postgres using a hostname like "proxy.glaredb.com/my_org/my_db", we would set the catalog alias to only "my_db", disallowing queries like `SELECT * FROM
"my_org/my_db"."my_schema"."my_table"` even though that's a valid query given the connect string.

This change makes the name returned from `SHOW database_name` consistent with the alias we set on the session catalog, allowing the previous query to work as expected.


talagluck:
I can't speak to the code here, but the functionality that you describe is as I expect.

PR #2980, opened 2024-05-13
fix: release should have cmake

tychoish:
hopefully it's in the path with this...


scsmithr (approved):

tychoish:
this doesn't fix the underlying issue with the rease build.

PR #3002, opened 2024-05-30
chore: reduce exported surface area from rust SDK

tychoish:


crates/glaredb/src/lib.rs
```diff
@@ -714,3 +724,76 @@ impl Operation {
...
+    #[error(transparent)]
+    Metastore(#[from] MetastoreError),
+    #[error(transparent)]
+    Exec(#[from] ExecError),
+    #[error(transparent)]
+    ConfigurationBuilder(#[from] ConnectOptionsBuilderError),
+
+    #[error("{0}")]
+    Other(String),
+}
```
scsmithr, line 784: This should be named something other than `Error` since `Error` is standard rust trait.

scsmithr (approved):

PR #3037, opened 2024-06-14
feat: COPY TO support for delta

tychoish:
This has some path handling bug possibly in the tests or in the way
the object storage is being used that causes it to fail, but the code
and integration is broadly speaking correct.


scsmithr (approved):
lgtm

PR #3043, opened 2024-06-20
fix: excel table function should read data more consistently from remote sources

tychoish:
We previously had different dispatching and validation rules for
`read_excel` and other tables. This should make the code paths more
similar for the table function and the external tables.

@talagluck maybe worth adding some more tests to this one


talagluck:
Tests pushed!

scsmithr (approved):

PR #3118, opened 2024-07-26
feat: excel table support for excel data sources

### LINKED ISSUE #2949: Hi,
Hi, 
will you consider supporting [Excel tables](https://support.microsoft.com/en-us/office/overview-of-excel-tables-7ab0bb7d-3a9e-4b56-a3c9-6c94334e492c) ? I think more and more data are stored in this structure, and they have many advantages over ranges (being relatively close to a database table, with column title, no merged columns). Tables are named, and AFAIK, their name is unique within a workbook (see for instance, in Apache POI : https://poi.apache.org/apidocs/dev/org/apache/poi/ss/usermodel/Table.html, ExcelJS: https://github.com/exceljs/exceljs?tab=readme-ov-file#tables, .NET importExcel :https://www.powershellgallery.com/packages/ImportExcel/6.5.0/Content/GetExcelTable.ps1 or https://learn.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.table?view=openxml-3.0.1 for giving a few example I came across). I don't know about any Rust implementation though (does Calamine support it?). It's frustrating having data in Excel tables, or data loaded into Excel through Power Query which creates Excel tables and not being able to load them by name.

Expected behaviour would :
```
select * from read_excel('path/to/file.xlxs', table => 'table1') -- Uses the specified table
```
Hope it's doable without too much effort. Thanks!

_Originally posted by @jgranduel in https://github.com/GlareDB/glaredb/issues/1994#issuecomment-2088797993_

tychoish:
Closes #2949

@talagluck can you write the tests for this (e.g. need a worksheet
with a table and corresponding tests.)

### DIFF
--- a/crates/datasources/src/excel/mod.rs
+++ b/crates/datasources/src/excel/mod.rs
@@ -34,20 +34,36 @@ impl ExcelTable {
         store_access: Arc<dyn ObjStoreAccess>,
         source_url: DatasourceUrl,
         sheet_name: Option<String>,
+        table_name: Option<String>,
         has_header: bool,
     ) -> Result<ExcelTable, ExcelError> {
         match source_url {
             DatasourceUrl::File(path) => {
                 let path = ioutil::resolve_path(&path)?;
-                let mut sheet = calamine::open_workbook_auto(path)?;
+                let mut sheets = calamine::open_workbook_auto(path)?;
 
                 let first_sheet = sheet_name.unwrap_or_else(|| {
-                    let sheets = sheet.sheet_names();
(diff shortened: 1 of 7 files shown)


talagluck:
I added a worksheet and tests, but I can't get the tests to pass locally. I get
```
Error while running test `sqllogictests/xlsx` error=test fail: query failed: internal error: cannot execute simple query: db error: ERROR: External error: External error: Table 'users' not found
```

I tried with a few different tables, but they are all not found.

PR #3122, opened 2024-07-30
fix: clearer error handling around nested arrays

### LINKED ISSUE #3120: BUG: Issue with jaq functions when filtering results in a list of elements
## Description
[sample.json](https://github.com/user-attachments/files/16425871/sample.json)

```json
{
    "paging": {
        "total": 100,
        "page": 1,
        "per_page": 20
    },
    "elements": [
        {
            "id": 1,
            "name": "Charlie",
            "age": 35,
            "city": "Paris",
            "active": true
        },
        {
            "id": 2,
            "name": "Alice",
            "age": 42,
            "city": "Tokyo",
            "active": false
        },
        {
            "id": 3,
            "name": "Eva",
            "age": 28,
            "city": "London",
            "active": true
        },
        {
            "id": 4,
            "name": "Bob",
            "age": 53,
            "city": "New York",
            "active": false
        },
        {
            "id": 5,
            "name": "David",
            "age": 39,
            "city": "Sydney",
            "active": true
        }
    ]
}
```

Given the above/attached json, I can do this:
```sql
create external table posts from json
	options (
	location => '~/Downloads/sample.json',
	jaq_filter => '.paging'
	);
```

`paging` just contains key, value pairs.

However, when I instead try and filter on `elements`:
```sql
create external table posts from json
	options (
	location => '~/Downloads/sample.json',
	jaq_filter => '.elements'
	);
```
I get an error when I try and select * from the external table:
```
Error: Error during planning: Unable to fetch table provider for 'posts': Unsupported json type: only objects and arrays of objects are supported
```
I think this is because `elements` contains a list of items.

tychoish:
Closes #3120

### DIFF
--- a/crates/datasources/src/json/errors.rs
+++ b/crates/datasources/src/json/errors.rs
@@ -4,7 +4,10 @@ use datafusion_ext::errors::ExtensionError;
 #[derive(Debug, thiserror::Error)]
 pub enum JsonError {
     #[error("Unsupported json type: {0}")]
-    UnspportedType(&'static str),
+    UnsupportedType(&'static str),
+
+    #[error("Unsupported json type: cannot unwind nested arrays: {0}")]
+    UnsupportedNestedArray(&'static str),
 
     #[error(transparent)]
     SerdeJson(#[from] serde_json::Error),
--- a/crates/datasources/src/json/stream.rs
+++ b/crates/datasources/src/json/stream.rs
@@ -192,9 +192,21 @@ impl JsonHandler {
                                 match v {
                                     Value::Object(doc) => out.push(Ok(doc)),
                                     Value::Null => out.push(Ok(Map::new())),
+                                    Value::Array(_) => {
+                                        out.push(if filter.is_some() {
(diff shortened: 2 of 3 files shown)


talagluck:
The current error returned is:
```
Error: Error during planning: Unable to fetch table provider for 'posts': Unsupported json type: cannot unwind nested arrays: must return objects from jaq expressions
```

Could we remove one of the "Error"s? 

Also, I wonder if it's worth adding an example in the error message here (like "must return objects from jaq expressions, e.g. '.elements[]' instead of '.elements'"). This feels very clumsy, but this set of operations is something that I'm not very familiar with, and I imagine most people working with GlareDB here will have more experience with tabular data and trying to join in a json object instead of the other way around. It would be good to help people get unstuck, and when I search for the current error message, I don't get anything immediately helpful.

tychoish:
`Error during planning` is a datafusion error, so I can't get rid of that error, and the outer error, I think is happening at the top level in the python bindings, so practically speaking there's not a good way to get out of that one.

PR #3131, opened 2024-08-02
fix: clarify lazy API in bindings

tychoish:
After a discussion we had yesterday, this is more clear than "we have
this API that's lazy and you can rexecute it, unless your statement
has side effects..." which is the other way to avoid confusion and
errors, it seems.


bindings/python/tests/test_execution_model.py
```diff
@@ -2,30 +2,20 @@
...
-
-    two = con.sql("insert into tblsqlhelper values (4, 2);")
-    assert con.sql("select * from tblsqlhelper;").to_arrow().num_rows == 1
 
-    with pytest.raises(Exception, match="Duplicate name"):
-        one.execute()
-
-    assert con.sql("select count(*) from tblsqlhelper;").to_arrow().to_pydict()["COUNT(*)"][0] == 1
+    with pytest.raises(Exception):
+        con.sql("select count(*) from tblsqlhelper;")
```
talagluck, line 9: Why does this raise an error?

bindings/python/tests/test_execution_model.py
```diff
@@ -2,30 +2,20 @@
...
+def test_sql_excludes_ddl():
+    con = glaredb.connect()
 
-    two.execute()
-    assert con.sql("select * from tblsqlhelper;").to_arrow().num_rows == 3
+    with pytest.raises(Exception):
+        con.sql("select count(*) from tblsqlhelper;").to_arrow().to_pydict()["COUNT(*)"][0]
 
+    with pytest.raises(Exception):
+        con.sql("create table tblsqlhelper (a int, b int);")
```
talagluck, line 18: I understand why this raises, but not why the selects would.

talagluck (commented):
Thanks for opening! I had a couple of questions about tests, and do want to chat about this strategy a bit more before we cut a release, since a bunch of content was made with `.sql()` DDL operations.

PR #3133, opened 2024-08-05
fix: clarify evaluation semantics for DDL operations in bindings

tychoish:
This restores the legacy semantics (ish) but caches the results for eagerly executed options so that calling the resolving methods (e.g. `to_pandas()`) won't error in the case of rexecution.

This will be weird in cases where `INSERT INTO foo VALUES(123, now()) "tick-tock")`, which you might expect to insert more than one row if it were being lazily evaluated in the old way.

### DIFF
--- a/crates/glaredb/src/lib.rs
+++ b/crates/glaredb/src/lib.rs
@@ -562,7 +562,7 @@ impl Operation {
 
         match self.op {
             OperationType::Sql => {
-                match &plan {
+                match plan.clone() {
                     sqlexec::LogicalPlan::Datafusion(dfplan) => match dfplan {
                         LogicalPlan::Dml(_)
                         | LogicalPlan::Ddl(_)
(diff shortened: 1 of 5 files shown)


crates/glaredb/src/lib.rs
```diff
@@ -635,15 +621,24 @@ impl Operation {
...
-                Ok(Self::process_result(
-                    ses.execute_logical_plan(plan, &op).await?.1,
-                ))
-            }
+            OperationType::Execute => match &self.results {
+                Some(batches) => {
+                    return Ok(Box::pin(RecordBatchStreamAdapter::new(
+                        self.schema().clone().unwrap(),
+                        #[allow(clippy::unnecessary_to_owned)]
+                        stream::iter(batches.to_owned().into_iter().map(Ok)).boxed(),
```
scsmithr, line 629: Why is to_owned needed?

crates/sqlexec/src/session.rs
```diff
@@ -581,6 +581,11 @@ impl Session {
         self.ctx.take_notices()
     }
 
+    pub fn push_notice(&mut self, notice: Notice) {
+        self.ctx.push_notice(notice)
+    }
+
```
scsmithr, line 587:
```suggestion
```

scsmithr (approved):

PR #3161, opened 2024-08-20
fix: python bindings optional dependency organization

tychoish:


talagluck:
Thanks! It looks like Pola.rs requires PyArrow too. Should we pull that into both the Pandas and Polars dependencies, and remove it as its own thing?

talagluck (approved):
LGTM

PR #3244, opened 2024-10-16
fix: external schema for mongodb

tychoish:


crates/sqlexec/src/dispatch/external.rs
```diff
@@ -513,7 +513,7 @@ impl<'a> ExternalDispatcher<'a> {
                 let table_info = MongoDbTableAccessInfo {
                     database: database.to_string(),
                     collection: collection.to_string(),
-                    fields: None,
+                    fields: schema.map(|s| s.fields),
```
tychoish, line 516: this is the real fix, everything else in this PR makes it so that more of the "schemas" you might define for a table will work for assuming the types can be converted from trivially.

PR #3362, opened 2024-12-13
feat: Lateral table functions + some refactoring

scsmithr:
```
>> create temp table t (a int, b int);
┌─────────────────────┐
│ Query success       │
│ No columns returned │
└─────────────────────┘

>> insert into t values (3,6), (8,9);
┌───────────────┐
│ rows_inserted │
│ UInt64        │
├───────────────┤
│             2 │
└───────────────┘

>> select * from t, generate_series(a, b) order by 1,2,3;
┌───────┬───────┬─────────────────┐
│ a     │ b     │ generate_series │
│ Int32 │ Int32 │ Int64           │
├───────┼───────┼─────────────────┤
│     3 │     6 │               3 │
│     3 │     6 │               4 │
│     3 │     6 │               5 │
│     3 │     6 │               6 │
│     8 │     9 │               8 │
│     8 │     9 │               9 │
└───────┴───────┴─────────────────┘
```


crates/rayexec_execution/src/execution/operators/table_inout.rs
```diff
@@ -0,0 +1,192 @@
...
+                        let arr = expr.eval(&batch)?;
+                        Ok(arr.into_owned())
+                    })
+                    .collect::<Result<Vec<_>>>()?;
+
+                state.additional_outputs = additional_outputs;
+
+                Ok(other)
+            }
+        }
```
scsmithr, line 128: Only allow two states out of the function.

